PR #190, opened 2026-04-11
fix: redact syu app root labels

### LINKED ISSUE #179: [Bug]: syu app exposes absolute workspace paths in its browser payload
## What happened?
The syu app API payload currently includes absolute workspace_root and spec_root paths, and the frontend displays those values directly.

A local reproduction of GET /api/app-data.json returned values like:
- workspace_root: /tmp/syu-sample-flow
- spec_root: /tmp/syu-sample-flow/docs/syu
- source_documents with raw YAML content

This is more host metadata than the browser UI needs for normal navigation, and if syu app is ever bound beyond localhost it leaks local filesystem structure along with the spec content.

The browser app should avoid disclosing absolute filesystem paths to the client by default.

Safer options would be:
- expose only workspace-relative display labels,
- gate absolute path disclosure behind an explicit debug mode, or
- redact the root values when serving the browser payload.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Repository: ugoite/syu

## Steps to reproduce
1. Run cargo run -- app /tmp/syu-sample-flow --bind 127.0.0.1 --port 4011
2. Request http://127.0.0.1:4011/api/app-data.json
3. Observe that the payload contains absolute workspace_root and spec_root values
4. Open the browser UI and notice that the absolute workspace root is also shown in the header

    workspace_root= /tmp/syu-sample-flow
    spec_root= /tmp/syu-sample-flow/docs/syu
    source_documents= 4
    first_path= core/core.yaml

tohboeh5:
## Summary
- redact browser-facing workspace and spec root labels instead of shipping absolute paths
- keep workspace-relative spec labels when the spec lives under the workspace and use a generic label for external spec roots
- add app payload coverage so the HTTP API no longer leaks absolute root paths

Closes #179

## Validation
- scripts/ci/quality-gates.sh
- cargo test --test app_command
- cd app && npm run test:e2e

### DIFF
--- a/src/command/app.rs
+++ b/src/command/app.rs
@@ -433,6 +433,7 @@ fn build_app_payload(workspace_root: &Path) -> Result<AppPayload> {
 
 fn build_app_payload_from_config(workspace_root: &Path, config: &SyuConfig) -> Result<AppPayload> {
     let spec_root = resolve_spec_root(workspace_root, config);
+    let (workspace_label, spec_label) = browser_root_labels(workspace_root, &spec_root);
 
     let mut source_documents = Vec::new();
     source_documents.extend(collect_yaml_sources_recursive(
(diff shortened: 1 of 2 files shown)


src/command/app.rs
```diff
@@ -454,13 +455,27 @@ fn build_app_payload_from_config(workspace_root: &Path, config: &SyuConfig) -> R
...
-        spec_root: spec_root.display().to_string(),
+        workspace_root: workspace_label,
+        spec_root: spec_label,
         source_documents,
         validation: validation_snapshot(collect_check_result(workspace_root)),
     })
 }
 
+fn browser_root_labels(workspace_root: &Path, spec_root: &Path) -> (String, String) {
+    let workspace_label = workspace_root
```
tohboeh5: Redacting the absolute workspace root is a good security fix, but reducing it all the way down to `file_name()` also makes different clones of the same repo indistinguishable in the UI and in screenshots. A cwd-relative or home-relative label would still avoid leaking the full path while preserving more context than just the last segment.

tohboeh5:
Overview: the PR fixes the highest-risk part of the payload by stopping the browser app from publishing absolute workspace/spec roots. The tradeoff I would revisit is usability: the new workspace label is private, but it also collapses different clones with the same directory name into the same UI label, so the redaction could preserve a bit more safe context.

tohboeh5:
Addressed the root-label feedback: the browser payload now prefers cwd-relative or home-relative workspace labels, with a short trailing-segment fallback, and the redaction behavior is covered by regression tests.

PR #191, opened 2026-04-11
fix: warn on public syu app binds

### LINKED ISSUE #178: [Bug]: syu app does not warn before exposing workspace data on non-loopback binds
## What happened?
syu app allows non-loopback binds such as 0.0.0.0 without any warning, confirmation, or opt-in flag.

In my local reproduction, this startup sequence was accepted as normal:

    cargo run -- app /tmp/syu-sample-flow --bind 0.0.0.0 --port 4010
    syu app listening on http://0.0.0.0:4010
    syu app ready: http://0.0.0.0:4010
    Open http://0.0.0.0:4010 in your browser.

The app is not just serving static HTML. Its unauthenticated API returns the current workspace root, spec root, validation data, and full source document contents. That means a bind-to-all-interfaces choice can expose real repository metadata and YAML content to other machines on the network with no security warning at all.

When syu app binds to a non-loopback address, the CLI should treat that as a security-sensitive mode.

Examples of safer behavior:
- print a prominent warning that the workspace data is reachable from other machines,
- require an explicit opt-in flag such as --public when the bind address is not loopback, or
- at minimum document the exposure before printing the ready URL.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Repository: ugoite/syu

## Steps to reproduce
1. Run cargo run -- app /tmp/syu-sample-flow --bind 0.0.0.0 --port 4010
2. Observe that startup succeeds with no security warning
3. Request the JSON payload endpoint from another process
4. Notice that the server returns workspace metadata and source document contents over HTTP

    syu app listening on http://0.0.0.0:4010
    syu app ready: http://0.0.0.0:4010
    Open http://0.0.0.0:4010 in your browser.
    Press Ctrl-C to stop.

tohboeh5:
## Summary
- warn when syu app binds to a non-loopback address because workspace data becomes reachable over the network
- keep loopback startup output unchanged
- cover the warning path in app command integration tests

Closes #178

## Validation
- scripts/ci/quality-gates.sh
- cargo test --test app_command


src/command/app.rs
```diff
@@ -151,8 +151,12 @@ pub fn run_app_command(args: &AppArgs) -> Result<i32> {
             .await
             .context("local app readiness probe panicked")??;
         println!("syu app ready: http://{local_addr}");
+        maybe_warn_on_non_loopback_bind(local_addr.ip());
```
tohboeh5: This warning only appears after the app prints the ready/open guidance. On a non-loopback bind that means the first UX still looks like a normal local-success path, and users can copy the URL before they notice the exposure warning. I would move the warning ahead of the ready/open instructions (or even before the readiness wait) so the security signal is the first thing they see for public binds.

tohboeh5:
Overview: adding an explicit warning for non-loopback binds is the right security direction, and the wording itself is clear. The main issue is timing: the warning is emitted after the app announces readiness and tells the user to open the URL, so the public-bind caution arrives later than the success path instead of leading it.

PR #192, opened 2026-04-11
fix: clarify workspace help defaults

### LINKED ISSUE #170: [Bug]: CLI WORKSPACE help text contradicts the actual default path
## What happened?
Several subcommands describe their WORKSPACE argument as "Workspace root containing syu.yaml and the spec tree (default: docs/syu)" while Clap also renders [default: .] in the same help output.

I reproduced this with syu browse --help, syu show --help, syu validate --help, syu app --help, and syu report --help. The actual default workspace is the current directory (.), not docs/syu, so the help text contradicts itself and nudges first-time users toward the spec directory instead of the workspace root.

This cuts against the project philosophy of keeping setup and navigation low-friction for new adopters.

The help text should describe the current-directory workspace default accurately.

If docs/syu needs to appear anywhere, it should be explained as the default spec.root inside the workspace, not as the default WORKSPACE argument.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Repository: ugoite/syu

## Steps to reproduce
1. Run cargo run -- validate --help
2. Observe the WORKSPACE argument description
3. Notice that the line says (default: docs/syu) while the generated default shown by Clap is [default: .]
4. Repeat with browse, show, app, or report

    Usage: syu validate [OPTIONS] [WORKSPACE]

    Arguments:
      [WORKSPACE]  Workspace root containing syu.yaml and the spec tree (default: docs/syu) [default: .]

tohboeh5:
## Summary
- remove the misleading docs/syu default from shared WORKSPACE help text
- keep Clap's current-directory default as the only default shown in command help
- trace the new help regression test to the documentation requirement

Closes #170

## Validation
- scripts/ci/quality-gates.sh

### DIFF
--- a/tests/help_command.rs
+++ b/tests/help_command.rs
@@ -26,7 +26,7 @@ fn root_help_includes_start_here_guidance() {
 
 #[test]
 fn workspace_help_uses_current_directory_default_consistently() {
-    for command in ["browse", "show", "app", "validate", "report"] {
+    for command in ["browse", "show", "app", "validate", "check", "report"] {
         let output = Command::cargo_bin("syu")
             .expect("binary should build")
             .args([command, "--help"])
--- a/tests/help_command.rs
+++ b/tests/help_command.rs
@@ -23,3 +23,30 @@ fn root_help_includes_start_here_guidance() {
     ));
     assert!(stdout.contains("Start a local HTTP server and browser UI for workspace exploration"));
 }
+
+#[test]
+fn workspace_help_uses_current_directory_default_consistently() {
+    for command in ["browse", "show", "app", "validate", "report"] {
+        let output = Command::cargo_bin("syu")
+            .expect("binary should build")
+            .args([command, "--help"])
+            .output()
+            .expect("help should render");
+
+        assert!(output.status.success(), "{command} help should succeed");
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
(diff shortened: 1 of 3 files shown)


tests/help_command.rs
```diff
@@ -23,3 +23,30 @@ fn root_help_includes_start_here_guidance() {
     ));
     assert!(stdout.contains("Start a local HTTP server and browser UI for workspace exploration"));
 }
+
+#[test]
+fn workspace_help_uses_current_directory_default_consistently() {
+    for command in ["browse", "show", "app", "validate", "report"] {
```
tohboeh5: The help text drift was fixed for `validate`, but this regression test still skips the public `check` alias. Because `check` is a visible command entry point, adding it here would keep the alias from silently drifting back to the old wording in a future CLI help refactor.

tohboeh5:
Overview: the cleanup is directionally right and the new shared `WORKSPACE_HELP` constant makes the visible CLI wording much more consistent. The remaining gap is regression coverage: this PR fixes user-facing help text for the validate/check surface, but the added test only exercises `validate` and not the public `check` alias, so the alias can drift again without this test catching it.

tohboeh5:
Added the workspace-help regression test for the public `check` alias as well, then reran the repository quality gates to keep the alias and `validate` in sync.

PR #193, opened 2026-04-11
fix: reject typoed list kinds before workspace lookup

### LINKED ISSUE #171: [Bug]: syu list treats a misspelled layer kind as a workspace path
## What happened?
syu list overloads a single positional argument so that one token can mean either a layer kind or a workspace path.

That becomes very confusing for first-time users when they make a small typo in the kind name. For example, syu list philsophy is not rejected as an invalid kind. Instead, syu treats philsophy as a workspace path and fails later with "failed to resolve workspace root".

As a result, the CLI reports a filesystem problem when the real mistake is just a misspelled layer name.

If a single positional argument is close to a known kind but does not match one, syu list should surface that as a kind/input error and suggest valid layer names.

At minimum, the recovery message should point users to syu list --help or the valid kinds instead of making them debug a nonexistent workspace path.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Repository: ugoite/syu

## Steps to reproduce
1. Run cargo run -- list philsophy
2. Observe that syu tries to resolve philsophy as a workspace path
3. Notice that the command does not explain that philsophy is probably an invalid layer kind

    error: failed to resolve workspace root philsophy: No such file or directory (os error 2)

tohboeh5:
## Summary
- reject single-argument list kind typos before falling through to workspace loading
- suggest the closest supported kind and point users back to `syu list --help`
- cover the new CLI behavior with parser and integration tests

Closes #171

### DIFF
--- a/src/command/list.rs
+++ b/src/command/list.rs
@@ -85,23 +85,76 @@ fn parse_list_positionals(positional: &[String]) -> Result<(Option<LookupKind>, 
         [first] => {
             if let Ok(kind) = LookupKind::from_str(first, true) {
                 Ok((Some(kind), PathBuf::from(".")))
+            } else if PathBuf::from(first).exists() {
+                Ok((None, PathBuf::from(first)))
+            } else if let Some(suggestion) = suggested_lookup_kind(first) {
+                Err(invalid_kind_error(first, None, Some(suggestion)))
             } else {
                 Ok((None, PathBuf::from(first)))
             }
         }
         [kind_str, workspace_str, ..] => {
-            let kind = LookupKind::from_str(kind_str, true).map_err(|_| {
-                anyhow!(
-                    "invalid value '{}' for KIND\n  \
-                     possible values: philosophy, policy, requirement, feature\n  \
-                     Hint: to list all kinds in a workspace, run `syu list {}`",
-                    kind_str,
-                    workspace_str
-                )
-            })?;
+            let kind = LookupKind::from_str(kind_str, true)
+                .map_err(|_| invalid_kind_error(kind_str, Some(workspace_str), None))?;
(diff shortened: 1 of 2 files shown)


src/command/list.rs
```diff
@@ -85,25 +85,78 @@ fn parse_list_positionals(positional: &[String]) -> Result<(Option<LookupKind>,
...
-                     possible values: philosophy, policy, requirement, feature\n  \
-                     Hint: to list all kinds in a workspace, run `syu list {}`",
-                    kind_str,
-                    workspace_str
-                )
-            })?;
+            let kind = LookupKind::from_str(kind_str, true)
+                .map_err(|_| invalid_kind_error(kind_str, Some(workspace_str), None))?;
             Ok((Some(kind), PathBuf::from(workspace_str)))
         }
```
tohboeh5, line 105: The typo hint only applies to the single-argument form. `syu list philsophy .` still goes through the two-positional branch and drops the new `did you mean` guidance, even though that typo is just as clear there. It would be better to thread `suggested_lookup_kind` through the two-argument error path as well so the UX stays consistent regardless of whether the user also passed a workspace.

tohboeh5:
Overview: the PR fixes the worst UX failure by rejecting a misspelled kind before turning it into a workspace lookup error, and the single-argument coverage is solid. The remaining inconsistency is that the new typo suggestion only appears for `syu list <typo>`; once a workspace argument is present, the parser falls back to the old generic error, so users still get two different typo experiences for the same mistake.

tohboeh5:
Threaded the lookup-kind suggestion through the two-argument parse path as well, and added regression coverage for both the parser and the CLI surface.

PR #194, opened 2026-04-11
fix: accept workspace-first syu list arguments

### LINKED ISSUE #172: [Bug]: syu list help and recovery hint mislead users about positional argument order
## What happened?
syu list --help currently exposes a variadic positional argument as [KIND_OR_WORKSPACE]... with the description "Optional kind (philosophy|policy|requirement|feature) and/or workspace path".

That reads like users can provide the layer kind and workspace in either order, but the implementation only accepts [KIND] [WORKSPACE].

If a first-time user follows the apparent workspace-first interpretation and runs syu list /tmp/syu-list-order-demo requirement, syu treats the workspace path as an invalid kind and then prints a recovery hint that drops the workspace entirely.

This creates a confusing failure in one of the core read-only commands.

One of these should happen consistently:
- accept both syu list requirement /tmp/syu-list-order-demo and syu list /tmp/syu-list-order-demo requirement, or
- make the help output explicitly document only [KIND] [WORKSPACE], and show an error hint that preserves the chosen workspace path.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Repository: ugoite/syu

## Steps to reproduce
1. Run syu init /tmp/syu-list-order-demo
2. Run cargo run -- list /tmp/syu-list-order-demo requirement
3. Observe that the workspace path is parsed as a kind and the hint suggests syu list requirement, which loses the workspace argument
4. Compare with the working order: cargo run -- list requirement /tmp/syu-list-order-demo

    Usage: syu list [OPTIONS] [KIND_OR_WORKSPACE]...

    Arguments:
      [KIND_OR_WORKSPACE]...  Optional kind (philosophy|policy|requirement|feature) and/or workspace path

    error: invalid value /tmp/syu-list-order-demo for KIND
      possible values: philosophy, policy, requirement, feature
      Hint: to list all kinds in a workspace, run syu list requirement

tohboeh5:
## Summary
- accept `syu list <workspace> <kind>` in addition to the existing kind-first order
- keep workspace-first recovery hints anchored to the workspace path when the kind is invalid
- document both argument orders in the list help examples and cover the behavior with tests

Closes #172

### DIFF
--- a/src/command/list.rs
+++ b/src/command/list.rs
@@ -2,7 +2,7 @@
 // FEAT-LIST-002
 // REQ-CORE-018
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use anyhow::{Result, anyhow};
 use clap::ValueEnum as _;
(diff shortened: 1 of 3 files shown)


src/command/list.rs
```diff
@@ -89,21 +89,39 @@ fn parse_list_positionals(positional: &[String]) -> Result<(Option<LookupKind>,
...
+}
+
+fn parse_list_two_positionals(first: &str, second: &str) -> Result<(Option<LookupKind>, PathBuf)> {
+    match (
+        LookupKind::from_str(first, true),
+        LookupKind::from_str(second, true),
+    ) {
+        (Ok(kind), _) => Ok((Some(kind), PathBuf::from(second))),
+        (_, Ok(kind)) => Ok((Some(kind), PathBuf::from(first))),
+        (Err(_), Err(_)) if looks_like_workspace_path(first) => {
```
tohboeh5: The workspace-first support makes sense, but this parser now treats `syu list philosophy feature` as `kind=philosophy, workspace=feature` and falls through to a workspace resolution error. Since both tokens are recognized layer kinds, this should probably be rejected earlier with a direct "pick only one KIND" style error instead of sending users down the workspace path.

tohboeh5:
Overview: supporting both `syu list KIND WORKSPACE` and `syu list WORKSPACE KIND` is a meaningful UX improvement, and the help text/tests cover the happy path well. The remaining edge case is ambiguous two-kind input: when both positionals are valid kinds, the parser still turns the second one into a workspace path and produces a misleading filesystem error instead of a direct CLI-usage error.

PR #195, opened 2026-04-11
docs: describe syu app validation snapshots accurately

### LINKED ISSUE #183: [Bug]: app.md says the browser validation panel comes from the last syu validate run
## What happened?
The app guide currently says that the validation panel lists issues from "the last syu validate run that was used to generate the app data."

That does not match the current implementation. The app payload is built from the current workspace and the current validation snapshot inside syu app itself, so readers can come away thinking they need to run syu validate before using the browser app or that the browser is showing older validation state than it really is.

The app guide should describe the current behavior accurately: syu app computes the payload and validation snapshot directly from the current workspace, and the browser refresh flow updates from that live snapshot.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Reading path: docs/guide/app.md
- Repository: ugoite/syu

## Steps to reproduce
1. Open docs/guide/app.md
2. Read the Validation panel section
3. Compare that text with src/command/app.rs, where the app payload is built from the current workspace state

    docs text: The validation panel appears at the bottom of the page and lists all issues found during the last syu validate run that was used to generate the app data.

tohboeh5:
## Summary
- explain that `syu app` computes its own validation snapshot instead of relying on a prior `syu validate` run
- update the workflow guidance to use the browser refresh flow instead of restarting the app
- keep `syu validate` positioned as an optional terminal view of the same diagnostics

Closes #183

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -108,7 +108,10 @@ evidence still needs to be added before the item becomes implemented.
 
 ## Validation panel
 
-The validation panel appears at the bottom of the page and lists all issues found during the last `syu validate` run that was used to generate the app data.
+The validation panel appears at the bottom of the page and lists the current
+issues from the workspace snapshot that `syu app` loaded itself. You do not
+need to run `syu validate` first: the app computes the same validation snapshot
+when it starts and whenever the browser refresh flow loads updated spec data.
 
 Each row shows:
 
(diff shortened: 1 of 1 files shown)


docs/guide/app.md
```diff
@@ -144,8 +147,9 @@ Use the validation flow in this order:
 2. Pick the requirement document you are working on from the sidebar.
 3. Open a requirement with `status: planned`.
 4. In the detail pane, look for the planned placeholder under **Tests**.
-5. Add the missing trace entries in YAML, run `syu validate .`, then refresh the
-   app snapshot by restarting `syu app .`.
+5. Add the missing trace entries in YAML, then wait for the browser refresh flow
```
tohboeh5: This wording makes the refresh sound unconditional, but the browser only polls for snapshot changes while the tab is visible (`document.hidden` short-circuits the refresh loop in `App.tsx`). If a user edits the spec with the app in a background tab, "just wait" is misleading. I would call out that the tab needs to stay visible (or be revisited) before the updated validation snapshot appears.

tohboeh5:
Overview: this is a clear improvement over the old wording because it correctly explains that `syu app` computes its own validation snapshot instead of depending on a prior `syu validate` run. The main doc gap left is refresh semantics: the new text reads as if the browser snapshot updates automatically in all cases, but the current frontend only polls while the tab is visible, so readers may still be surprised when a background tab stays stale.

PR #196, opened 2026-04-11
docs: warn about public syu app binds

### LINKED ISSUE #184: [Bug]: app.md shows 0.0.0.0 bind examples without explaining the exposure risk
## What happened?
The app guide presents syu app . --bind 0.0.0.0 --port 8080 as a normal starting option, but it does not explain that this exposes the browser app to other machines on the network.

That omission matters because the app serves workspace metadata and source document contents over HTTP. A newcomer reading the guide can easily copy the 0.0.0.0 example for a container or remote environment without realizing that it changes the security model from localhost-only browsing to network-visible access.

The app guide should explain that 127.0.0.1 is the safe local default and that 0.0.0.0 should only be used deliberately when network exposure is intended.

A short warning in the Starting options section would be enough to keep newcomers from treating both examples as equally safe defaults.

## syu version
0.0.1-alpha.7

## Environment
- OS: Linux
- Reading path: docs/guide/app.md
- Repository: ugoite/syu

## Steps to reproduce
1. Open docs/guide/app.md
2. Read the Starting options section
3. Notice that the 0.0.0.0 bind example is shown without any warning about network exposure
4. Compare that guidance with the fact that syu app serves workspace_root, spec_root, and source documents over HTTP

    syu app . --port 8080
    syu app . --bind 0.0.0.0 --port 8080

tohboeh5:
## Summary
- explain that `127.0.0.1` is the safe local default for `syu app`
- warn that `0.0.0.0` exposes the browser UI to other machines on the network
- keep the bind examples while making the security tradeoff explicit

Closes #184

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -178,6 +178,11 @@ syu app . --port 8080
 syu app . --bind 0.0.0.0 --port 8080
 ```
 
+Keep a loopback bind such as `127.0.0.1` or `::1` for normal local use.
+Binding to `0.0.0.0`, `::`, or any other non-loopback address makes the app
+reachable from other machines on the network, so only do that deliberately
+when remote access is part of your setup.
+
 ### Persistent config
 
 Set defaults in `syu.yaml` so you do not have to pass flags every time:


docs/guide/app.md
```diff
@@ -178,6 +178,10 @@ syu app . --port 8080
 syu app . --bind 0.0.0.0 --port 8080
 ```
 
+Keep `127.0.0.1` for normal local use. Switching to `0.0.0.0` makes the app
```
tohboeh5: This is a good start, but the warning is still IPv4-specific. `syu app` accepts any `IpAddr`, so `::` has the same public-bind exposure story as `0.0.0.0`. I would mention both forms here so readers do not assume the risk only applies to the IPv4 wildcard address shown in the example.

tohboeh5:
Overview: the added warning improves the guide by finally calling out that `0.0.0.0` is not the normal safe local-development bind. The remaining doc gap is scope: the text frames the risk around the IPv4 wildcard example, but the CLI accepts IPv6 binds too, so the guidance should cover public non-loopback addresses more generally.

PR #198, opened 2026-04-11
fix: keep duplicate validation issues independently selectable

### LINKED ISSUE #169: [Bug]: Browser app validation detail collapses duplicate rule-code rows
## What happened?

In the browser app, the validation panel cannot distinguish between multiple issues that share the same rule code. The selected issue state is keyed only by `issue.code`, so clicking a second row with the same code still keeps the first matching issue selected.

This is reproducible with `tests/fixtures/workspaces/failing`, which contains duplicate `SYU-trace-id-001` and duplicate `SYU-trace-symbol-003` issues. In practice that means an advanced user can see multiple rows in the issue list, but cannot reliably inspect the matching detail panel for the specific row they clicked.

Each validation row should map to its own detail state, even when several issues share the same rule code. Clicking the second `SYU-trace-id-001` row should show that row's subject, location, and suggestion, not the first `SYU-trace-id-001` issue.

## syu version

0.0.1-alpha.7

## Environment

Linux, local source checkout, `cargo run -- app tests/fixtures/workspaces/failing`, browser app served from the checked-in repo.

## Steps to reproduce

1. Run `cargo run -- app tests/fixtures/workspaces/failing`.
2. Open the printed local URL.
3. Scroll to **Current issues**.
4. Click the first `SYU-trace-id-001` row and note the selected issue detail.
5. Click the second `SYU-trace-id-001` row.
6. Observe that the detail panel still resolves by code and does not reliably switch to the second issue instance.

Relevant implementation detail in `app/src/App.tsx`:

```ts
const [selectedIssueCode, setSelectedIssueCode] = useState<string | null>(null);
const activeIssue = useMemo(() => {
  return (
    workspace?.validation.issues.find((issue) => issue.code === selectedIssueCode) ??
    workspace?.validation.issues[0] ??
    null
  );
}, [selectedIssueCode, workspace]);
```

The row identity needs to include more than the rule code.

tohboeh5:
## Summary
- track the selected validation row by snapshot index so duplicate rule codes no longer collapse in the browser UI
- add a Playwright case that exercises duplicate issue codes against the failing fixture workspace
- rebuild the checked-in browser bundle so the Rust-served app matches the updated frontend source

Closes #169


app/src/App.tsx
```diff
@@ -82,11 +82,11 @@ function App() {
       setSelectedItemId(firstDocument?.items[0]?.id ?? "");
     }
 
-    setSelectedIssueCode((current) => {
-      if (current && browserWorkspace.validation.issues.some((issue) => issue.code === current)) {
+    setSelectedIssueIndex((current) => {
```
tohboeh5: Switching from `code` to `index` fixes duplicate-row selection in the static case, but it also means the selected issue can silently jump to a different item whenever the validation list reorders on refresh. This app already auto-refreshes its payload, so preserving selection by ordinal slot is brittle. I would keep a stable identity (for example the new `validationIssueSelectionKey`) in state and resolve the current index from that key after each reload.

tohboeh5:
Overview: the PR correctly addresses the original bug where duplicate rule codes could not be inspected independently, and the added Playwright coverage is a strong improvement. The remaining weakness is refresh behavior: selection is now tied to array index, so if validation issues reorder after an auto-refresh, the detail pane can drift to a different issue even though the user never changed their selection.

PR #199, opened 2026-04-11
feat: add document paths to syu list output

### LINKED ISSUE #167: [Feature]: Show source document paths in syu list output

From a first-time CLI user perspective, `syu list requirement` and `syu list feature` are easy to read but they omit the source document path that should be edited next.

On the self-hosted repository, for example, the output is a long flat list of IDs and titles:

```text
REQ-CORE-010  Provide English documentation for concepts, workflows, and the docs site
REQ-CORE-016  Ship an agent skill for maintaining syu-driven repositories
...
```

That is enough to discover that an item exists, but not enough to answer the next practical question: "Which YAML file should I open?" A new contributor who is still learning the four-layer layout has to run another command (`syu show`) or inspect the tree manually. That adds friction right where `syu` tries to stay non-intrusive.

## What change do you want?

Add an optional source-path view to terminal listing output, for example one of these shapes:

```bash
syu list requirement --with-document
syu list feature --with-path
```

Suggested behavior:

1. Keep the current compact default output unchanged.
2. Add an opt-in flag that appends the checked-in YAML document path for each item.
3. In text output, keep the format easy to scan in a terminal.
4. In JSON output, include the same document path field so editor or shell tooling can jump directly to the file.
5. Reuse the existing lookup layer so `list`, `show`, and the browser app keep agreeing on source document paths.

## Specification impact

This likely extends `FEAT-LIST-001` / `FEAT-LIST-002` and the docs around terminal-first browsing. It also supports `PHIL-003` by making discovery lead directly into editing instead of forcing extra ceremony.

## Alternatives considered

- Keep using `syu show <ID>` after every `syu list`. Works, but it turns discovery into a two-command workflow.
- Tell users to switch to `syu app` for file context. Helpful, but some contributors deliberately stay terminal-first.
- Print paths by default. Probably too noisy for the small, review-friendly default output.

tohboeh5:
## Summary
- add an opt-in `--with-path` flag to `syu list` for checked-in YAML document paths in text and JSON output
- resolve document paths relative to the workspace root when possible, falling back to absolute paths for external `spec.root` layouts
- cover the new list output modes, external spec roots, and lookup/document loader error paths in tests

Closes #167


src/command/lookup.rs
```diff
@@ -120,4 +183,113 @@ impl<'a> WorkspaceLookup<'a> {
...
+        match kind {
+            LookupKind::Philosophy => {
+                let mut paths = BTreeMap::new();
+                let documents = load_philosophy_documents_with_paths(
+                    &self.workspace.spec_root.join("philosophy"),
+                )?;
+                for loaded in documents {
+                    let path = workspace_relative_display(self.workspace, &loaded.path);
+                    for item in loaded.document.philosophies {
+                        paths.insert(item.id, path.clone());
```
tohboeh5: The new path lookup is keyed only by `item.id`, so duplicate IDs in an invalid workspace will overwrite each other and every colliding row will show the last document path seen for that ID. `syu list` is often useful precisely when the workspace is broken, so this can misdirect users while they are debugging a duplicate-ID problem. I would preserve the document path alongside each loaded item (or key by document+id) instead of rebuilding it through a plain ID map.

tohboeh5:
Overview: adding an opt-in `--with-path` mode is a good compromise because it exposes document provenance without breaking the default short list output. The main correctness risk is duplicate IDs: the implementation reconstructs document paths through an `id -> path` map, so invalid workspaces with colliding IDs can show the wrong path for some rows, which undercuts the debugging value of the new flag.

PR #200, opened 2026-04-11
feat: detect unregistered feature documents

### LINKED ISSUE #173: [Feature]: Detect feature YAML files that exist on disk but are missing from features.yaml
When a developer extends a real workspace, it is easy to create a new feature YAML file under docs/syu/features and forget to add it to docs/syu/features/features.yaml.

In that state, syu validate still passes, syu list feature does not show the new feature, and syu browse --non-interactive also hides it. The repository now contains a real feature document that syu silently ignores.

That makes early adoption feel brittle because one of the first realistic maintenance mistakes produces a false sense of correctness.

## What change do you want?
Add a validation or discovery check for feature YAML files that exist on disk under the configured features tree but are missing from the feature registry.

A small implementation could:
- scan the feature root for YAML files other than features.yaml,
- compare them with the registered entries, and
- emit a focused validation error when a file exists on disk but is not registered.

The error should point directly at docs/syu/features/features.yaml and tell the user which file is missing from the registry.

## Specification impact
This likely needs a new workspace or graph validation rule plus updates to the getting-started guide, tutorial, and examples so the feature registry remains explicit but no longer fails silently.

It also supports the philosophy that the specification should stay accountable to the real repository state instead of quietly ignoring files that look authoritative to contributors.

## Alternatives considered
The current workaround is to remember the registry rule manually or wait until someone notices that a feature never appears in list or browse output.

A broader syu add command could reduce the chance of the mistake, but even with better scaffolding the validator should still catch stray feature documents already present on disk.

tohboeh5:
## Summary

Add a workspace validation rule that detects feature YAML documents on disk which are missing from `docs/syu/features/features.yaml`, while ignoring non-feature YAML such as the built-in validation rule catalog. This touches the validator, its rule catalog, validate command coverage, and the getting-started/tutorial docs.

- Issue: Closes #173
- Requirement / feature IDs: FEAT-CHECK-001

## Validation

## Release notes

- [ ] No release note needed


src/command/check.rs
```diff
@@ -1014,6 +1015,144 @@ fn validate_unique_ids<'a>(
...
+        .collect();
+
+    let mut discovered_paths = Vec::new();
+    collect_feature_yaml_paths(feature_root, &mut discovered_paths)?;
+    discovered_paths.sort();
+
+    for path in discovered_paths {
+        if path == registry_path
+            || registered_paths.contains(&path)
+            || !looks_like_feature_document(&path)?
```
tohboeh5: `looks_like_feature_document` can fail because an unregistered YAML file is malformed, and the `?` here bubbles that up as a generic registry-comparison failure located on `features.yaml`. That sends users to the wrong file: the actual defect is the stray candidate document, not the registry. I would surface parse/read failures against the offending path (or skip non-parseable candidates) instead of collapsing them into a registry error.

tohboeh5:
Overview: checking the on-disk feature tree against `features.yaml` is a strong addition because it closes a real discovery gap in `list`, `browse`, and validation. The main issue is diagnostics quality: malformed unregistered feature files currently get reported as if `features.yaml` were at fault, which will make the new rule harder to debug in the very drift scenarios it is supposed to help with.

PR #201, opened 2026-04-11
ci: check app dist freshness

### LINKED ISSUE #180: [Feature]: Add a CI freshness check for the checked-in app/dist bundle
The repository intentionally checks in app/dist and embeds that production bundle into the Rust CLI so end users do not need a separate frontend build step.

That makes the checked-in bundle part of the product, not just a local artifact. But the current CI and quality gates do not appear to verify that the committed app/dist contents are fresh relative to app/src, the WASM build, and the browser app build inputs.

As a result, a pull request can change the browser app source while accidentally leaving a stale embedded bundle behind.

## What change do you want?
Add a freshness check for the checked-in browser bundle.

A simple shape would be:
- build the browser app from a clean checkout,
- compare the regenerated app/dist with the committed app/dist, and
- fail CI when they differ.

That check could live in the browser-app job, quality gates, or a dedicated repository-quality test if the project prefers a code-based contract.

## Specification impact
This should be reflected in contributor guidance because app/dist is intentionally checked in and shipped by the CLI.

The change also supports the philosophy of keeping implementation, release delivery, and day-to-day repository reality aligned instead of letting the shipped bundle drift away from reviewed source.

## Alternatives considered
One alternative is to stop checking in app/dist and build it only during release packaging.

Given the current architecture, the smaller change is to keep the checked-in bundle but enforce freshness automatically.

tohboeh5:
## Summary

Add a reusable CI script that rebuilds the browser app and fails when `app/dist` changes, wire it into the `browser-app` workflow, document the contributor workflow, and refresh the checked-in bundle so the new gate lands green.

- Issue: Closes #180
- Requirement / feature IDs: FEAT-QUALITY-001

## Validation

## Release notes

- [x] No release note needed


scripts/ci/check-app-dist-freshness.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+# FEAT-QUALITY-001
+
+set -euo pipefail
+
+ensure_app_dependencies() {
```
tohboeh5, line 6: The CI job is deterministic because it runs `npm ci` right before this script, but the local helper is not: any pre-existing `app/node_modules` directory makes the check reuse whatever happens to be installed. That can hide stale-bundle problems behind outdated dependencies or lockfile drift. If this script is meant to mirror CI for contributors, I would either always run `npm ci` or at least validate that `node_modules` matches the current lockfile before skipping it.

tohboeh5:
Overview: adding a dedicated browser-bundle freshness check is a good repository-quality improvement, and folding it into the browser-app CI job keeps the checked-in `app/dist` artifact honest. The main weakness is reproducibility for local contributors: the helper script only does a best-effort dependency install, so it does not reliably recreate the same dependency state that CI uses when deciding whether the bundle is fresh.

PR #202, opened 2026-04-11
ci: check generated docs freshness

### LINKED ISSUE #181: [Feature]: Add a CI freshness check for committed docs/generated artifacts
The repository checks in generated documentation artifacts under docs/generated, including the checked-in validation report and the generated site-spec pages.

Those files are user-facing outputs, but the current quality gates only generate an ephemeral report under target/quality and the docs-site build regenerates content for the site build without obviously checking that the committed docs/generated files are still up to date.

That means a pull request can change docs/syu, validation behavior, or generated reference content while accidentally leaving stale checked-in generated docs behind.

## What change do you want?
Add a freshness check for the committed docs/generated artifacts.

A simple shape would be:
- regenerate docs/generated/site-spec with scripts/generate-site-docs.py,
- regenerate docs/generated/syu-report.md from a clean checkout,
- fail CI if the regenerated outputs differ from what is committed.

This could run in quality gates, in the docs-site workflow, or in a dedicated repository-quality check.

## Specification impact
This affects contributor workflow and generated-doc expectations, so CONTRIBUTING and any relevant docs should mention how to refresh these checked-in artifacts locally.

It also reinforces the repository philosophy that checked-in documentation should stay close to the real source of truth.

## Alternatives considered
The alternative is to stop checking these generated files into git and only publish them from CI.

Given the current repository design, an automated freshness gate is the smallest change that preserves the existing workflow while removing silent drift.

tohboeh5:
## Summary

Add a shared CI/local script that regenerates the checked-in `docs/generated` artifacts and fails on drift, wire it into the shared quality gate, document the refresh workflow, and make the self-hosted report render a stable relative workspace path so the generated report stays deterministic across machines.

- Issue: Closes #181
- Requirement / feature IDs: FEAT-QUALITY-001, FEAT-REPORT-001

## Validation

## Release notes

- [x] No release note needed

### DIFF
--- a/scripts/ci/check-generated-docs-freshness.sh
+++ b/scripts/ci/check-generated-docs-freshness.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# FEAT-QUALITY-001
+
+set -euo pipefail
+
+snapshot_generated_docs() {
+  if [[ ! -d docs/generated ]]; then
+    printf '__missing__\n'
+    return 0
+  fi
+
+  while IFS= read -r path; do
+    sha256sum "$path"
+  done < <(find docs/generated -type f | LC_ALL=C sort)
+}
+
+check_generated_docs_freshness() {
+  local repo_root
+  local before_snapshot
+  local after_snapshot
+
+  repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+  before_snapshot="$(mktemp)"
+  after_snapshot="$(mktemp)"
+  trap "rm -f '$before_snapshot' '$after_snapshot'" EXIT
+
+  cd "$repo_root"
+  snapshot_generated_docs >"$before_snapshot"
+  python3 scripts/generate-site-docs.py
+  cargo run --quiet -- report . --output docs/generated/syu-report.md >/dev/null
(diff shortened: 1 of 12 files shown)


scripts/ci/check-generated-docs-freshness.sh
```diff
@@ -0,0 +1,42 @@
...
+set -euo pipefail
+
+snapshot_generated_docs() {
+  if [[ ! -d docs/generated ]]; then
+    printf '__missing__\n'
+    return 0
+  fi
+
+  while IFS= read -r path; do
+    sha256sum "$path"
```
tohboeh5: This helper is documented for contributors, but `sha256sum` is not available by default on macOS. The repo already handles that portability issue in `package-release.sh` and even calls it out in the README, so this script should probably reuse the same `sha256sum`/`shasum -a 256` fallback pattern instead of hard-requiring the GNU tool.

tohboeh5:
Overview: adding a generated-docs freshness check is a sensible way to keep the checked-in site-spec artifacts honest, and making the report path stable for the committed snapshot addresses the obvious source of drift. The main gap is contributor portability: the new local script hardcodes `sha256sum`, even though this repository already acknowledges macOS users elsewhere and uses a checksum-tool fallback in other scripts.

PR #203, opened 2026-04-11
feat: add temporary validate overrides

### LINKED ISSUE #177: [Feature]: Add temporary CLI validation profiles or overrides for configuration-only checks
The configuration guide currently states that validate.allow_planned, validate.require_non_orphaned_items, validate.require_reciprocal_links, and validate.require_symbol_trace_coverage are configuration-only.

That means an advanced user who wants to trial stricter validation in CI, relax a rule during migration, or compare a local branch against a stricter profile has to edit syu.yaml just to run one experiment.

For mature teams, that makes advanced validation more intrusive than it needs to be, because temporary workflow choices become repository configuration churn.

## What change do you want?
Add temporary CLI-level overrides or named validation profiles for the configuration-only validation switches.

Possible shapes:
- syu validate . --require-reciprocal-links
- syu validate . --allow-planned=false
- syu validate . --profile strict
- syu validate . --profile migration

The important part is that users can try or automate stricter and looser validation modes without editing the committed syu.yaml for every experiment.

## Specification impact
This would extend the validate CLI contract and the precedence rules in the configuration guide.

It also matches the philosophy that stricter checks should be optional and easy to adopt, instead of requiring repository-wide config edits before a team can even evaluate them.

## Alternatives considered
Today the workaround is to edit syu.yaml temporarily, keep multiple config variants outside the tool, or write wrapper scripts that patch the config before running validate.

Those approaches work, but they add exactly the kind of ceremony that a small, non-intrusive workflow should avoid.

tohboeh5:
## Summary
- add temporary CLI overrides for the config-driven validate strictness switches
- cover the new override behavior in parse/help/integration tests and trace the new tests in the self-hosted spec
- update README, guides, self-hosted spec docs, and regenerated docs artifacts

Closes #177

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -335,7 +335,7 @@ impl ValidationGenreFilter {
 mod tests {
     use clap::Parser;
 
-    use super::{Cli, Commands, LookupKind, ValidationGenreFilter, ValidationSeverityFilter};
+    use super::{Cli, LookupKind, ValidationGenreFilter, ValidationSeverityFilter};
 
     #[test]
     fn cli_enums_expose_expected_labels() {
@@ -365,14 +365,12 @@ mod tests {
         ])
         .expect("validate args should parse");
 
-        let Commands::Validate(args) = cli.command.expect("validate command") else {
-            panic!("expected validate command");
-        };
-
-        assert_eq!(args.allow_planned, Some(false));
-        assert_eq!(args.require_non_orphaned_items, Some(false));
-        assert_eq!(args.require_reciprocal_links, Some(true));
-        assert_eq!(args.require_symbol_trace_coverage, Some(true));
+        let rendered = format!("{cli:?}");
+        assert!(rendered.contains("command: Some(Validate("));
+        assert!(rendered.contains("allow_planned: Some(false)"));
+        assert!(rendered.contains("require_non_orphaned_items: Some(false)"));
+        assert!(rendered.contains("require_reciprocal_links: Some(true)"));
+        assert!(rendered.contains("require_symbol_trace_coverage: Some(true)"));
     }
 
     #[test]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -7,7 +7,7 @@
 // FEAT-INIT-002
 // REQ-CORE-001
 
-use clap::{ArgAction, Args, Parser, Subcommand, ValueEnum};
+use clap::{ArgAction, Args, Parser, Subcommand, ValueEnum, builder::BoolishValueParser};
 use std::path::PathBuf;
 
 const ROOT_AFTER_HELP: &str = "\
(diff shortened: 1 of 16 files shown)


src/command/check.rs
```diff
@@ -567,6 +568,22 @@ fn effective_fix(args: &CheckArgs, config: &SyuConfig) -> bool {
     }
 }
 
+fn with_validate_overrides(mut workspace: Workspace, args: &CheckArgs) -> Workspace {
+    if let Some(value) = args.allow_planned {
```
tohboeh5, line 573: Now that `with_validate_overrides` can tighten `allow_planned` from the CLI, the downstream `SYU-delivery-planned-001` diagnostic becomes misleading: reproducing with `syu validate <workspace> --allow-planned=false` still says "`syu.yaml` forbids planned items" and suggests editing the config file, even when the checked-in config explicitly allows them. Please thread the effective source into the message/suggestion here so the new override workflow does not point users at the wrong fix.

src/cli.rs
```diff
@@ -301,4 +351,36 @@ mod tests {
...
+            ".",
+            "--allow-planned=false",
+            "--require-non-orphaned-items=false",
+            "--require-reciprocal-links",
+            "--require-symbol-trace-coverage",
+        ])
+        .expect("validate args should parse");
+
+        let Commands::Validate(args) = cli.command.expect("validate command") else {
+            panic!("expected validate command");
```
tohboeh5: This test shape is what drops the coverage gate to 99.99%: the `panic!("expected validate command")` line is now the only uncovered line in the whole report, so both `coverage` and `precommit` stay red even though the new feature works. Could you rewrite the extraction so it does not leave a dedicated panic-only line behind (for example by restructuring the assertion or collapsing the unreachable path onto an already-executed line)?

tohboeh5:
Overview:

I like the scope here: the flag surface stays intentionally small, the help text is discoverable, and the end-to-end fixtures make the override behavior easy to follow. The main blockers are around polish rather than the core idea.

First, the current CLI parser test introduces the single uncovered line that drops `coverage`/`precommit` below the repository gate, so the PR is red even though the new command path works. Second, once the overrides are applied, at least one validation diagnostic still tells the user that `syu.yaml` is responsible and nudges them toward editing config, which undermines the new one-off override workflow. After those two pieces are tightened up, this should be in good shape.

src/command/check.rs
tohboeh5: Fixed in d13d4a8: the planned-item diagnostic now points at `--allow-planned=false` instead of `syu.yaml`, so the one-off override workflow no longer suggests the wrong next step.

src/cli.rs
tohboeh5: Fixed in d13d4a8: I rewrote the parser assertion to avoid leaving a panic-only line behind, and the local coverage gate is back to 100%.

PR #204, opened 2026-04-11
feat: allow syu init custom spec roots

### LINKED ISSUE #176: [Feature]: Let syu init scaffold a workspace with a custom spec.root
syu already supports custom spec roots through syu.yaml, and the docs explicitly say existing repositories can keep another layout such as docs/spec.

But syu init always scaffolds docs/syu and has no way to generate the starter tree somewhere else. An advanced adopter who wants spec/contracts, docs/spec, or another repository-specific path must run init, move the generated tree by hand, and then edit syu.yaml manually.

That makes non-default adoption feel more ceremonial than the philosophy suggests.

## What change do you want?
Add a syu init option that can scaffold the initial workspace into a custom spec.root.

Example shapes:
- syu init . --spec-root spec/contracts
- syu init . --spec-root docs/spec

The command should write the generated philosophy, policy, requirement, feature, and feature registry files into that root and emit a matching spec.root value in syu.yaml.

## Specification impact
This changes the init command contract and the generated workspace layout, so the README, getting-started guide, tutorial, and migration guidance should all mention the new option.

It also reinforces the philosophy that syu should fit repositories without forcing a single layout when the repository already has a clear convention.

## Alternatives considered
The current workaround is to run syu init, move the scaffolded files manually, and then edit syu.yaml to match.

That works, but it is exactly the kind of avoidable ceremony that makes advanced adoption feel intrusive.

tohboeh5:
## Summary
- add `syu init --spec-root` for repository-relative starter layouts
- validate custom roots in the CLI and update next-step guidance
- document and trace the new bootstrap flow across tests and self-hosted spec

Closes #176

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,6 +2,7 @@
 // FEAT-APP-001
 // FEAT-BROWSE-001
 // FEAT-BROWSE-002
+// FEAT-INIT-003
 // FEAT-LIST-002
 // FEAT-REPORT-001
 // FEAT-INIT-002
@@ -280,6 +281,12 @@ pub struct InitArgs {
     #[arg(long)]
     pub name: Option<String>,
 
+    #[arg(
+        help = "Repository-relative spec.root to scaffold (for example docs/spec or spec/contracts)"
+    )]
+    #[arg(long)]
+    pub spec_root: Option<PathBuf>,
+
     #[arg(help = "Overwrite generated files when they already exist")]
     #[arg(long)]
     pub force: bool,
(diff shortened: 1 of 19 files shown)


src/cli.rs
```diff
@@ -381,4 +389,15 @@ mod tests {
...
         assert!(message.contains("--allow-planned"));
     }
+
+    #[test]
+    fn init_args_accept_custom_spec_root() {
+        let cli = Cli::try_parse_from(["syu", "init", ".", "--spec-root", "docs/spec"])
+            .expect("init args should parse");
+
+        let Some(Commands::Init(args)) = cli.command else {
+            panic!("expected init command");
```
tohboeh5: This new parser test adds an uncovered `panic!` fallback on the `else` branch. Because the repository keeps line coverage at 100%, that one executable line is enough to leave both `precommit` and `coverage` red at 99.99%. Could we rewrite the assertion so every executable line is hit here (for example by asserting on the parsed debug output, or by structuring the test without a fallback branch)?

tohboeh5:
I reviewed the `--spec-root` change end to end. The feature direction is good: the new flag fits the existing `init` workflow, the generated layout/docs stay coherent, and the CLI behavior is easy to understand from the updated help text.

The main blocker is CI. Both `precommit` and `coverage` are red because the new parser test in `src/cli.rs` introduces an uncovered fallback branch, which drops the repository back to 99.99% line coverage. Once that is rewritten so the coverage gate passes, this PR looks close.

PR #205, opened 2026-04-11
feat: add syu init starter templates

### LINKED ISSUE #175: [Feature]: Add language-oriented starter templates to syu init
syu ships good examples for rust-only, python-only, and polyglot repositories, but syu init always scaffolds the same generic starter workspace.

A developer adopting syu in a real project still has to study the examples and manually translate their repository shape into the new workspace. That is especially noticeable for multi-language projects where trace conventions and runtime expectations differ by language.

This makes the first hour with syu feel more manual than it needs to be.

## What change do you want?
Add lightweight starter templates to syu init so a new workspace can begin closer to the target repository style.

Examples:
- syu init . --template rust-only
- syu init . --template python-only
- syu init . --template polyglot

A template does not need to generate large sample code. It could stay small and only change the scaffolded YAML structure, starter titles, and any obviously relevant runtime sections.

## Specification impact
This would affect the init command contract and should be documented in the README, getting-started guide, tutorial, and example guidance.

It also fits the philosophy that syu should remain language-agnostic while still being easy to adopt in repositories that already know their primary language mix.

## Alternatives considered
Today the workaround is to run syu init and then manually copy patterns from examples/rust-only, examples/python-only, or examples/polyglot.

That keeps syu small, but it spreads the initial setup knowledge across multiple docs and examples instead of letting the CLI offer a closer starting point.

tohboeh5:
## Summary
- add `syu init --template` with generic, rust-only, python-only, and polyglot layouts
- tailor starter IDs, file locations, and next-step guidance to the selected template
- document and trace the new bootstrap flow across tests and self-hosted spec

Closes #175

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,6 +2,7 @@
 // FEAT-APP-001
 // FEAT-BROWSE-001
 // FEAT-BROWSE-002
+// FEAT-INIT-004
 // FEAT-LIST-002
 // FEAT-REPORT-001
 // FEAT-INIT-002
@@ -22,6 +23,12 @@ const APP_AFTER_HELP: &str = concat!(
     "Use GET /health for readiness checks once the app is serving.\n",
     "Press Ctrl-C to stop the local app server."
 );
+
+const INIT_AFTER_HELP: &str = "\
+Examples:
+  syu init .
+  syu init . --template rust-only
+  syu init path/to/workspace --name my-project --template polyglot";
 
 const WORKSPACE_HELP: &str = "Workspace root containing syu.yaml and the configured spec tree";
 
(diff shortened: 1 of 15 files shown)


src/cli.rs
```diff
@@ -381,4 +406,15 @@ mod tests {
...
         assert!(message.contains("--allow-planned"));
     }
+
+    #[test]
+    fn init_args_accept_template_values() {
+        let cli = Cli::try_parse_from(["syu", "init", ".", "--template", "rust-only"])
+            .expect("init args should parse");
+
+        let Some(Commands::Init(args)) = cli.command else {
+            panic!("expected init command");
```
tohboeh5: This parser assertion introduces an uncovered `panic!` fallback on the `else` branch. Because the repository enforces 100% line coverage, that single executable line keeps both `precommit` and `coverage` red at 99.99%. Could we restructure this test so every executable line is hit (for example by asserting on the parsed debug output, or by avoiding a fallback branch altogether)?

tohboeh5:
I reviewed the starter-template change end to end. The overall direction is strong: the new `--template` flag is easy to discover from help output, the generated Rust/Python/polyglot stubs stay small, and the scaffolded workspaces validate cleanly after generation.

The remaining blocker is CI. Both `precommit` and `coverage` are red because the new parser test in `src/cli.rs` adds an uncovered fallback branch, which leaves the repository at 99.99% line coverage. Once that test is rewritten so the coverage gate goes green, this looks close.

PR #206, opened 2026-04-11
feat: add init id prefix options

### LINKED ISSUE #174: [Feature]: Let syu init scaffold project-specific starter ID prefixes
The starter workspace generated by syu init uses generic IDs like PHIL-001, POL-001, REQ-001, and FEAT-001.

In a real project, teams usually want stable domain-specific IDs from the beginning, such as PHIL-STORE-001 or REQ-AUTH-001. Today, the first serious tutorial step after init is often to rename every starter ID and update all reciprocal links by hand.

That adds avoidable churn right at the moment when a team is deciding whether syu feels lightweight enough to adopt.

## What change do you want?
Add an init option that can seed project-specific ID prefixes into the starter scaffold.

Examples:
- syu init . --name filestore --id-prefix STORE
- syu init . --requirement-prefix REQ-STORE --feature-prefix FEAT-STORE

Even a single shared stem such as STORE would be enough if syu can render PHIL-STORE-001, POL-STORE-001, REQ-STORE-001, and FEAT-STORE-001 consistently.

## Specification impact
This would change the generated starter documents only, but it should be reflected in the init docs, tutorial, and examples because it affects the recommended shape of stable IDs.

It also supports the philosophy of explicit, durable identifiers without forcing users through a full manual rename pass immediately after bootstrap.

## Alternatives considered
The current workaround is to accept the generic IDs and rewrite them manually before the workspace becomes real.

That is workable for one item, but it becomes unnecessary ceremony when init could generate meaningful starter IDs directly.

tohboeh5:
## Summary
- add `syu init --id-prefix` plus per-layer `--*-prefix` overrides for starter IDs
- update init scaffolding, CLI help, docs, and self-hosted spec coverage
- add CLI and integration tests for shared stems, per-layer overrides, and validation

## Validation
- cargo test
- scripts/ci/quality-gates.sh

Closes #174

### DIFF
--- a/src/command/init.rs
+++ b/src/command/init.rs
@@ -523,8 +523,8 @@ mod tests {
             spec_root: None,
             template: StarterTemplate::RustOnly,
             id_prefix: Some("store".to_string()),
-            philosophy_prefix: None,
-            policy_prefix: None,
+            philosophy_prefix: Some("phil-guiding".to_string()),
+            policy_prefix: Some("pol-governance".to_string()),
             requirement_prefix: Some("req-auth".to_string()),
             feature_prefix: Some("feat-auth".to_string()),
             force: false,
@@ -532,8 +532,8 @@ mod tests {
         };
 
         let prefixes = resolve_init_id_prefixes(&args).expect("prefixes should resolve");
-        assert_eq!(prefixes.philosophy, "PHIL-STORE");
-        assert_eq!(prefixes.policy, "POL-STORE");
+        assert_eq!(prefixes.philosophy, "PHIL-GUIDING");
+        assert_eq!(prefixes.policy, "POL-GOVERNANCE");
         assert_eq!(prefixes.requirement, "REQ-AUTH");
         assert_eq!(prefixes.feature, "FEAT-AUTH");
     }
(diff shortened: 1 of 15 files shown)


src/command/init.rs
```diff
@@ -354,6 +515,49 @@ mod tests {
         }
     }
 
+    #[test]
+    fn resolve_init_id_prefixes_accepts_shared_stems_and_overrides() {
```
tohboeh5, line 519: The new override coverage still misses the `--philosophy-prefix` and `--policy-prefix` branches. On this PR, `syu-coverage-gate` is still red because the `resolve_init_id_prefixes` path for those flags never executes in CI. Please add a case that passes those two flags through `syu init` (or extends this unit test accordingly) so all four override branches are exercised.

tohboeh5:
Overview: the feature direction is solid. The CLI surface is small and consistent with `syu init`, the generated docs/spec updates are thorough, and the end-to-end behavior for shared stems and per-layer overrides looks correct. I did not find security or UX regressions in the implementation itself.

The main issue is test completeness: the PR adds several new validation/override branches, but the current suite still leaves some of those paths uncovered, so `syu-coverage-gate` and the pre-push hook stay red. In particular, the philosophy/policy override paths and the invalid typed-prefix error paths still need direct coverage. Once those cases are exercised, this will be in much better shape.

src/command/init.rs
```diff
@@ -354,6 +515,49 @@ mod tests {
...
+
+        let prefixes = resolve_init_id_prefixes(&args).expect("prefixes should resolve");
+        assert_eq!(prefixes.philosophy, "PHIL-STORE");
+        assert_eq!(prefixes.policy, "POL-STORE");
+        assert_eq!(prefixes.requirement, "REQ-AUTH");
+        assert_eq!(prefixes.feature, "FEAT-AUTH");
+    }
+
+    #[test]
+    fn resolve_init_id_prefixes_rejects_full_prefixes_as_shared_stems() {
```
tohboeh5, line 542: This only covers the shared-stem rejection. The new validation branches in `normalize_typed_prefix` and `normalize_id_token` are still uncovered, which keeps the coverage job below the repository 100% line gate. Please add at least one failing typed-override case, for example `--philosophy-prefix STORE` or an illegal token such as `--feature-prefix feat_store`, so the error messages are exercised before merge.

Issue #168, opened 2026-03-29
[Feature]: Add a terminal-first syu search command


tohboeh5:
## What problem are you trying to solve?

When using `syu` on a real repository, terminal-first exploration scales poorly once the spec grows beyond a handful of items. The browser app already has a search box, but the CLI has no equivalent keyword search workflow.

Today the fallback is usually one of these:

- run `syu list requirement` or `syu list feature` and visually scan a long list
- pipe CLI output into external `grep`
- leave the terminal and open `syu app` just to search by title or ID

That creates a mismatch between the browser and CLI surfaces. A contributor who prefers terminal workflows can browse and show exact IDs, but cannot ask a simple question like "show me anything related to audit" without leaving `syu`.

## What change do you want?

Add a dedicated terminal search command, for example:

```bash
syu search audit
syu search traceability --kind requirement
syu search FEAT-CHECK --format json
```

Suggested behavior:

1. Match against IDs, titles, summaries, and descriptions.
2. Support optional kind scoping (`philosophy`, `policy`, `requirement`, `feature`).
3. Return stable, script-friendly text and JSON output.
4. Reuse the same item index concepts as the browser app so results stay consistent across both surfaces.
5. Keep the command intentionally small: this should be lightweight repository navigation, not a full-text search engine.

## Specification impact

This likely belongs near `FEAT-LIST-001`, `FEAT-SHOW-001`, and `FEAT-BROWSE-001` as a terminal-first navigation feature. It also supports `PHIL-003` by reducing the ceremony required to find the next relevant spec item during maintenance work.

## Alternatives considered

- Continue using shell pipelines like `syu list feature | grep audit`. Simple, but it only searches whatever fields the current output happens to print.
- Rely on `syu app` search only. Good for some users, but it leaves terminal-first usage behind.
- Expand `syu list` with too many ad-hoc filters. Possible, but a small focused `search` command seems easier to explain.

tohboeh5:
This request is already implemented on main by PR #207. I re-verified the merged search command, its tests, and the current CLI help/output before closing this stale issue.

PR #207, opened 2026-04-11
Add terminal-first syu search command

tohboeh5:
## Summary\n- add a new `syu search` CLI for terminal-first lookup across IDs, titles, summaries, and descriptions\n- keep search behavior aligned with the browser search fields while supporting optional kind scoping and JSON output\n- document and trace the new command through the checked-in spec, generated docs, README, and tests\n\n## Validation\n- bash scripts/ci/quality-gates.sh\n- cargo run -- search source tests/fixtures/workspaces/passing --kind feature\n- cargo run -- search traceability . --kind requirement\n\nCloses #168

### DIFF
--- a/src/command/lookup.rs
+++ b/src/command/lookup.rs
@@ -1,3 +1,5 @@
+// FEAT-SEARCH-001
+// REQ-CORE-019
 // REQ-CORE-018
 
 use std::path::Path;
@@ -21,6 +23,13 @@ pub(crate) struct EntitySummary {
     pub title: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub document_path: Option<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub(crate) struct SearchResult {
+    pub id: String,
+    pub kind: &'static str,
+    pub title: String,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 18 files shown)


tohboeh5:
Overview: This is a strong feature addition overall. The command fits the existing CLI vocabulary well, the kind scoping/JSON output make sense for automation, and the spec/doc updates keep the change easy to follow. I also confirmed the main happy paths locally. The weak spot is test completeness: the repository requires a 100% coverage gate, and this PR still leaves the philosophy-positive branch plus the whitespace-only error path uncovered, so CI stays red. Once those gaps are covered, this should be in much better shape.

src/command/lookup.rs
```diff
@@ -214,6 +244,67 @@ impl<'a> WorkspaceLookup<'a> {
...
     }
+
+    fn extend_search_results(self, kind: LookupKind, query: &str, results: &mut Vec<SearchResult>) {
+        match kind {
+            LookupKind::Philosophy => {
+                for item in &self.workspace.philosophies {
+                    if field_matches_query(&item.id, query)
+                        || field_matches_query(&item.title, query)
+                    {
+                        results.push(SearchResult {
```
tohboeh5, line 255: This new positive LookupKind::Philosophy branch is still uncovered. The current suite only proves that philosophy free-text fields stay out of scope, so cargo llvm-cov is still missing src/command/lookup.rs:254-259 and the coverage gate fails. Please add at least one positive philosophy ID/title search case so this branch is exercised and CI can go green.

src/command/search.rs
```diff
@@ -0,0 +1,56 @@
...
+    query: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    kind: Option<&'static str>,
+    items: Vec<SearchResult>,
+}
+
+pub fn run_search_command(args: &SearchArgs) -> Result<i32> {
+    let query = args.query.trim();
+    if query.is_empty() {
+        bail!("search query must not be empty or whitespace");
```
tohboeh5, line 25: This whitespace-only guard is untested right now. I reproduced the CLI path locally and it exits with the expected error, but without a regression test this branch remains uncovered at src/command/search.rs:25, which is enough to keep the pre-push and coverage jobs red under the repo 100 percent line-coverage gate. Please add a command test for the whitespace-only query path.

tohboeh5:
Follow-up: GitHub blocked a formal review state here because this account is also the PR author. The two inline review comments above are the blocking items I found: one missing positive philosophy-search coverage case and one missing whitespace-query regression test.

Issue #148, opened 2026-03-29
[Feature]: Add a syu add workflow for ongoing spec authoring after syu init


tohboeh5:
## What problem are you trying to solve?

After `syu init`, the only built-in scaffolding flow is the initial workspace bootstrap. Once a team starts using syu for real work, adding new items becomes a repetitive manual YAML task:

- choose the next ID yourself
- copy/paste an existing philosophy/policy/requirement/feature block
- place it in the right file or folder
- update links to adjacent layers
- if it is a feature, remember to also update `docs/syu/features/features.yaml`

The current getting-started guide reflects this manual flow by telling users to edit `docs/syu/...` directly. That works for the first few items, but on a real project it becomes ceremony-heavy and easy to get wrong.

This is a concept-level friction point: syu aims to stay simple enough to adopt without resenting the tool, but repeated item authoring currently has no first-class command support after the initial `init` step.

## What change do you want?

Add a follow-up scaffolding command for ongoing spec authoring, for example:

```bash
syu add philosophy PHIL-002 --file docs/syu/philosophy/foundation.yaml
syu add policy POL-002 --file docs/syu/policies/policies.yaml
syu add requirement REQ-AUTH-001 --file docs/syu/requirements/auth/auth.yaml
syu add feature FEAT-AUTH-001 --file docs/syu/features/auth/login.yaml --kind auth
```

Suggested behavior:

1. Generate a correctly shaped YAML stub for the requested layer.
2. Offer sensible defaults for title/summary/status fields.
3. For features, update `docs/syu/features/features.yaml` automatically when needed.
4. Optionally support an interactive mode when ID or file is omitted.
5. Keep output intentionally small and review-friendly so the command stays aligned with syu's "small, specification-driven" philosophy.

## Specification impact

This likely needs a new CLI feature around authoring/scaffolding beyond FEAT-INIT-002. It also touches documentation because the getting-started and authoring guides could recommend `syu add ...` instead of copy/paste workflows.

## Alternatives considered

- Continue with manual YAML editing and editor snippets. This keeps the CLI smaller, but it pushes repetitive registry/link maintenance onto every contributor.
- Add more documentation only. Helpful, but it does not remove the ceremony or the easy-to-miss feature registry update.
- Build a fully interactive wizard for the entire spec graph. Probably too heavy; a focused `syu add` command is a smaller step.

PR #208, opened 2026-04-11
Add syu add authoring workflow

tohboeh5:
## Summary\n- add a new `syu add` command that scaffolds philosophy, policy, requirement, and feature YAML stubs after init\n- infer default document paths from IDs, honor custom `spec.root` layouts, and update the feature registry when new feature files are created\n- update guides, README, checked-in spec docs, generated docs, and tests around the new authoring workflow\n\n## Validation\n- bash scripts/ci/quality-gates.sh\n- cargo run -- add requirement REQ-AUTH-001 <temp workspace>\n- cargo run -- add feature FEAT-AUTH-LOGIN-001 <temp workspace> --kind auth\n\nCloses #148

### DIFF
--- a/src/command/add.rs
+++ b/src/command/add.rs
@@ -1,0 +1,673 @@
+// FEAT-ADD-001
+// REQ-CORE-020
+
+use std::{
+    fs,
+    path::{Component, Path, PathBuf},
+};
+
+use anyhow::{Context, Result, bail};
+
+use crate::{
+    cli::{AddArgs, LookupKind},
+    coverage::normalize_relative_path,
+    model::{
+        FeatureDocument, FeatureRegistryDocument, PhilosophyDocument, PolicyDocument,
+        RequirementDocument,
+    },
+    workspace::{Workspace, load_workspace},
+};
+
+use super::{lookup::WorkspaceLookup, shell_quote_path};
+
+pub fn run_add_command(args: &AddArgs) -> Result<i32> {
+    let workspace = load_workspace(&args.workspace)?;
+    let parsed_id = ParsedId::parse(args.layer, &args.id)?;
+
+    if args.layer != LookupKind::Feature && args.kind.is_some() {
+        bail!("`--kind` is only supported when scaffolding features");
+    }
+    if WorkspaceLookup::new(&workspace)
+        .find(&parsed_id.normalized)
+        .is_some()
+    {
+        bail!(
+            "{} `{}` already exists in `{}`",
+            args.layer.label(),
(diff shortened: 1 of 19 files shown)


tohboeh5:
Follow-up: GitHub will not let this account submit a formal review state on a pull request it authored itself. The inline review comments above capture the blocking items I found.

src/command/add.rs
```diff
@@ -0,0 +1,673 @@
...
+    let target = resolve_target_path(
+        &workspace,
+        args.layer,
+        args.file.as_deref(),
+        &parsed_id,
+        feature_kind.as_deref(),
+    )?;
+    let created_target_file = !target.absolute.exists();
+
+    write_stub_document(args.layer, &parsed_id, feature_kind.as_deref(), &target)?;
```
tohboeh5: This writes the feature document before we know whether the registry update can succeed. I reproduced it locally with: syu init <ws> && syu add feature FEAT-AUTH-NEW-001 <ws> --kind auth --file features/core/core.yaml. The command exits with "feature registry already tracks core/core.yaml under kind core", but docs/syu/features/core/core.yaml is already mutated with the new stub. Please make this path atomic (validate the registry conflict before mutating, or roll back on failure) and add a regression test.

tests/add_command.rs
```diff
@@ -0,0 +1,155 @@
...
+    let output = command.output().expect("init should run");
+    assert!(
+        output.status.success(),
+        "stdout:\n{}\nstderr:\n{}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    );
+}
+
+#[test]
```
tohboeh5, line 18: The happy-path coverage here is useful, but it still leaves the repo coverage gate red (99.29 percent locally) because none of the new failure/rollback paths are exercised. At minimum we need regression coverage for the registry-kind mismatch / no-partial-write case and one rejected --file path so the new validation branches in add.rs are actually protected.

tohboeh5:
Overview: The overall direction is good. The command fits the terminal-first workflow well, the default scaffolding paths feel sensible, and the docs/spec updates make the authoring story much easier to discover. I also exercised the main happy paths locally (syu init, syu add requirement, syu add feature --kind ...) and those flows behaved as advertised. The blockers are on failure handling and test completeness: there is a partial-write bug when a feature file is mutated before a registry-kind conflict is rejected, and the new error paths still leave the repository coverage gate red. Once those are addressed, this will be much closer to mergeable.

PR #220, opened 2026-04-12
fix: correct syu list help examples

### LINKED ISSUE #210: [Bug]: `syu list --help` shows examples that fail when used verbatim
ISSUE 1
### What happened?

`syu list --help` currently advertises these examples:

- `syu list docs/syu`
- `syu list requirement docs/syu`
- `syu list docs/syu requirement`

Those commands fail in a workspace created by `syu init`, because `docs/syu` is the spec root, not the workspace root that contains `syu.yaml`.

The current help text lives in `src/cli.rs` (`LIST_AFTER_HELP`).

The built in help examples should succeed when copied verbatim, or they should explicitly explain the difference between the workspace root and `spec.root`.

### syu version

### Environment

- OS: Linux
- Install method: local checkout, `cargo run --quiet -- ...`
- Repo: `ugoite/syu`

### Steps to reproduce

1. Run `cargo run --quiet -- init /tmp/syu-help-audit`.
2. Run `cargo run --quiet -- list --help` and copy one of the `docs/syu` examples.
3. Run `cargo run --quiet -- list /tmp/syu-help-audit/docs/syu`.

```text
error: missing philosophy directory `/tmp/syu-help-audit/docs/syu/docs/syu/philosophy`
```

tohboeh5:
## Summary
- replace the broken docs/syu list help examples with workspace-root examples
- explain that docs/syu is the default spec root inside the workspace
- add a regression test for the help text

Closes #210.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -58,7 +58,7 @@ Examples:
 
 Note:
   Pass the workspace root that contains syu.yaml.
-  docs/syu is the default spec.root inside that workspace.";
+  The configured spec.root lives inside that workspace; do not pass it directly.";
 
 const SEARCH_AFTER_HELP: &str = "\
 Examples:
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -51,10 +51,14 @@ const WORKSPACE_HELP: &str = "Workspace root containing syu.yaml and the configu
 const LIST_AFTER_HELP: &str = "\
 Examples:
   syu list
-  syu list docs/syu
+  syu list path/to/workspace
   syu list requirement
-  syu list requirement docs/syu
-  syu list docs/syu requirement";
+  syu list requirement path/to/workspace
+  syu list path/to/workspace requirement
+
+Note:
+  Pass the workspace root that contains syu.yaml.
+  docs/syu is the default spec.root inside that workspace.";
 
 const SEARCH_AFTER_HELP: &str = "\
 Examples:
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -51,10 +51,14 @@ const WORKSPACE_HELP: &str = "Workspace root containing syu.yaml and the configu
...
+  syu list path/to/workspace
   syu list requirement
-  syu list requirement docs/syu
-  syu list docs/syu requirement";
+  syu list requirement path/to/workspace
+  syu list path/to/workspace requirement
+
+Note:
+  Pass the workspace root that contains syu.yaml.
+  docs/syu is the default spec.root inside that workspace.";
```
tohboeh5: This note reintroduces the scaffold default into generic help text. `syu list` cares about the workspace root, not whether the repo happens to use `docs/syu`, so repositories initialized with a custom `--spec-root` will still read this as if `docs/syu` were part of the contract. Please reword the explanation around the configured spec root living under the workspace root instead of naming a specific path.

tohboeh5:
Overview: I like that this stops advertising obviously failing `syu list --help` examples and that the regression coverage now locks both argument orders in. The remaining problem is that the new explanatory note is still too repo-layout-specific, because it hard-codes `docs/syu` instead of describing the configured `spec.root` generically. Once that wording is made configuration-agnostic, this will read much better for real-world workspaces.

PR #221, opened 2026-04-12
docs: clarify quick-start validation timing

### LINKED ISSUE #211: [Bug]: Quick start examples imply `syu validate` will pass immediately after `syu add requirement`
ISSUE 2
### What happened?

The top level quick starts in both `README.md` and `docs/guide/getting-started.md` currently show this flow:

1. `syu init .`
2. `syu add requirement REQ-AUTH-001`
3. `syu validate .`

When I reproduced that flow in a fresh workspace, `syu validate` failed because the generated requirement stub is still orphaned until the user edits the reciprocal links.

`syu add` itself already prints the correct next step (`Edit the generated stub fields and reciprocal links`), so the high level docs are more optimistic than the command they describe.

The newcomer facing quick start should either stop after `syu add` and explicitly tell the user to edit links before validating, or use a full example that actually passes `syu validate`.

### syu version

### Environment

- OS: Linux
- Install method: local checkout, `cargo run --quiet -- ...`
- Repo: `ugoite/syu`

### Steps to reproduce

1. Run `cargo run --quiet -- init /tmp/syu-quickstart-audit`.
2. Run `cargo run --quiet -- add requirement REQ-AUTH-001 /tmp/syu-quickstart-audit`.
3. Run `cargo run --quiet -- validate /tmp/syu-quickstart-audit`.

```text
- [Error] SYU-graph-orphaned-001 ... requirement REQ-AUTH-001 is isolated from the layered graph
- [Warning] SYU-graph-links-001 ... Requirement does not link to any features
- [Warning] SYU-graph-links-001 ... Requirement does not link to any policies
```

tohboeh5:
## Summary
- stop implying that a newly added requirement stub validates immediately
- tell users to edit reciprocal links before running validate
- lock the guidance into the repository-quality docs test

Closes #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,11 @@ Understand the model first? Start with [`docs/guide/concepts.md`](docs/guide/con
 before you begin editing YAML.
 
 ```bash
-syu init .          # 1. Create spec scaffold
-syu add requirement REQ-AUTH-001  # 2. Add spec items with generated YAML stubs
-syu validate .      # 3. Check everything is linked
-syu app .           # 4. Browse in the browser
+syu init .                           # 1. Create spec scaffold
+syu add requirement REQ-AUTH-001     # 2. Generate a requirement stub
+# 3. Edit the new requirement and reciprocal links before validating
+syu validate .                       # 4. Check everything is linked
+syu app .                            # 5. Browse in the browser
 ```
 
 Use `syu browse .` when you want terminal-first exploration, or `syu app .`
--- a/README.md
+++ b/README.md
@@ -101,9 +101,16 @@ before you begin editing YAML.
 ```bash
 syu init .                           # 1. Create spec scaffold
 syu add requirement REQ-AUTH-001     # 2. Generate a requirement stub
-# 3. Edit the new requirement and reciprocal links before validating
-syu validate .                       # 4. Check everything is linked
-syu app .                            # 5. Browse in the browser
+```
+
+Before step 3, open the generated requirement YAML under `docs/syu/requirements/`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -99,10 +99,11 @@ Understand the model first? Start with [`docs/guide/concepts.md`](docs/guide/con
 before you begin editing YAML.
 
 ```bash
-syu init .          # 1. Create spec scaffold
-syu add requirement REQ-AUTH-001  # 2. Add spec items with generated YAML stubs
-syu validate .      # 3. Check everything is linked
-syu app .           # 4. Browse in the browser
+syu init .                           # 1. Create spec scaffold
+syu add requirement REQ-AUTH-001     # 2. Generate a requirement stub
+# 3. Edit the new requirement and reciprocal links before validating
```
tohboeh5: This is closer, but the quick start is still not fully actionable for a first-time user because the fix lives only as a shell comment inside the code block. The next executable command is still `syu validate .`, yet the reader still is not told which generated file to edit or what reciprocal links need to be added. A short prose step, or a tiny passing example, would close the gap much more effectively than an inline comment.

tohboeh5:
Overview: This is moving in the right direction because the quick start no longer outright promises that a freshly generated requirement stub will validate immediately, and the docs regression test is a good backstop. The part I still do not love is the UX of the fix: the key instruction only exists as a shell comment inside the code block, so a newcomer still does not see a concrete successful next step after `syu add requirement`. I would move that guidance into prose or a tiny passing example before calling it done.

PR #222, opened 2026-04-12
docs: clarify bug report version field

### LINKED ISSUE #212: [Bug]: The bug-report form requires a `syu version` even for docs and repository-automation defects
ISSUE 3
### What happened?

`.github/ISSUE_TEMPLATE/bug_report.yml` says:

> Please include a minimal reproduction and the exact `syu` version when possible.

But the same form marks the `syu version` input as `required: true`.

That is awkward for the very categories the form says it supports (`syu`, documentation, and repository automation), because a docs only issue or a workflow issue may not have a meaningful CLI version.

The bug report form should be internally consistent. A few reasonable options would be to make the version field optional, allow `not applicable`, or split docs and automation reports away from CLI runtime bugs.

### syu version

0.0.1-alpha.7 (current repo version; the form problem itself is template only)

### Environment

- OS: Linux
- Repo: `ugoite/syu`
- Affected file: `.github/ISSUE_TEMPLATE/bug_report.yml`

### Steps to reproduce

1. Open `.github/ISSUE_TEMPLATE/bug_report.yml`.
2. Read the introductory markdown block.
3. Compare it with the `version` field validation settings.

```yaml
Please include a minimal reproduction and the exact `syu` version when possible.

- type: input
  id: version
  validations:
    required: true
```

tohboeh5:
## Summary
- explain when reporters should provide an exact syu version
- explicitly allow `not applicable` for docs-only and automation defects
- keep the field required while making the form internally consistent

Closes #212.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -9,7 +9,9 @@ body:
     attributes:
       value: |
         Thanks for taking the time to improve `syu`.
-        Please include a minimal reproduction and the exact `syu` version when possible.
+        Please include a minimal reproduction.
+        If the report involves CLI behavior, include the exact `syu` version.
+        If the report is documentation-only or about repository automation, enter `not applicable` in the version field.
 
   - type: textarea
     id: summary
@@ -29,9 +31,9 @@ body:
   - type: input
     id: version
     attributes:
-      label: syu version
-      description: "Example: `0.0.1-alpha.7` or `v0.0.1-alpha.4`"
-      placeholder: 0.0.1-alpha.7
+      label: syu version (or `not applicable`)
+      description: "Example: `0.0.1-alpha.7`, `v0.0.1-alpha.4`, `main`, or `not applicable`"
+      placeholder: 0.0.1-alpha.7 or not applicable
     validations:
       required: true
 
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -29,9 +31,9 @@ body:
   - type: input
     id: version
     attributes:
-      label: syu version
-      description: "Example: `0.0.1-alpha.7` or `v0.0.1-alpha.4`"
-      placeholder: 0.0.1-alpha.7
+      label: syu version (or `not applicable`)
```
tohboeh5: Making `not applicable` explicit fixes the human-facing contradiction, but it also overloads the version field with two different meanings. After this change, triage and future automation can no longer distinguish a docs or repository-automation report from a CLI bug with an unknown version because both land in the same free-form input. A small scope selector, or a separate docs/automation bug form, would keep the form consistent without throwing away that structured signal.

tohboeh5:
Overview: The wording here is much clearer than the current contradiction between `when possible` and a required version field, so I agree with the direction. My concern is that the current fix makes the version input less structured by asking two different categories of report to share the same field. I would prefer a small scope choice or template split before relying on `not applicable` as part of the version value itself.

PR #223, opened 2026-04-12
docs: narrow app guide status examples

### LINKED ISSUE #213: [Feature]: Either document the `deprecated` status end to end or stop advertising it in the app guide
ISSUE 4

`docs/guide/app.md` currently tells users to expect a status badge of `planned`, `implemented`, or `deprecated`.

I could not find matching newcomer facing documentation for `deprecated` anywhere else in the repo. The main guides explain `planned` and `implemented`, but a repo wide search for `deprecated` only turned up the app guide after filtering out lockfiles.

That leaves a new user with an extra status word in the UI guide but no explanation of when to use it or what validation semantics it should carry.

### What change do you want?

Please pick one of these directions and make it consistent everywhere:

1. fully document `deprecated` across the README, guides, and self hosted spec, including when to use it and how validation treats it, or
2. remove the `deprecated` mention from the app guide until the rest of the product contract explains it.

### Specification impact

This likely touches `docs/guide/app.md`, the general status guidance in `README.md` and `docs/guide/getting-started.md`, and possibly the self hosted app or delivery spec files.

### Alternatives considered

Leaving the guide as is keeps an unexplained state in the user facing docs, which is the confusing part this issue is trying to remove.

tohboeh5:
## Summary
- stop advertising `deprecated` as a beginner-facing app guide status example
- describe the badge more generically as the YAML lifecycle state
- add a regression check in the repository docs test

Closes #213.

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -89,7 +89,7 @@ Below the sections panel, the documents for the active section are listed. Click
 Selecting an item shows:
 
 - **ID and title** — the stable identifier and human-readable name
-- **Status badge** — `planned`, `implemented`, or `deprecated`
+- **Status badge** — the lifecycle state from YAML (for example `planned` or `implemented`)
 - **Summary / description** — the prose from the YAML
 - **Links panel** — the upstream and downstream relationships (e.g. which requirements a feature satisfies; which policies a requirement enforces)
 - **Traces panel** — the declared test and implementation traces, with file path and symbol name
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -89,7 +89,7 @@ Below the sections panel, the documents for the active section are listed. Click
 Selecting an item shows:
 
 - **ID and title** — the stable identifier and human-readable name
-- **Status badge** — the lifecycle state from YAML (for example `planned` or `implemented`)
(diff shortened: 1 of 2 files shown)


docs/guide/app.md
```diff
@@ -89,7 +89,7 @@ Below the sections panel, the documents for the active section are listed. Click
 Selecting an item shows:
 
 - **ID and title** — the stable identifier and human-readable name
-- **Status badge** — `planned`, `implemented`, or `deprecated`
+- **Status badge** — the lifecycle state from YAML (for example `planned` or `implemented`)
```
tohboeh5: Removing `deprecated` helps, but `the lifecycle state from YAML` is still vague for a newcomer because it never names the field they should look for. Calling out the `status:` field explicitly, or linking to the status guidance elsewhere in the docs, would keep the guide self-serve instead of replacing one unexplained term with another.

tohboeh5:
Overview: Good direction overall. This removes the unexplained `deprecated` example and adds a regression check so the guide does not drift back. The remaining rough edge is that the replacement wording is still abstract from a new user perspective because it does not say which YAML field drives the badge. I would tighten that last sentence before merging.

PR #224, opened 2026-04-12
docs: prioritize verified installer flow

### LINKED ISSUE #214: [Feature]: Make the verified installer flow more prominent than `curl | bash` in the README
ISSUE 5

The README currently leads with `curl -fsSL ... | bash` installation examples, while checksum verification and attestation verification are documented later as optional follow up steps.

That structure nudges the shortest path toward executing a network fetched script before the user reaches the trust verification guidance.

### What change do you want?

Please make the trust preserving flow more prominent. For example:

- put a verify first example before the pipe to shell examples,
- add a strong callout that the checksum and attestation steps are the recommended path for security sensitive environments, or
- restructure the install section so the default path does not normalize skipping verification.

### Specification impact

This likely touches `README.md` and the self hosted installation and release documentation under `docs/syu/features/repository/install.yaml`.

### Alternatives considered

Keeping the current order is simple, but it means the most visible install path is also the least trustworthy one.

tohboeh5:
## Summary
- move the verify-first installer path ahead of the curl-pipe-bash shortcut
- add a security-sensitive environments callout and keep the shortcut as an explicit tradeoff
- add a regression test for the README ordering

Closes #214.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,35 @@ GHCR packages under `ghcr.io/ugoite/syu`. Download the installer from the
 current CLI release; the script prefers the matching package artifact and falls
 back to GitHub release assets if anonymous package pulls are not available yet.
 
+For security-sensitive environments, prefer the verified download flow below
+before running the installer script.
+
+### Recommended: verify before running
+
+Each release publishes a `checksums.sha256` file alongside the installer.
+Download both files, verify the checksum, then run the local copy:
+
+```bash
+RELEASE=v0.0.1-alpha.7
+curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/install-syu.sh" -o install-syu.sh
+curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/checksums.sha256" -o checksums.sha256
+sha256sum --ignore-missing -c checksums.sha256
+bash install-syu.sh
+```
+
+On macOS, replace `sha256sum` with `shasum -a 256`.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -41,6 +41,35 @@ GHCR packages under `ghcr.io/ugoite/syu`. Download the installer from the
...
+RELEASE=v0.0.1-alpha.7
+curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/install-syu.sh" -o install-syu.sh
+curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/checksums.sha256" -o checksums.sha256
+sha256sum --ignore-missing -c checksums.sha256
+bash install-syu.sh
+```
+
+On macOS, replace `sha256sum` with `shasum -a 256`.
+
+Published release archives also carry GitHub artifact attestations. Verify one with:
```
tohboeh5: Can we add one sentence clarifying that this attestation example is for the release archives, not for `install-syu.sh` itself? In the new verify-before-running section the checksum verifies the script, but this snippet suddenly switches to `syu-x86_64-unknown-linux-gnu.tar.gz`. Without that bridge text it is easy for a newcomer to think the attestation also covers the installer script they are about to execute.

tohboeh5:
Overview: I like the direction here a lot because the README finally leads with a verify-first path and the regression test locks that ordering in. The part I still want changed is the attestation note, because it jumps from verifying the installer script to verifying a different archive artifact without explaining the relationship. A small clarification would make the security guidance much harder to misread.

PR #225, opened 2026-04-12
feat: activate devcontainer pre-commit hooks

### LINKED ISSUE #215: [Feature]: Have the devcontainer finish pre-commit hook setup automatically
ISSUE 6

`.devcontainer/devcontainer.json` installs `pre-commit`, but it does not run `scripts/install-precommit.sh`, so a contributor opening the repo in the devcontainer or Codespaces still has to discover and run a second setup command before the hooks are actually active.

That weakens the open the devcontainer and start contributing story.

### What change do you want?

Please make the devcontainer activate the hooks automatically, or print an explicit post create instruction if that is intentionally left manual.

A simple fix would be to extend `postCreateCommand` so it also runs `scripts/install-precommit.sh` after installing the Python package.

### Specification impact

This likely touches `.devcontainer/devcontainer.json`, `scripts/install-precommit.sh`, and the contributor setup documentation and spec files.

### Alternatives considered

Leaving the install manual is workable, but it keeps a surprising extra step in the lowest friction onboarding environment.

tohboeh5:
## Summary
- have the devcontainer run the shared hook installer during post-create setup
- document that Codespaces/devcontainers finish hook activation automatically
- add repository-quality coverage for the new onboarding path

Closes #215.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -10,7 +10,7 @@
     },
     "ghcr.io/devcontainers/features/github-cli:1": {}
   },
-  "postCreateCommand": "cargo install cargo-llvm-cov --locked && python3 -m pip install --upgrade pip pre-commit",
+  "postCreateCommand": "cargo install cargo-llvm-cov --locked && bash scripts/install-precommit.sh",
   "customizations": {
     "vscode": {
       "extensions": [
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,9 @@ If you use the hooks, install them once:
 scripts/install-precommit.sh
 ```
 
+The devcontainer/Codespaces post-create step runs this script automatically so
+the hooks are active as soon as the environment finishes provisioning.
+
 ## Dependency security
 
 Dependency advisories are checked automatically on a weekly schedule (every Monday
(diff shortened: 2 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -10,7 +10,7 @@
     },
     "ghcr.io/devcontainers/features/github-cli:1": {}
   },
-  "postCreateCommand": "cargo install cargo-llvm-cov --locked && python3 -m pip install --upgrade pip pre-commit",
+  "postCreateCommand": "cargo install cargo-llvm-cov --locked && bash scripts/install-precommit.sh",
```
tohboeh5, line 13: This post-create path now relies on `scripts/install-precommit.sh`, but that script can install `pre-commit` with `python -m pip install --user` and then immediately execute the `pre-commit` binary. In a non-login devcontainer post-create shell, `~/.local/bin` is not guaranteed to be on PATH, so this can still fail before the hooks are installed. Calling the module through the resolved Python interpreter, or exporting the user bin dir, would make the automation reliable.

tohboeh5:
Overview: Nice improvement for contributor UX. Pushing hook activation into the devcontainer setup is exactly the kind of low-friction onboarding this repo wants. I do see one reliability gap, though: the automation still depends on the `pre-commit` executable being on PATH immediately after a `--user` install, which is brittle in post-create shells. I would harden that invocation path before merging.

PR #226, opened 2026-04-12
fix: surface app refresh failures in the browser UI

### LINKED ISSUE #216: [Bug]: `syu app` refresh failures are only surfaced in logs after the initial page load
ISSUE 7
### What happened?

After the browser app has loaded once, refresh failures are only logged to the console.

`app/src/App.tsx` sets a visible `error` state for the initial load, but the refresh path only calls `console.error(...)`. At the same time, `src/command/app.rs` flips the server into an error state and the endpoints start returning HTTP 500 responses.

I reproduced that by starting `syu app` against a temporary workspace, moving the workspace `docs/syu` directory away while the app was already running, and then observing both `GET /api/version` and `GET /api/app-data.json` return `500 Internal Server Error`.

The browser UI should show a visible degraded state banner or retry guidance when refresh and poll requests fail after the initial load, instead of only writing to the browser console.

### syu version

### Environment

- OS: Linux
- Install method: local checkout, `cargo run --quiet -- app ...`
- Repo: `ugoite/syu`

### Steps to reproduce

1. Run `cargo run --quiet -- init /tmp/syu-app-refresh-audit`.
2. Run `cargo run --quiet -- app /tmp/syu-app-refresh-audit --port 0`.
3. After the app reports ready, rename `/tmp/syu-app-refresh-audit/docs/syu`.
4. Query the running app again, or leave the browser tab open and wait for the next poll.

```text
# Server side
syu app refresh failed: failed to read spec root `/tmp/syu-app-refresh-audit/docs/syu` because it is not a directory

# HTTP responses
HTTP/1.1 500 Internal Server Error
app data refresh failed
```

tohboeh5:
## Summary
- keep the last successful syu app snapshot visible when live refresh fails
- show a visible degraded-state banner for version poll and workspace reload failures
- document and cover the behavior with Playwright while updating the embedded app bundle

Closes #216.


app/dist/index.html
```diff
@@ -9,8 +9,8 @@
       name="description"
       content="A local browser UI for exploring syu philosophy, policies, features, requirements, and validation issues."
     />
-    <script type="module" crossorigin src="/assets/index-sGlNQwrf.js"></script>
-    <link rel="stylesheet" crossorigin href="/assets/index-CoGKGUAh.css">
+    <script type="module" crossorigin src="/assets/index-49Yzh7qw.js"></script>
```
tohboeh5: I pulled the failing `browser-app` CI logs for this branch, and the breakage is the checked-in bundle itself: `scripts/ci/check-app-dist-freshness.sh` rebuilds `app/dist` to `index-BKDyx8-d.js` and `syu_app_wasm_bg-C58AZfzD.wasm`, while this branch commits `index-49Yzh7qw.js` and `xtraQ45A.wasm`. That means CI never reaches the new refresh-banner tests because the browser bundle is already stale. Please regenerate `app/dist` from this branch and commit the fresh artifacts.

tohboeh5:
Overview: The runtime UX change is going in the right direction because keeping the last good snapshot visible and surfacing a refresh banner is much better than silent console errors. Right now the branch is not mergeable as-is because the checked-in browser bundle is stale and the `browser-app` job fails before the new tests can validate the behavior. Once the dist output is regenerated from this branch, I want another pass on it.

PR #227, opened 2026-04-12
perf: skip app payload rebuilds when snapshots are unchanged

### LINKED ISSUE #217: [Feature]: Stop rebuilding the full `syu app` snapshot every second when nothing changed
ISSUE 8

The browser app server currently refreshes its full in memory snapshot every second, and the browser polls `/api/version` every two seconds.

That means `syu app` is continuously re reading YAML, hashing source documents, and rebuilding validation state even when nothing changed.

### What change do you want?

Please replace the unconditional polling and rebuild loop with a cheaper strategy such as:

- filesystem notifications,
- a lightweight change detector before full rebuilds,
- lazy refresh on demand, or
- a longer or backing off interval when the workspace is idle.

### Specification impact

This likely touches `src/command/app.rs`, `app/src/App.tsx`, and the self hosted browser app docs and spec.

### Alternatives considered

The current loop is simple and robust, but it will scale poorly as repositories and spec trees get larger.

tohboeh5:
## Summary
- split syu app refresh into a lightweight snapshot check and a conditional payload rebuild
- keep the current ready state untouched when the snapshot hash has not changed
- cover the unchanged-snapshot and payload-error branches so the coverage gate stays at 100%

Closes #217.

### DIFF
--- a/src/command/app.rs
+++ b/src/command/app.rs
@@ -835,6 +835,10 @@ mod tests {
         state
     }
 
+    fn failing_payload() -> anyhow::Result<AppPayload> {
+        Err(anyhow::anyhow!("payload load failed"))
+    }
+
     #[cfg(unix)]
     fn set_mode(path: &Path, mode: u32) {
         fs::set_permissions(path, fs::Permissions::from_mode(mode)).expect("permissions");
@@ -1575,13 +1579,7 @@ mod tests {
             )
             .expect("initial refresh should succeed");
         state
-            .refresh_current_with(
-                || Ok("same-snapshot".to_string()),
-                || {
-                    payload_loads.fetch_add(1, Ordering::SeqCst);
-                    Err(anyhow::anyhow!("payload load should be skipped"))
-                },
-            )
+            .refresh_current_with(|| Ok("same-snapshot".to_string()), failing_payload)
             .expect("unchanged snapshots should not rebuild the payload");
 
         assert_eq!(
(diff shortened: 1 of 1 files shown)


src/command/app.rs
```diff
@@ -96,6 +102,40 @@ impl AppState {
...
+        load_payload: LoadPayload,
+    ) -> Result<()>
+    where
+        LoadSnapshot: FnOnce() -> Result<String>,
+        LoadPayload: FnOnce() -> Result<AppPayload>,
+    {
+        let current_snapshot = self.current_snapshot()?;
+        let next = match load_snapshot() {
+            Ok(snapshot) => {
+                if current_snapshot.as_deref() == Some(snapshot.as_str()) {
```
tohboeh5, line 127: This early return only compares `spec_snapshot()`, but the payload also includes `validation_snapshot(collect_check_result(workspace_root))` further down the file. Validation can change when source or test files outside `docs/syu` move, even if the spec YAML is identical. With this skip condition those updates will never reach the browser until some spec document changes, so the optimization is currently dropping legitimate refreshes.

tohboeh5:
Overview: I like the idea here and the new tests show the intended happy path clearly. The problem is that the cache key is narrower than the actual app payload: you only hash spec documents, while the payload also depends on validation inputs from the wider workspace. As written, the app can serve stale validation data indefinitely after source and test changes, so I think the refresh key needs another pass before this can land.

PR #228, opened 2026-04-12
docs: clarify list vs browse usage

nightcityblade:
## Summary

Clarify when to use `syu list` versus `syu browse --non-interactive` across CLI help and the getting-started docs.

## How I tested this

- `cargo fmt --check`


src/cli.rs
```diff
@@ -49,12 +49,17 @@ Examples:
 const WORKSPACE_HELP: &str = "Workspace root containing syu.yaml and the configured spec tree";
 
 const LIST_AFTER_HELP: &str = "\
+Use `syu list` when you want a filtered layer view. Use `syu browse --non-interactive` when you want the full spec tree in text form.
+
 Examples:
   syu list
-  syu list docs/syu
   syu list requirement
+  syu list docs/syu/philosophy
```
tohboeh5: This new example is not runnable as written. On this branch, `cargo run -- list docs/syu/philosophy` fails with `missing philosophy directory .../docs/syu/philosophy` because a lone positional argument is parsed as the workspace path, not a spec subtree. Since this change is meant to clarify the UX for first-time users, shipping a broken example here would send them straight into an error. Please swap this to a command that `list` actually accepts, such as `syu list philosophy` or another valid kind + workspace example.

tohboeh5:
Overview: The direction is good: explicitly separating `syu list` from `syu browse --non-interactive` makes the CLI easier to learn, and adding the `browse` after-help text improves discoverability in the terminal itself. The remaining problem is that one of the new `list` examples is broken on this branch, so the help text currently teaches a command path that immediately errors. I left an inline comment on that example because it undercuts the UX goal of the change.

tohboeh5 (approved):
The list help now uses valid workspace-root examples, keeps both accepted positional orders, and explains not to pass spec.root directly.

PR #229, opened 2026-04-12
docs(app): surface search keyboard shortcuts

nightcityblade:
## Summary

Add a lightweight keyboard-shortcuts hint below the app search box and document the same shortcuts in the browser UI guide.

## How I tested this

- `cd app && npx vp check --fix src/App.tsx`
- `cd app && npx vp check src/App.tsx src/main.tsx src/types.ts src/index.css tests/app.spec.ts`


app/src/App.tsx
```diff
@@ -674,6 +674,10 @@ function App() {
                 className="w-full rounded-2xl border border-white/10 bg-slate-900/60 py-2 pl-9 pr-4 text-sm text-slate-100 placeholder-slate-500 focus:border-sky-400/60 focus:outline-none focus:ring-1 focus:ring-sky-400/40"
               />
             </div>
+            <p className="mt-2 text-xs text-slate-400">
+              Tip: use ArrowUp and ArrowDown to move through results, Enter to open the highlighted
```
tohboeh5, line 698: The new UI tip is close, but it now diverges from both the handler and the guide text. `onKeyDown` also opens the only result when there is exactly one match (lines 664-668), and the new docs section calls that out, but the in-app hint only mentions opening the highlighted item. Since this PR is specifically about surfacing shortcuts in the UI, leaving that case out means the browser hint still under-documents one of the supported Enter behaviors. Please either mention the single-match shortcut here too or trim the docs to keep both surfaces in sync.

tohboeh5:
Overview: The change is headed in the right direction: surfacing keyboard search shortcuts in the UI and documenting them in the guide both make the browser app easier to discover without trial and error. The remaining issue is that the new inline tip and the docs no longer describe exactly the same Enter behavior, so the UX guidance is already drifting across surfaces. I left an inline comment where the UI text should be tightened up so the browser and the guide stay aligned.

tohboeh5 (approved):
The Enter shortcut hint now matches the implemented single-result behavior and the guide text. I also added a Playwright assertion so the UI copy stays aligned.

PR #231, opened 2026-04-12
docs: clarify browse and list command selection

### LINKED ISSUE #219: [Feature]: Clarify when to use `syu browse --non-interactive` versus `syu list`
ISSUE 10

The CLI currently offers two text oriented read only surfaces:

- `syu browse --non-interactive`, which prints the whole spec tree, and
- `syu list`, which prints a filtered layer view.

Both are positioned as script friendly terminal commands, but the docs and help do not clearly explain when a newcomer should choose one over the other.

That confusion is amplified by the current `syu list --help` examples, which already need a workspace root clarification.

### What change do you want?

Please add an explicit command selection note in the CLI docs and help that explains:

- when to use `browse --non-interactive`,
- when to use `list`, and
- how they complement each other in scripts and ad hoc terminal use.

### Specification impact

This likely touches command help text, `README.md`, `docs/guide/getting-started.md`, and the self hosted CLI docs and spec.

### Alternatives considered

The current help text is short, but it leaves users to infer the product boundary between two similar read only commands.

tohboeh5:
## Summary
- clarify the command-selection guidance for syu list versus syu browse --non-interactive in CLI help, the README, and the getting-started guide
- align the self-hosted CLI feature specs and generated site-spec pages with the refined output-shape distinction
- add regression coverage for the new help text and documentation guidance

Closes #219


tohboeh5:
Overview:

I reviewed the changed files and did not find blocking issues in this diff. The command boundary is much clearer now: `syu list` is described as the automation-friendly, list-shaped surface, while `syu browse --non-interactive` is framed as the broader snapshot view with workspace context and current validation state.

I checked the updated help text and exercised the commands locally, including `browse --help`, `list --help`, `browse . --non-interactive`, `list requirement . --format json`, and the targeted regression tests added in this PR. The remaining red browser-app check appears to be outside the touched files here, so the doc/help changes themselves look solid.

PR #245, opened 2026-04-12
docs: clarify pinned installer examples

### LINKED ISSUE #235: [Bug]: Installer examples hardcode alpha release tags across the docs
## What area is affected?

## What happened?

Several installer examples hardcode `v0.0.1-alpha.7` in release URLs.

Examples:

- `README.md:53-57`
- `README.md:78-97`
- `docs/guide/getting-started.md:34-39`

That is correct for the current checkout, but it creates ongoing maintenance pressure and makes it easy for the docs to drift when the next prerelease is cut. The docs already recommend `SYU_VERSION=alpha` in some examples, so the current mix of pinned release URLs plus floating track selection is easy to misread.

I expected the installer documentation to use a maintainable pattern that stays correct across prereleases, or to explain more clearly why the examples are intentionally pinned.

For example, the docs could:

- promote a stable entrypoint pattern plus `SYU_VERSION=alpha` for the default path
- generate the pinned release tag automatically during release prep
- isolate pinned examples to a dedicated `install a specific version` section

## syu version

main

## Environment

- Linux development environment
- Repository checkout at `/workspaces/syu-1`
- Reviewed `README.md` and `docs/guide/getting-started.md`

## Steps to reproduce

1. Open the install examples in `README.md`.
2. Open the install example in `docs/guide/getting-started.md`.
3. Notice that the examples embed a concrete alpha tag in user-facing instructions.
4. Compare that with the separate `SYU_VERSION=alpha` guidance in the same README section.

```text
README.md:78
curl -fsSL https://github.com/ugoite/syu/releases/download/v0.0.1-alpha.7/install-syu.sh | bash

README.md:85
curl -fsSL https://github.com/ugoite/syu/releases/download/v0.0.1-alpha.7/install-syu.sh | env SYU_VERSION=alpha bash

getting-started.md:34
curl -fsSL https://github.com/ugoite/syu/releases/download/v0.0.1-alpha.7/install-syu.sh | env SYU_VERSION=alpha bash
```

## Before submitting

tohboeh5:
## Summary
- switch installer examples to a reusable RELEASE variable instead of repeating a hard-coded tag in every block
- explain why the URL stays pinned while SYU_VERSION selects a different published package track
- update repository-quality checks to keep the new explanation and example shape in place

Closes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,29 +122,38 @@ gh attestation verify syu-x86_64-unknown-linux-gnu.tar.gz --repo ugoite/syu
 If you already trust the release source and want the shortest path, use the
 one-line entrypoint:
 
+The download URL stays pinned to the installer shipped with this checked-in
+documentation version. Use `SYU_VERSION=alpha`, `stable`, or an explicit
+version selector when you want that installer to fetch a different published
+package track after it starts.
+
 Current installer entrypoint:
 
 ```bash
-curl -fsSL https://github.com/ugoite/syu/releases/download/v0.0.1-alpha.7/install-syu.sh | bash
+RELEASE=v0.0.1-alpha.7
+curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/install-syu.sh" | bash
 ```
 
 During the alpha phase, prefer the `alpha` track selector so the same installer
 entrypoint always resolves to the latest published alpha:
 
 ```bash
-curl -fsSL https://github.com/ugoite/syu/releases/download/v0.0.1-alpha.7/install-syu.sh | env SYU_VERSION=alpha bash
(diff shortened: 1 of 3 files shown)


tohboeh5 (commented):
Overview: This makes the installer examples easier to read by separating the checked-in release tag from the runtime `SYU_VERSION` selector, and the accompanying repository-quality assertions keep those examples synchronized with the docs. I did not find any substantive correctness, security, or UX issues in the diff.

tohboeh5 (commented):
Overview: The rebased branch preserves the pinned-release explanation while keeping the verified-install and Windows guidance current. I didn't find any additional correctness, security, or UX issues on the updated diff.

PR #246, opened 2026-04-12
docs: require concepts before quick start

### LINKED ISSUE #236: [Feature]: Make the README Quick start treat concepts as a required first step

The README Quick start still treats the concepts guide as an optional aside even though the next steps immediately depend on understanding the four-layer model.

Current flow:

- `README.md:106-109` says `Understand the model first?` and links to `docs/guide/concepts.md`.
- `README.md:111-123` then asks a newcomer to scaffold a requirement, edit `linked_policies`, edit `linked_features`, and update reciprocal links in other documents.
- The dedicated docs guide uses a much stronger signal (`docs/guide/getting-started.md:5-8`) by putting the concepts prerequisite in a visible tip before the workflow starts.

The README wording makes it sound like the concepts guide is optional, even though the Quick start depends on those concepts almost immediately.

## What change do you want?

Make the concepts guide an explicit prerequisite in the README Quick start.

Possible directions:

1. move the concepts link into a short `Before you start` block before the command snippet
2. mirror the docs-site phrasing more closely
3. explain that the next step assumes the reader understands reciprocal links and the four layers

## Specification impact

Documentation-only change, but it directly affects `FEAT-DOCS-001` because this is one of the primary first-run surfaces.

## Alternatives considered

- Keeping the current inline sentence: it is easy to skip because it sits between the heading and the first commands.
- Relying on validation errors to teach the model later: that pushes foundational concepts into a failure path.

## Before submitting

tohboeh5:
## Summary
- move the README concepts guidance into an explicit required pre-step ahead of Quick start
- explain that the next edits rely on understanding the four layers and reciprocal links
- lock the stronger README wording into repository-quality coverage

Closes #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,10 +162,11 @@ below. The rest of this section assumes you installed the published CLI.
 
 ## Quick start
 
-Step 0: required — run `syu init .` before any of the other commands in a new repository.
+Before you start: required — read [`docs/guide/concepts.md`](docs/guide/concepts.md)
+first. The next steps assume you already understand the four layers and the
+reciprocal links between policies, requirements, and features.
 
-Understand the model first? Start with [`docs/guide/concepts.md`](docs/guide/concepts.md)
-before you begin editing YAML.
+Step 0: required — run `syu init .` before any of the other commands in a new repository.
 
 ```bash
 syu init .                           # 1. Create spec scaffold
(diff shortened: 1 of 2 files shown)


tohboeh5 (commented):
Overview: Reframing the concepts guide as a required prerequisite is a solid newcomer-facing improvement. The new wording makes the dependency on the four-layer model explicit without adding much extra ceremony, and I did not find any substantive correctness or UX issues in the diff.

tohboeh5 (commented):
Overview: The rebased branch keeps the newcomer path chooser from the README while making the concepts-first prerequisite explicit in the updated quick-start copy. I didn't find any additional correctness, security, or UX issues on the refreshed diff.

Issue #238, opened 2026-04-12
[Feature]: Include actionable error text in syu browse --non-interactive


tohboeh5:
## What problem are you trying to solve?

`syu browse --non-interactive` is documented as the plain-text browse snapshot, but when the workspace is broken it only prints each error as `[CODE] subject`.

From a failing fixture workspace, the output currently looks like this:

```text
=== errors (5) ===
  [SYU-graph-reference-001] policy POL-FAIL-001
  [SYU-trace-id-001] feature FEAT-FAIL-001
  [SYU-trace-id-001] requirement REQ-FAIL-001
  [SYU-trace-symbol-003] feature FEAT-FAIL-001
  [SYU-trace-symbol-003] requirement REQ-FAIL-001
```

By contrast, `syu validate` prints the message, rule title, and suggestion for each issue. That makes `browse --non-interactive` much less actionable in logs or plain terminal review, even though `src/cli.rs` already positions it as a way to inspect `the current validation errors`.

## What change do you want?

Make the non-interactive browse error section more actionable.

Possible directions:

1. print the issue message next to each `[CODE] subject`
2. optionally include the suggestion as well
3. add a `--verbose-errors` mode if the default output must stay compact

## Specification impact

This likely touches `FEAT-BROWSE-002`, `REQ-CORE-015`, and the explainability goals in `POL-002`.

Evidence:

- `src/cli.rs:66-73`
- `src/command/browse.rs:353-376`
- `docs/syu/features/cli/browse.yaml:26-41`

## Alternatives considered

- Telling users to run `syu validate` separately: workable, but it weakens the value of the browse snapshot as a self-contained plain-text review tool.
- Keeping only codes and subjects: compact, but not enough to explain what to fix.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

tohboeh5:
This appears to already be fixed on `main`.

I re-ran `cargo run -- browse tests/fixtures/workspaces/failing --non-interactive`, and the current output includes the rule title/message plus `suggestion:` lines for each issue. The behavior is also covered by the existing regression checks in `src/command/browse.rs` (`non_interactive_issue_format_includes_message_and_suggestion`) and `tests/browse_command.rs` (`browse_command_non_interactive_shows_errors_when_workspace_is_failing`).

Closing this as already resolved in the current codebase.

PR #251, opened 2026-04-12
feat: make browse errors actionable

tohboeh5:
## Summary\n- include rule titles, messages, locations, and suggestions in non-interactive browse error output\n- cover the richer plain-text output with browse command tests and a formatting helper fallback test\n- update the self-hosted browse feature spec and generated docs to reflect the more actionable error snapshot\n\n## Testing\n- cargo test --test browse_command\n- cargo run --quiet -- browse tests/fixtures/workspaces/failing --non-interactive\n- scripts/ci/quality-gates.sh\n- scripts/ci/coverage.sh summary\n\nCloses #238

### DIFF
--- a/src/command/browse.rs
+++ b/src/command/browse.rs
@@ -718,4 +718,22 @@ mod tests {
             "    suggestion: Add symbol `missingTsSymbol` to `frontend/broken-feature.ts` or update the YAML mapping for `FEAT-FAIL-001`."
         );
     }
+
+    #[test]
+    fn non_interactive_issue_format_falls_back_when_rule_metadata_is_missing() {
+        let lines = format_non_interactive_issue(&Issue::warning(
+            "SYU-custom-unknown-001",
+            "workspace .",
+            None,
+            "Custom validation output should still stay readable.",
+            None,
+        ));
+
+        assert_eq!(
+            lines,
+            vec![
+                "  [SYU-custom-unknown-001] workspace .: Custom validation output should still stay readable."
+            ]
+        );
+    }
 }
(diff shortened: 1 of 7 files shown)


src/command/browse.rs
```diff
@@ -494,20 +496,49 @@ fn collapse_whitespace(value: &str) -> String {
     value.split_whitespace().collect::<Vec<_>>().join(" ")
 }
 
+fn format_non_interactive_issue(issue: &Issue) -> Vec<String> {
```
tohboeh5, line 502: This formatter now carries the same rule-title / message / suggestion shaping that `syu validate` already has in `render_text_report`. Because FEAT-BROWSE-002 is explicitly about making the non-interactive snapshot actionable, keeping a second formatter here means the two CLI surfaces can drift the next time someone tweaks issue wording or layout. Please consider extracting a shared issue-rendering helper so `browse --non-interactive` and `validate` stay aligned.

tohboeh5:
I checked the diff against #238, ran `cargo test --test browse_command`, and exercised `cargo run --quiet -- browse tests/fixtures/workspaces/failing --non-interactive` locally.

Overall this moves the UX in the right direction: the non-interactive snapshot is much more actionable now, and the spec/test updates match the changed output. The remaining concern is maintainability rather than behavior — the new formatter in `browse.rs` duplicates logic that `syu validate` already owns, so the two CLI surfaces can drift the next time issue wording evolves. I left one line comment on that point.

tohboeh5:
Addressed in f36c7f6 by extracting a shared internal issue-text helper under src/command/issue_text.rs so syu validate text output and syu browse --non-interactive now share the same rendering logic without expanding the public API surface.

Issue #237, opened 2026-04-12
[Feature]: Let syu add guide reciprocal links before the next validation run


tohboeh5:
## What problem are you trying to solve?

`syu add requirement` currently gives a concise stub, but the next obvious command still fails immediately in a brand-new workspace unless the user already understands how to wire adjacent links by hand.

I reproduced this locally with:

1. `cargo run -- init /tmp/syu-add-audit-Qmo0kq`
2. `cargo run -- add requirement REQ-AUTH-001 /tmp/syu-add-audit-Qmo0kq`
3. `cargo run -- validate /tmp/syu-add-audit-Qmo0kq`

The command succeeds, but the first validation run then fails with:

- `SYU-graph-orphaned-001`
- `SYU-graph-links-001` for missing policies
- `SYU-graph-links-001` for missing features

That is technically correct, but it means the newcomer path becomes "generate a
stub, then immediately hit a broken workspace" unless the user already knows
how to update multiple files.

This feels at odds with the low-ceremony promise behind `POL-004`,
`REQ-CORE-020`, and `FEAT-ADD-001`.

## What change do you want?

Add a more guided follow-up path for `syu add`, especially for requirements and
features.

Some possible directions:

1. extend `--interactive` so it can also prompt for adjacent links
2. print concrete follow-up commands after `syu add requirement` (for example, scaffold a matching feature or policy next)
3. optionally write placeholder adjacent links when the user explicitly asks for a guided flow

## Specification impact

This would likely touch `REQ-CORE-020`, `FEAT-ADD-001`, and the low-friction workflow policy in `POL-004`.

Evidence:

- `docs/syu/requirements/core/workspace.yaml:189-221`
- `docs/syu/features/cli/add.yaml:5-19`
- `README.md:116-120`
- local reproduction from `cargo run -- add requirement ...` followed by `cargo run -- validate ...`

## Alternatives considered

- Fixing this only in documentation: helpful, but it still leaves the first successful `add` flow ending in an immediate validation failure.
- Keeping `syu add` intentionally small forever: that preserves minimalism, but it pushes too much first-run ceremony onto hand-edited follow-up steps.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

PR #252, opened 2026-04-12
feat: guide add reciprocal links

tohboeh5:
## Summary\n- make syu add print layer-aware reciprocal-link follow-up steps before the next validation run\n- cover the new guidance in add/help tests and keep the output concise but specific\n- update the add feature/requirement specs, generated docs, CLI help, and README newcomer flow\n\n## Testing\n- cargo test --test add_command --test help_command\n- cargo run --quiet -- init "/workspace" && cargo run --quiet -- add requirement REQ-AUTH-001 "/workspace"\n- cargo run --quiet -- add feature FEAT-AUTH-LOGIN-001 "/workspace" --kind auth\n- scripts/ci/quality-gates.sh\n- scripts/ci/coverage.sh summary\n\nCloses #237

### DIFF
--- a/src/command/add.rs
+++ b/src/command/add.rs
@@ -102,6 +102,41 @@ struct ParsedId {
     file_slug: String,
 }
 
+#[derive(Debug, Clone, Copy)]
+struct ReciprocalLinkField {
+    yaml_key: &'static str,
+    linked_kind_label: &'static str,
+}
+
+const PHILOSOPHY_RECIPROCAL_LINK_FIELDS: &[ReciprocalLinkField] = &[ReciprocalLinkField {
+    yaml_key: "linked_policies",
+    linked_kind_label: "policy",
+}];
+const POLICY_RECIPROCAL_LINK_FIELDS: &[ReciprocalLinkField] = &[
+    ReciprocalLinkField {
+        yaml_key: "linked_philosophies",
+        linked_kind_label: "philosophy",
+    },
+    ReciprocalLinkField {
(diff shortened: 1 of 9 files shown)


src/command/add.rs
```diff
@@ -726,11 +726,65 @@ fn print_add_summary(
...
+    layer: LookupKind,
+    id: &str,
+    target: &Path,
+) -> Vec<String> {
+    let target_label = display_workspace_path(workspace, target);
+    let mut steps = vec![format!(
+        "Edit {target_label} and fill the stub fields for `{id}`."
+    )];
+
+    match layer {
```
tohboeh5: This `match` now hardcodes the reciprocal-link schema (`linked_policies`, `linked_requirements`, `linked_features`, `linked_philosophies`) in a second place. Because the same topology already exists in the document model and validation rules, a future schema rename can update validation while silently leaving `syu add` guidance lying to newcomers. Please consider deriving these labels from one shared mapping/helper so the CLI copy stays correct.

tohboeh5:
I checked the diff against #237, ran `cargo test --test add_command --test help_command`, and exercised the `init` + `add requirement` + `add feature` flow locally.

The newcomer UX is clearly better now: `syu add` no longer drops you into a dead-end after scaffolding a stub, and the README/spec/help updates keep the contract coherent. The remaining concern is maintainability rather than behavior — the new follow-up copy hardcodes the reciprocal-link schema in `add.rs`, so it can drift away from validation if those field names ever change. I left one line comment on that point.

PR #254, opened 2026-04-12
feat: guide syu add with adjacent scaffold suggestions

### LINKED ISSUE #237: [Feature]: Let syu add guide reciprocal links before the next validation run

tohboeh5:
## Summary
- add matching `syu add ...` suggestions to `syu add` follow-up output so adjacent stubs are easier to scaffold
- document the more guided add flow in the self-hosted spec, README, getting-started guide, and generated docs
- extend add/help/repository-quality coverage for the new guidance

## Testing

Closes #237

### DIFF
--- a/src/command/add.rs
+++ b/src/command/add.rs
@@ -669,6 +669,59 @@ fn format_instruction_list(items: &[String]) -> String {
     }
 }
 
+fn suggested_linked_kinds(layer: LookupKind) -> &'static [LookupKind] {
+    match layer {
+        LookupKind::Philosophy => &[LookupKind::Policy],
+        LookupKind::Policy => &[LookupKind::Philosophy, LookupKind::Requirement],
+        LookupKind::Requirement => &[LookupKind::Policy, LookupKind::Feature],
+        LookupKind::Feature => &[LookupKind::Requirement],
+    }
+}
+
+fn suggested_linked_id(id: &str, kind: LookupKind) -> String {
+    let mut segments = id.split('-').collect::<Vec<_>>();
+    segments[0] = match kind {
+        LookupKind::Philosophy => "PHIL",
+        LookupKind::Policy => "POL",
+        LookupKind::Requirement => "REQ",
+        LookupKind::Feature => "FEAT",
+    };
+    segments.join("-")
+}
+
+fn scaffold_missing_link_instruction(workspace: &Workspace, layer: LookupKind, id: &str) -> String {
+    let linked_kinds = suggested_linked_kinds(layer);
+    let linked_labels = linked_kinds
+        .iter()
+        .map(|kind| kind.label().to_string())
+        .collect::<Vec<_>>();
+    let workspace_arg = shell_quote_path(&workspace.root);
+    let commands = linked_kinds
+        .iter()
+        .map(|kind| {
+            format!(
+                "`syu add {} {} {workspace_arg}`",
+                kind.label(),
+                suggested_linked_id(id, *kind)
+            )
+        })
+        .collect::<Vec<_>>();
+
+    if linked_kinds.len() == 1 {
+        format!(
+            "If the linked {} stub does not exist yet, scaffold it with {}.",
+            linked_kinds[0].label(),
+            commands[0]
+        )
+    } else {
+        format!(
+            "If linked {} stubs are still missing, scaffold them with {}.",
+            format_instruction_list(&linked_labels),
+            format_instruction_list(&commands)
+        )
+    }
+}
+
 fn reciprocal_link_entry_instruction(layer: LookupKind, id: &str) -> String {
     let fields = reciprocal_link_fields(layer)
         .iter()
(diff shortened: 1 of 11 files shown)


src/command/add.rs
```diff
@@ -669,6 +669,59 @@ fn format_instruction_list(items: &[String]) -> String {
...
+        LookupKind::Philosophy => "PHIL",
+        LookupKind::Policy => "POL",
+        LookupKind::Requirement => "REQ",
+        LookupKind::Feature => "FEAT",
+    };
+    segments.join("-")
+}
+
+fn scaffold_missing_link_instruction(workspace: &Workspace, layer: LookupKind, id: &str) -> String {
+    let linked_kinds = suggested_linked_kinds(layer);
```
tohboeh5: Following the new guided flow can still lead to an immediate already exists error because this helper never checks the workspace before printing suggestions. Repro: in a fresh workspace, run syu add requirement REQ-AUTH-001, then the suggested syu add policy POL-AUTH-001; the second command now tells the user to run syu add requirement REQ-AUTH-001 even though that ID already exists. Since the updated requirement text now says to print scaffold commands when adjacent definitions do not exist yet, can we filter these suggestions through WorkspaceLookup and omit any ID that already resolves?

src/command/add.rs
```diff
@@ -669,6 +669,59 @@ fn format_instruction_list(items: &[String]) -> String {
...
+        LookupKind::Philosophy => &[LookupKind::Policy],
+        LookupKind::Policy => &[LookupKind::Philosophy, LookupKind::Requirement],
+        LookupKind::Requirement => &[LookupKind::Policy, LookupKind::Feature],
+        LookupKind::Feature => &[LookupKind::Requirement],
+    }
+}
+
+fn suggested_linked_id(id: &str, kind: LookupKind) -> String {
+    let mut segments = id.split('-').collect::<Vec<_>>();
+    segments[0] = match kind {
```
tohboeh5, line 685: The prefix-swap heuristic breaks down for philosophy suggestions. syu add policy POL-AUTH-001 currently recommends syu add philosophy PHIL-AUTH-001, but the starter workspace and the checked-in spec model philosophy IDs as project-level principles like PHIL-001 and PHIL-002, not topic-scoped companions of each policy. That nudges users toward inventing extra top-level philosophy stubs just to satisfy the suggestion. I would either special-case philosophy here or skip philosophy scaffold suggestions until we have a heuristic that maps policy IDs to an existing or expected philosophy.

tohboeh5 (commented):
Thanks for pushing this. The requirement-side guided flow is a real UX improvement, and I like that the spec/docs/tests were updated together so the behavior is documented consistently. I manually exercised the new syu add output locally and the first step is much clearer than main today. I left two review comments on the suggestion heuristics: the current chain can still point users at commands that immediately fail with already exists, and the philosophy suggestion is too naive to be a safe default. Once those are tightened up, the overall direction looks solid.

src/command/add.rs
tohboeh5: Handled in this branch by filtering suggestions through the workspace lookup and by limiting policy follow-up suggestions to requirement companions only. `add_command_skips_existing_or_unreliable_scaffold_suggestions` covers the duplicate-and-philosophy case now.

Issue #255, opened 2026-04-12
[Feature]: Clarify the README Quick start prerequisites so it matches the shortest-path promise


tohboeh5:
## What problem are you trying to solve?

The README says the Quick start path is the shortest path from install to syu validate ., but the actual Quick start section immediately requires reading docs/guide/concepts.md first.

That makes the headline promise misleading for newcomers. A reader who picks Quick start expects the lowest-ceremony path, but the document still hides a mandatory prerequisite before the first command. This feels out of step with PHIL-003 and POL-004, which both emphasize low-friction adoption.

Evidence:
- README.md lines 34-35 promise the shortest path from install to syu validate .
- README.md lines 165-167 require reading docs/guide/concepts.md before starting Quick start
- README.md lines 172-188 then present the actual first-run commands

## What change do you want?

Make the README match reality more clearly. Any of these would help:

1. rename Quick start so it no longer implies zero prerequisite reading
2. inline a minimal one-paragraph explanation of the four layers before the commands so readers can truly stay in the README
3. move the mandatory concepts prerequisite into the path chooser so newcomers know the cost before they choose Quick start

## Specification impact

This touches PHIL-003, POL-004, and FEAT-DOCS-001 because it changes how the repo presents the lowest-friction entry point.

## Alternatives considered

- Leaving the README as-is: this preserves the current wording, but the current wording is exactly what creates the mismatch.
- Sending all newcomers to the full getting-started guide: that would be clearer, but it gives up on having a true README-first path.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

PR #265, opened 2026-04-12

tohboeh5:
## Summary
- make the README path chooser warn that Quick start includes only a short layer refresher
- replace the mandatory concepts-guide detour with an inline four-layer refresher in Quick start
- update the repository quality assertion to match the new README-first wording

- Issue: #255
- Requirement / feature IDs: PHIL-003, POL-004, FEAT-DOCS-001

## Validation

## Release notes

- [x] No release note needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ verified.
 Pick the newcomer path that matches what you need next:
 
 - **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .`.
+  to `syu validate .` and only need a short layer refresher before the first
+  commands.
 - **Tutorial**: follow [`docs/guide/tutorial.md`](docs/guide/tutorial.md) when you
   want a realistic end-to-end repository story instead of a short scaffold flow.
 - **Troubleshooting**: jump to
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -162,9 +163,18 @@ below. The rest of this section assumes you installed the published CLI.
 
 ## Quick start
 
-Before you start: required — read [`docs/guide/concepts.md`](docs/guide/concepts.md)
-first. The next steps assume you already understand the four layers and the
-reciprocal links between policies, requirements, and features.
+Stay in this README for the shortest install-to-validate path. The four layers
```
tohboeh5: The repository overview already defines the four layers near the top of README. Re-defining them here with different wording and pluralization creates two glossaries in the same document, which can drift later and teach a slightly different model. Please either reuse the same wording from the overview or link back to that section instead of restating the terms.

README.md
```diff
@@ -162,9 +163,18 @@ below. The rest of this section assumes you installed the published CLI.
...
+are:
+
+- `philosophy`: the ideal state and values the project protects
+- `policy`: the repository-wide rules that make philosophy operational
+- `requirement`: a specific obligation that satisfies policy
+- `feature`: the implemented capability that satisfies a requirement
+
+Policies, requirements, and features should link to each other so
```
tohboeh5: Now that README no longer makes the concepts guide mandatory, the phrase link to each other is probably too abstract for the first manual-edit step. The quick start later asks readers to fill linked_policies, linked_features, and reciprocal linked_requirements fields, but this refresher never names those keys. A concrete pointer here would make the new prerequisite guidance much more actionable for first-time users.

tohboeh5:
Overview: I like the goal of making the README quick start more self-contained, and the chooser text is moving in the right direction. The main thing that still feels risky is that this section now carries a second, slightly different glossary of the four layers without yet translating the refresher into the actual YAML fields readers must edit. I would tighten that before merging so the shortest-path promise is clearer in practice, not only in tone.

Issue #257, opened 2026-04-12
[Bug]: Fix the tutorial feature-registry step so it does not imply a new entry when the walkthrough still uses core/core.yaml


tohboeh5:
## What area is affected?

Documentation or generated docs

## What happened?

The tutorial says that readers should open docs/syu/features/features.yaml and add the new file entry under files, but the walkthrough is still using the generated core feature document at docs/syu/features/core/core.yaml. The YAML block that follows only shows the existing core/core.yaml registry entry, so the text implies a manual registry edit even though the walkthrough has not actually introduced a second feature document.

That makes the tutorial feel more ceremonial than the code path really is, especially for first-time users who are trying to learn when the explicit feature registry must change.

## What did you expect to happen?

The tutorial should distinguish between these two cases:
- staying in the generated core/core.yaml file, where no registry change is needed because the starter entry already exists
- adding a second feature document later, where the reader really does need to append a new files entry

## syu version

main

## Environment

Docs review from the checked-out repository on Linux

## Steps to reproduce

1. Open docs/guide/tutorial.md
2. Read the feature section around the feature registry explanation
3. Notice that the narrative says to add a new files entry, but the example YAML only repeats the existing core/core.yaml starter entry
4. Compare that with the earlier tutorial steps, which never moved the feature into a new file

## Logs, screenshots, or failing output

Relevant lines:
- docs/guide/tutorial.md lines 193-204

## Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
Closing this because the current tutorial already distinguishes the two cases the issue asked for:

- when the walkthrough stays on `docs/syu/features/core/core.yaml`, the existing `features.yaml` entry is reused
- only later, when a second feature document is introduced, does the reader need to add another `files` entry

That wording is already in `docs/guide/tutorial.md` on `main`, so there is no remaining code or docs change to land here.

PR #267, opened 2026-04-12
docs: clarify tutorial feature registry step

tohboeh5:
## Summary
- explain that the tutorial keeps using the generated `core/core.yaml` feature document
- show that the starter feature registry entry already covers that walkthrough path
- reserve new `files` entries for the later case where a second feature document is introduced

- Issue: #257
- Requirement / feature IDs: PHIL-003, FEAT-DOCS-001

## Validation

## Release notes

- [x] No release note needed

### DIFF
--- a/docs/guide/tutorial.md
+++ b/docs/guide/tutorial.md
@@ -192,9 +192,8 @@ features:
 
 Because this walkthrough keeps using the generated
 `docs/syu/features/core/core.yaml`, the starter registry entry in
-`docs/syu/features/features.yaml` already points at the right file. Keep the
-generated metadata header from `syu init`, and make sure the `files` section
-still looks like this:
+`docs/syu/features/features.yaml` already points at the right file. For this
+tutorial step, just make sure the `files` section still looks like this:
 
 ```yaml
 files:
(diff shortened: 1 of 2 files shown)


docs/guide/tutorial.md
```diff
@@ -190,18 +190,21 @@ features:
       - REQ-STORE-001
 ```
 
-In `docs/syu/features/features.yaml`, keep the generated metadata header from
-`syu init` and add the new file entry under `files`:
+Because this walkthrough keeps using the generated
+`docs/syu/features/core/core.yaml`, the starter registry entry in
+`docs/syu/features/features.yaml` already points at the right file. Keep the
+generated metadata header from `syu init`, and make sure the `files` section
```
tohboeh5: The paragraph says to keep the generated metadata header, but the example only shows the files subsection. For a newcomer, that still leaves the key question open: should they preserve the version and metadata fields too, or only verify the existing registry entry? Please either show the full registry snippet or drop the header reference so the instruction and example match.

tohboeh5:
Overview: This clarification heads in the right direction, and I verified that a fresh scaffold already carries the core/core.yaml feature registry entry the tutorial is talking about. The remaining gap is that the new explanation mentions the generated metadata header while the snippet only shows the files section, so a first-time reader can still be unsure what exactly needs to stay unchanged.

Issue #258, opened 2026-04-12
[Feature]: Add four-layer model orientation to CONTRIBUTING before the GitHub Flow instructions


tohboeh5:
## What problem are you trying to solve?

CONTRIBUTING.md jumps straight into GitHub Flow and local checks, but it never tells a new contributor to understand the four-layer model first.

That means a contributor can start reading the workflow without learning why this repository separates philosophy, policy, requirements, and features, even though that separation is the core product idea. The README and getting-started guide both surface the conceptual model much earlier than CONTRIBUTING.md does.

Evidence:
- CONTRIBUTING.md lines 1-16 start with GitHub Flow and branch workflow
- README.md lines 19-28 explain the hierarchy of intent
- docs/guide/getting-started.md lines 5-8 explicitly tell new readers to read concepts first

## What change do you want?

Add a short orientation block near the top of CONTRIBUTING.md that:
1. points contributors at docs/guide/concepts.md
2. explains that the contribution workflow exists to keep philosophy, policy, requirements, and features connected to code and maintenance
3. clarifies that spec edits are part of normal feature work, not optional side documentation

## Specification impact

This touches PHIL-001, PHIL-003, POL-006, and FEAT-CONTRIB-002 because it changes how the self-hosted repo teaches contributors to use the layered model.

## Alternatives considered

- Leaving CONTRIBUTING.md as a pure workflow file: simple, but it keeps the conceptual prerequisite hidden from the contributors who most need it.
- Duplicating the full concepts guide inside CONTRIBUTING.md: too long; a short pointer and framing section would be enough.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

tohboeh5:
Closing this because the current `CONTRIBUTING.md` already adds the orientation block the issue asked for:

- it points contributors to `docs/guide/concepts.md` before workflow details
- it explains the four-layer model (`philosophy -> policies -> requirements -> features`)
- it says spec edits under `docs/syu/` are part of normal feature and bug-fix work

That guidance is already present on `main`, so there is no remaining change to implement here.

PR #268, opened 2026-04-12
docs: add contributor model orientation

tohboeh5:
## Summary
- add a short four-layer orientation section near the top of CONTRIBUTING.md
- point contributors at `docs/guide/concepts.md` before the GitHub Flow workflow
- clarify that spec edits under `docs/syu/` are normal feature and bug-fix work

- Issue: #258
- Requirement / feature IDs: PHIL-001, PHIL-003, POL-006, FEAT-CONTRIB-002

## Validation

## Release notes

- [x] No release note needed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,9 +7,11 @@ releaseable at all times.
 
 ## Start with the model
 
-Before you change code, docs, or release workflow, read
-[`docs/guide/concepts.md`](docs/guide/concepts.md). `syu` is built around the
-four-layer model: philosophy -> policy -> requirement -> feature.
+Before you change behavior, specs, contributor guidance, or release workflow,
+read [`docs/guide/concepts.md`](docs/guide/concepts.md). For small typo-only or
+mechanical CI fixes, you can start with the workflow below and come back to the
+guide when you need the fuller model. `syu` is built around the four-layer
+model: philosophy -> policies -> requirements -> features.
 
 The contribution workflow exists to keep those layers connected to real code,
 tests, docs, and maintenance work. In practice that means spec edits under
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -5,6 +5,17 @@
 `syu` uses GitHub Flow. `main` is the only long-lived branch and it should stay
 releaseable at all times.
 
+## Start with the model
+
+Before you change code, docs, or release workflow, read
```
tohboeh5: I like surfacing the model near the top, but making the concepts guide a blanket prerequisite for every change type adds ceremony even for narrow typo or CI fixes. That feels a bit at odds with the repository low-friction contributor story. Would it be better to frame this as required context for behavior and specification changes, while keeping drive-by fixes lightweight?

CONTRIBUTING.md
```diff
@@ -5,6 +5,17 @@
...
+[`docs/guide/concepts.md`](docs/guide/concepts.md). `syu` is built around the
+four-layer model: philosophy -> policy -> requirement -> feature.
```
tohboeh5: Everywhere else in the repo and UI the layers are labeled as policies, requirements, and features. Using the singular chain here makes the orientation text drift from the labels contributors immediately see in docs/syu, the browser app, and the CLI. Please keep the terminology aligned so the first mental model matches the actual navigation labels.

tohboeh5:
Overview: I like the addition of an explicit model-orientation block, especially the part that frames docs/syu edits as first-class work rather than optional documentation. My main concern is the blanket read-concepts-first phrasing for every change type, plus the terminology drift from the plural section labels used elsewhere. Tightening those two points would make this much more consistent with the rest of the contributor experience.

Issue #260, opened 2026-04-12
[Feature]: Make the devcontainer post-create setup self-explanatory for contributors


tohboeh5:
## What problem are you trying to solve?

The devcontainer postCreateCommand installs a large toolchain in one long shell line:
- cargo-llvm-cov
- wasm-pack
- app npm dependencies
- Playwright Chromium
- pre-commit hooks

A contributor can wait through that setup without learning why each tool is needed or which workflows it supports. The result is a ready environment, but not an understandable one.

Evidence:
- .devcontainer/devcontainer.json line 13 contains the full postCreateCommand in a single opaque command string
- CONTRIBUTING.md lines 106-115 says the devcontainer preinstalls these tools, but it still does not explain why each tool matters or which checks they unlock

## What change do you want?

Make the devcontainer setup explain itself.

For example:
1. move the setup into a named shell script under .devcontainer or scripts
2. document each installed tool and the check or workflow it supports
3. link back to CONTRIBUTING.md so contributors can map setup time to the gates they are expected to run

## Specification impact

This touches PHIL-002, PHIL-003, POL-006, and FEAT-CONTRIB-001 because the self-hosted repository is supposed to demonstrate a contributor-friendly polyglot environment, not only provision one.

## Alternatives considered

- Keeping the current one-line command: shortest implementation, but still opaque to contributors.
- Adding a comment in CONTRIBUTING.md only: helpful, but the devcontainer file itself would still hide the setup rationale.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

tohboeh5:
This is already resolved on main by #270 (commit 0d744ee). The devcontainer now delegates post-create work to .devcontainer/post-create.sh, logs why each tool is installed, and CONTRIBUTING.md maps the setup to the workflows it unlocks, so I’m closing this as completed.

PR #270, opened 2026-04-12
docs: explain devcontainer post-create setup

tohboeh5:
Summary:
- move the devcontainer post-create flow into a named shell script under .devcontainer
- make the setup log explain which local checks each installed tool unlocks
- update contributor docs, self-spec references, and repository quality assertions for the new script

- Issue: #260
- Requirement / feature IDs: PHIL-002, PHIL-003, POL-006, FEAT-CONTRIB-001

Validation:
- [x] cargo run -- validate .

Release notes:
- [x] No release note needed

### DIFF
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+# FEAT-CONTRIB-001
+
+set -euo pipefail
+
+repo_root() {
+  cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd
+}
+
+log_step() {
+  printf "\n[devcontainer] %s\n" "$1"
+}
+
+install_coverage_tooling() {
+  log_step "Installing cargo-llvm-cov for scripts/ci/coverage.sh summary."
+  cargo install cargo-llvm-cov --locked
+}
+
+install_wasm_tooling() {
+  log_step "Installing wasm-pack for scripts/ci/check-app-dist-freshness.sh."
+  cargo install wasm-pack --locked
+}
+
+install_browser_tooling() {
+  log_step "Installing browser-app dependencies for scripts/ci/check-app-dist-freshness.sh."
+  npm --prefix app ci
+
+  log_step "Installing Playwright Chromium for npm --prefix app run test:e2e."
+  npx --prefix app playwright install --with-deps chromium
+}
+
+install_precommit_tooling() {
+  log_step "Installing local hooks with scripts/install-precommit.sh."
+  bash scripts/install-precommit.sh
+}
+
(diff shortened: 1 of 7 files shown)


.devcontainer/post-create.sh
```diff
@@ -0,0 +1,51 @@
...
+  cargo install cargo-llvm-cov --locked
+}
+
+install_wasm_tooling() {
+  log_step "Installing wasm-pack for scripts/ci/check-app-dist-freshness.sh."
+  cargo install wasm-pack --locked
+}
+
+install_browser_tooling() {
+  log_step "Installing browser-app dependencies for scripts/ci/check-app-dist-freshness.sh."
```
tohboeh5: I reran `.devcontainer/post-create.sh` locally. This message still undersells what `npm --prefix app ci` unlocks: the same install is also a prerequisite for `npm --prefix app run test:e2e`, not only `scripts/ci/check-app-dist-freshness.sh`. As written, a contributor can still come away thinking the Playwright workflow depends only on the browser install step. Could we mention both workflows here so the setup log stays accurate?

CONTRIBUTING.md
```diff
@@ -109,10 +109,18 @@ If you use the hooks, install them once:
...
+`.devcontainer/post-create.sh` automatically so the setup explains itself while
+it provisions. That script:
+
+- installs `cargo-llvm-cov` for `scripts/ci/coverage.sh summary`
+- installs `wasm-pack` plus the `app/` dependencies for
+  `scripts/ci/check-app-dist-freshness.sh`
+- installs Playwright Chromium for `npm --prefix app run test:e2e`
+- runs `scripts/install-precommit.sh` so local hooks match the contributor path
+
+Read the script output or this section when you want to map setup time to the
```
tohboeh5, line 122: I reran the post-create script in the container. The output is clearer for coverage and browser-app checks, but it still never mentions that docs-site work remains manual because `npm --prefix website ci` is not part of the bootstrap. That makes "map setup time to the checks it unlocks" read broader than what the script actually covers. Please call out that website/docs-site tooling is still opt-in so contributors do not assume the container is ready for docs-site work too.

tohboeh5:
Overview: Moving the post-create flow into a named script is a good maintainability win, and rerunning `.devcontainer/post-create.sh` locally confirmed that the new log lines make the coverage and browser-app bootstrap much easier to follow than the old one-liner. The remaining gap is that the messaging still does not fully describe what the bootstrap covers: the app dependency install is worded as if it only unlocks bundle freshness, and the docs read broader than the actual setup because website/docs-site tooling is still manual. Once those two UX gaps are tightened, this will be a solid contributor-experience improvement.

Issue #261, opened 2026-04-12
[Bug]: Make the interactive syu browse menu follow the documented philosophy -> policy -> requirement -> feature order


tohboeh5:
## What area is affected?

CLI or runtime behavior

## What happened?

The interactive syu browse menu presents the top-level layers in this order:
1. philosophy
2. policy
3. feature
4. requirement

That puts feature before requirement, which conflicts with the documented hierarchy of intent and with syu browse --non-interactive, which prints requirement before feature.

This makes the interactive browser teach a different mental model than the README, tutorial, and non-interactive browse output.

## What did you expect to happen?

Interactive browse should follow the same hierarchy that the docs and non-interactive browse use:
philosophy -> policy -> requirement -> feature

## syu version

main

## Environment

Linux, local source checkout, cargo run CLI smoke tests

## Steps to reproduce

1. Run syu init on a temporary workspace
2. Run syu browse on that workspace
3. Observe the interactive menu order
4. Compare it with syu browse --non-interactive and the hierarchy described in README.md and docs/guide/tutorial.md

## Logs, screenshots, or failing output

Relevant evidence:
- src/command/browse.rs lines 104-124 print philosophy, policy, feature, requirement in the interactive menu
- src/command/browse.rs lines 356-364 print philosophy, policy, requirement, feature in non-interactive mode
- docs/guide/tutorial.md line 5 describes the model as philosophy -> policy -> requirement -> feature

## Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
This is already resolved on main. The interactive browse menu now uses philosophy -> policy -> requirement -> feature in src/command/browse.rs, and tests/browse_command.rs covers both the documented order and the selection mapping for options 3 and 4, so I’m closing this as completed.

PR #271, opened 2026-04-12
fix: reorder interactive browse layers

tohboeh5:
Summary:
- reorder the interactive browse top-level menu to philosophy, policy, requirement, feature, errors
- align the shared browse summary counts with the same documented hierarchy
- add browse command coverage that locks the interactive menu order in place

- Issue: #261
- Requirement / feature IDs: PHIL-003, POL-004, FEAT-CLI-BROWSE

Validation:
- [x] cargo run -- validate .

Release notes:
- [x] No release note needed

### DIFF
--- a/src/command/browse.rs
+++ b/src/command/browse.rs
@@ -23,8 +23,8 @@ use super::{
 enum TopLevelSection {
     Philosophy,
     Policy,
-    Feature,
     Requirement,
+    Feature,
     Errors,
 }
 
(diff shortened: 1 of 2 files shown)


src/command/browse.rs
```diff
@@ -105,8 +105,8 @@ impl BrowseState {
         const SECTIONS: [TopLevelSection; 5] = [
```
tohboeh5: The behavior is correct right now, but this PR still leaves two different canonical orders in the same file: `TopLevelSection` is declared earlier as philosophy -> policy -> feature -> requirement, while `SECTIONS` and the printed menu now use philosophy -> policy -> requirement -> feature. That makes the browse order harder to reason about for the next maintainer and easier to drift again later. Could we reorder the enum variants too, or derive the menu from a single ordered source?

tests/browse_command.rs
```diff
@@ -105,6 +105,40 @@ fn browse_command_walks_all_sections_in_a_passing_workspace() {
     assert!(stdout.contains("No validation issues are currently reported."));
 }
 
+#[test]
+fn browse_command_menu_uses_documented_layer_order() {
+    let output = Command::cargo_bin("syu")
+        .expect("binary should build")
+        .arg("browse")
+        .arg(fixture_path("passing"))
+        .write_stdin("0\n")
```
tohboeh5, line 114: This locks the printed order, but it does not prove that choosing `3` actually opens requirements and choosing `4` opens features. The risky regression in this PR is a display/mapping mismatch between the numbered labels and `SECTIONS`, and this test would still pass in that state because it exits immediately. Could we drive the menu through selections 3 and 4 and assert the resulting headings so the numeric dispatch is pinned down too?

tohboeh5:
Overview: I manually reran `cargo run -- browse tests/fixtures/workspaces/passing` and the user-facing menu now follows the documented philosophy -> policy -> requirement -> feature order, so the core UX change is landing in the right place. The two remaining review concerns are both about keeping that fix durable: the file still carries two different canonical orders (`TopLevelSection` vs. the new menu array), and the new test only locks the printed labels rather than the actual `3`/`4` dispatch behavior. Tightening those would make this much safer to maintain.

PR #272, opened 2026-04-12

tohboeh5:
Summary:
- clarify that the README quick start only depends on the in-page four-layer refresher
- inline the minimal layer chain before the first commands so the README path stays self-contained
- keep the concepts guide as an explicit deeper detour instead of an implied prerequisite

- Issue: #255
- Requirement / feature IDs: PHIL-003, POL-004, FEAT-DOCS-001

Validation:
- [x] cargo run -- validate .

Release notes:
- [x] No release note needed


README.md
```diff
@@ -163,18 +162,20 @@ below. The rest of this section assumes you installed the published CLI.
 
 ## Quick start
 
-Stay in this README for the shortest install-to-validate path. If you skipped
-[`docs/guide/concepts.md`](docs/guide/concepts.md), use the
-[Why four layers?](#why-four-layers) section above as the refresher on
-`philosophy`, `policy`, `requirements`, and `features`.
+Stay in this README for the shortest install-to-validate path. The only layer
+refresher you need before the commands is [Why four layers?](#why-four-layers):
```
tohboeh5: This paragraph re-spells all four layers immediately after the existing `Why four layers?` section. Because the README quick start is supposed to be the shortest install-to-validate path, this makes the section denser without really changing the prerequisite. I would keep this to a short pointer back to `Why four layers?` instead of repeating the layer definitions here.

tohboeh5:
I like the direction here: the quick-start prerequisites are more explicit, and the `concepts.md` detour reads more naturally than the old wording. The remaining issue is that the new refresher paragraph duplicates the layer definitions that are already introduced just above, so the "shortest path" section gets heavier instead of lighter. I left one inline comment on that duplication.

README.md
```diff
@@ -31,9 +31,8 @@ verified.
 
 Pick the newcomer path that matches what you need next:
 
-- **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .` and only need a short layer refresher before the first
-  commands.
+- **Quick start**: stay in this README when the in-page four-layer refresher is
```
tohboeh5: Calling the four-layer section a "refresher" still implies that Quick start readers were expected to learn the model somewhere else first. For the newcomer path this is often their *first* introduction, so the wording works against the goal of removing hidden prerequisites. I would switch this to something like "overview" or "introduction" here and in the quick-start lead-in below.

tohboeh5:
Overall impression: this improves the README in the right direction by making Quick start feel more self-contained. The remaining issue is wording: calling the four-layer section a “refresher” still suggests there was prerequisite reading somewhere else, which undercuts the stated goal of a lowest-ceremony path for first-time users. I left that inline.

tohboeh5:
Overview

Issue #256, opened 2026-04-12
[Feature]: Split the README quick start and getting-started guide into clearly different first-run paths


tohboeh5:
## What problem are you trying to solve?

The README Quick start flow and docs/guide/getting-started.md currently teach almost the same first-run workflow:
- syu init .
- syu add requirement REQ-AUTH-001
- hand-edit reciprocal links
- syu validate .
- syu app .

Because the two paths are so similar, newcomers do not get a clear answer to which one they should follow. This also increases maintenance cost because every workflow tweak has to be updated in two places.

Evidence:
- README.md lines 172-188 teach the five-step first run
- docs/guide/getting-started.md lines 19-35 teach the same flow with nearly identical commands and narrative
- README.md lines 30-40 points readers at multiple entry points but does not make the boundary between README and getting-started clear

## What change do you want?

Clarify the split between the README and getting-started guide.

For example:
1. keep the README as a compact reference card and move the full step-by-step narrative into getting-started
2. or make getting-started the canonical walkthrough and shorten the README to links plus a minimal command list
3. or explicitly document that the README is for readers who already know the four-layer model while getting-started is for first-time guided setup

## Specification impact

This is mainly a FEAT-DOCS-001 documentation-structure change, with product-shaping impact on PHIL-003 and POL-005 because it affects adoption friction and guide clarity.

## Alternatives considered

- Keeping both pages as they are: easiest in the short term, but it preserves both reader confusion and duplicated maintenance.
- Deleting one page entirely: simpler, but likely too extreme unless the remaining page is carefully expanded.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

PR #273, opened 2026-04-12
docs: distinguish README quick start from getting started

tohboeh5:
Summary:
- describe the README quick start as the compact first-run command card
- describe getting-started as the narrated first-run path that explains the manual link edits
- lock the new docs split in repository-quality assertions

- Issue: #256
- Requirement / feature IDs: PHIL-003, POL-005, FEAT-DOCS-001

Validation:
- [x] cargo run -- validate .

Release notes:
- [x] No release note needed


README.md
```diff
@@ -168,6 +168,10 @@ Stay in this README for the shortest install-to-validate path. If you skipped
 [Why four layers?](#why-four-layers) section above as the refresher on
 `philosophy`, `policy`, `requirements`, and `features`.
 
+Treat the rest of this section as a compact first-run command card. If you want
```
tohboeh5: This distinction arrives after the reader has already committed to the README path. The top-level `Choose your path` bullets still do not surface `docs/guide/getting-started.md`, so a first-time user will miss the fork you are introducing until they are already deep into the page. I would move the quick-start vs. narrated-guide split up to that entry chooser.

tohboeh5:
The split itself is a good idea: the README now points to a slower narrated path, and the getting-started guide explains how it differs from the compact quick start. The main gap is placement. Right now the new distinction shows up only after a reader has already chosen the README path, so the first decision point at the top of the README still sends newcomers down the same route. I left one inline comment on moving that fork to the chooser.

README.md
```diff
@@ -163,10 +166,9 @@ below. The rest of this section assumes you installed the published CLI.
 
 ## Quick start
 
-Stay in this README for the shortest install-to-validate path. If you skipped
-[`docs/guide/concepts.md`](docs/guide/concepts.md), use the
-[Why four layers?](#why-four-layers) section above as the refresher on
-`philosophy`, `policy`, `requirements`, and `features`.
+Stay in this README for the shortest install-to-validate path. If you want a
```
tohboeh5: The chooser now distinguishes Quick start from Getting started, but once a reader lands in the Quick start section itself there is still no pointer back to the narrated alternative. That matters because people often jump straight to `#quick-start` from links/search and will miss the new split entirely. I would add a short “want the step-by-step version instead?” sentence here that links to `docs/guide/getting-started.md`, so the distinction survives outside the top-of-page chooser.

tohboeh5:
Overall impression: the new README / guide split is directionally right, and the Getting started guide description makes the different reader journeys much clearer. The remaining issue is that the distinction disappears once someone lands directly in the Quick start section, because that section still does not point back to the narrated alternative. I left that inline.

tohboeh5:
Overview

Issue #262, opened 2026-04-12
[Bug]: Reorder the browser app sections so they match the documented four-layer hierarchy


tohboeh5:
## What area is affected?

CLI or runtime behavior

## What happened?

The browser app orders its top-level sections as philosophy, policies, features, requirements.

That ordering conflicts with both the product's documented hierarchy and the app's own onboarding copy, which says Philosophy -> Policies -> Requirements -> Features. The UI therefore teaches a different layer order than the rest of the product.

## What did you expect to happen?

The browser app should render top-level sections in the same order the rest of syu documents and explains:
philosophy -> policies -> requirements -> features

## syu version

main

## Environment

Linux, local source checkout, browser app code review and existing Playwright coverage review

## Steps to reproduce

1. Start the browser app with syu app on any workspace
2. Open the page and inspect the top-level section tabs
3. Notice that Features appears before Requirements
4. Compare that with the onboarding copy and docs, which both describe Requirements before Features

## Logs, screenshots, or failing output

Relevant evidence:
- app/src/App.tsx line 30 sets SECTION_ORDER to philosophy, policies, features, requirements
- app/src/App.tsx also contains onboarding copy that says Philosophy -> Policies -> Requirements -> Features
- app/tests/browser-app.spec.ts expects the Features tab before Requirements today, so the test suite has codified the mismatch

## Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
Closing as already fixed on `main`: the browser app now renders the top-level sections in the documented philosophy -> policies -> requirements -> features order, and `app/tests/browser-app.spec.ts` asserts that order.

PR #274, opened 2026-04-12
fix: reorder browser app section hierarchy

tohboeh5:
Summary:
- reorder the browser app top-level section list to philosophy, policies, requirements, features
- keep the rendered navigation order aligned with the onboarding copy and documented hierarchy
- update Playwright coverage and the checked-in app bundle for the new ordering

- Issue: #262
- Requirement / feature IDs: PHIL-003, POL-004, FEAT-APP-001

Validation:
- [x] cargo run -- validate .
- [x] scripts/ci/check-app-dist-freshness.sh

Release notes:
- [x] No release note needed


app/src/App.tsx
```diff
@@ -27,7 +27,7 @@ type SectionSummary = {
   itemCount: number;
 };
 
-const SECTION_ORDER: SectionKind[] = ["philosophy", "policies", "features", "requirements"];
+const SECTION_ORDER: SectionKind[] = ["philosophy", "policies", "requirements", "features"];
```
tohboeh5, line 36: Changing `SECTION_ORDER` also changes the checked-in browser bundle, but this PR only updates the existing `app/dist` filenames instead of regenerating them from the new source. `ci/browser-app` is already red because `scripts/ci/check-app-dist-freshness.sh` rebuilds to a different asset set. Please rebuild the app and commit the fresh `app/dist` output rather than hand-editing the checked-in bundle references.

tohboeh5:
Reordering the browser layers to match the documented hierarchy is the right change, and I like that the Playwright assertion now checks the order explicitly instead of only checking visibility. The blocker is the generated browser bundle: the checked-in `app/dist` output was not rebuilt from this source change, so `ci/browser-app` is failing and the embedded assets do not actually match `app/src/App.tsx`. I left one inline comment on that.

app/src/App.tsx
```diff
@@ -27,7 +27,7 @@ type SectionSummary = {
   itemCount: number;
```
tohboeh5, line 27: Changing `SECTION_ORDER` fixes the visible tab ordering, but the fallback selection path still uses `firstPopulatedSection()` / `workspace.sections.find(...)`, which depends on the payload order instead of this hierarchy. That means a fresh load can still land on whatever section the backend serialized first, so the app only partially adopts the documented order. If the goal is to make the browser hierarchy consistently philosophy → policies → requirements → features, the default-selection helpers should probably derive from the same constant too.

tohboeh5:
Overall impression: reordering the browser sections to match the documented four-layer hierarchy is the right UX direction, and the tab-order assertion in Playwright makes the intended surface clearer. The remaining concern is that the change only affects the rendered order; the default-selection fallback still uses the backend section order, so fresh loads can bypass the hierarchy this PR is trying to enforce. I left that inline.

tohboeh5:
Overview

Issue #263, opened 2026-04-12
[Bug]: Give the browser-app search UI proper accessible combobox/listbox semantics


tohboeh5:
## What area is affected?

Documentation or generated docs

## What happened?

The browser-app search box supports keyboard navigation visually, but it does not expose the usual combobox or listbox accessibility semantics.

The input has aria-describedby for shortcut help, but the search results container is just a div, the result rows are plain buttons, and the input does not expose aria-controls or aria-activedescendant. Screen-reader users therefore do not get the expected relationship between the search field and the highlighted result.

## What did you expect to happen?

The search experience should expose accessible search-result semantics, for example by wiring the input and result list together with combobox or listbox style attributes so assistive technologies can announce the active result.

## syu version

main

## Environment

Linux, local source checkout, browser app code review

## Steps to reproduce

1. Open app/src/App.tsx
2. Inspect the search field and result list markup around the search UI
3. Notice that the input only exposes aria-describedby and the result list has no listbox role or active-descendant wiring
4. Compare that with the keyboard behavior implemented in the same component

## Logs, screenshots, or failing output

Relevant evidence:
- app/src/App.tsx lines 661-748 implement keyboard navigation and the result list
- the input uses aria-describedby but not aria-controls or aria-activedescendant
- the result container is a div with id search-results-list rather than a listbox-like structure

## Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
Closing as already fixed on `main`: the browser-app search input now exposes combobox/listbox relationships (`role="listbox"`, `aria-controls`, and `aria-activedescendant`) and the Playwright suite covers that behavior.

PR #275, opened 2026-04-12
fix: add combobox semantics to app search

tohboeh5:
Summary:
- expose the browser-app search field as a combobox with listbox-backed results
- wire aria-controls, aria-expanded, aria-activedescendant, and aria-selected to the highlighted result
- update Playwright coverage and the checked-in browser bundle for the accessible search UI

- Issue: #263
- Requirement / feature IDs: PHIL-003, POL-004, FEAT-APP-001

Validation:
- [x] cargo run -- validate .
- [x] scripts/ci/check-app-dist-freshness.sh

Release notes:
- [x] No release note needed

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -27,7 +27,14 @@ type SectionSummary = {
   itemCount: number;
 };
 
+type SearchResult = {
+  id: string;
+  title: string;
+  kind: SectionKind;
+};
+
 const SECTION_ORDER: SectionKind[] = ["philosophy", "policies", "features", "requirements"];
+const SEARCH_RESULTS_LIST_ID = "spec-search-results-list";
 
 const SECTION_COPY: Record<SectionKind, string> = {
   philosophy: "Project intent and enduring values.",
@@ -40,6 +47,10 @@ const ONBOARDING_STORAGE_KEY = "syu-onboarding-dismissed";
 const SEARCH_RESULT_LIMIT = 20;
 const SEARCH_SHORTCUT_KEY_CLASS_NAME =
   "inline-flex items-center rounded-md border border-white/10 bg-white/5 px-1.5 py-0.5 font-mono text-[10px] text-slate-300";
+
+function searchResultOptionId(id: string, index: number) {
+  return `spec-search-result-${index}-${id.toLowerCase().replace(/[^a-z0-9_-]/g, "-")}`;
+}
 
 function App() {
   const [workspace, setWorkspace] = useState<BrowserWorkspace | null>(null);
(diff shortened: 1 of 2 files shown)


app/src/App.tsx
```diff
@@ -661,10 +680,18 @@ function App() {
               <input
                 id="spec-search"
                 type="search"
+                role="combobox"
+                aria-autocomplete="list"
                 aria-describedby="spec-search-shortcuts"
+                aria-controls={SEARCH_RESULTS_LIST_ID}
```
tohboeh5: The combobox always advertises `aria-controls="spec-search-results-list"`, but the listbox is only rendered when there are results. When the query is empty or there are no matches, the input points at a popup element that does not exist, which breaks the semantics this PR is trying to add. I would either keep a stable listbox container in the DOM or only set `aria-controls` when the listbox is actually rendered.

app/dist/index.html
```diff
@@ -9,7 +9,7 @@
       name="description"
       content="A local browser UI for exploring syu philosophy, policies, features, requirements, and validation issues."
     />
-    <script type="module" crossorigin src="/assets/index-DK0RrL4H.js"></script>
+    <script type="module" crossorigin src="/assets/index-WsPIO0eC.js"></script>
```
tohboeh5: The checked-in browser bundle is still stale. CI is rebuilding a different asset set from this source tree, so updating `app/dist/index.html` to a new hash without regenerating the rest of `app/dist` leaves both `ci/browser-app` and `ci/precommit` red. Please rebuild the app and commit the fresh bundle output.

tohboeh5:
Adding explicit combobox/listbox semantics is the right direction, and I like that the Playwright coverage now asserts `aria-activedescendant` and `aria-selected` instead of only searching by role. I left two blockers, though: the combobox now points `aria-controls` at a listbox that is sometimes absent, and the checked-in `app/dist` bundle was not regenerated from the source change, which is why both `ci/browser-app` and `ci/precommit` are red.

app/src/App.tsx
```diff
@@ -712,34 +741,47 @@ function App() {
...
+                    role="listbox"
+                    aria-label="Search results"
+                    className="space-y-1"
+                  >
+                    {searchResults.map((result, index) => (
+                      <div
+                        key={result.id}
+                        id={searchResultOptionId(result.id, index)}
+                        role="option"
+                        aria-selected={index === focusedResultIndex}
```
tohboeh5: `aria-selected` is being used to track the currently focused option, but in the combobox/listbox pattern that attribute communicates *selection*, not temporary keyboard focus. Screen readers will announce the active option as "selected" while users arrow through results, even though this UI does not keep a persistent selected value. The focus state is already conveyed by `aria-activedescendant`, so I would remove `aria-selected` here (or keep it `false` for every option).

tohboeh5:
Overall impression: the combobox work is a strong accessibility improvement. The role wiring, listbox linkage, and keyboard-coverage tests all move the search UI in the right direction. The one blocker I found is the `aria-selected` usage on the active option, which makes screen readers announce navigation focus as a persistent selection state; I left that inline.

tohboeh5:
Overview

Issue #278, opened 2026-04-12
[Bug]: Exclude generated build directories from symbol ownership and trace coverage checks


tohboeh5:
### What area is affected?
CLI or runtime behavior

### What happened?
Trace validation currently assumes repository files under source and test roots are candidates for ownership checks, while the repository also tracks generated browser output under app/dist. Generated bundles and wasm loader files are not meaningful implementation symbols, so scanning them for ownership or symbol existence creates a class of false positives and future adoption friction for any repository with frontend build output.

The current spec also makes ownership strict: REQ-CORE-002 requires optional full symbol ownership checks, and SYU-trace-id-001 requires traced files to mention their owning specification ID. That combination is reasonable for authored source files, but not for dist-like directories.

### What did you expect to happen?
Ownership and symbol coverage should ignore standard generated directories such as dist, build, coverage, and target by default, while still letting repositories opt in if they genuinely want to validate generated artifacts.

### syu version
main

### Environment
Repository audit on Linux in the main syu workspace.

### Steps to reproduce
1. Inspect REQ-CORE-002 in docs/syu/requirements/core/validation.yaml.
2. Inspect the SYU-trace-id-001 rule description in docs/syu/features/validation/validation.yaml.
3. Observe that the repository also tracks app/dist build output.
4. Consider enabling or tightening symbol ownership in a workspace that tracks generated bundles.

### Logs, screenshots, or failing output
```shell
git ls-files app/dist
app/dist/assets/index-C1DQWMlk.css
app/dist/assets/index-DK0RrL4H.js
app/dist/assets/syu_app_wasm-BLC3TU7T.js
app/dist/assets/syu_app_wasm_bg-C58AZfzD.wasm
app/dist/favicon.svg
app/dist/index.html
```

### Before submitting
- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

PR #287, opened 2026-04-12
fix: ignore generated coverage directories

tohboeh5:
Summary:
- skip common generated directories when strict symbol trace coverage inventories source and test trees
- add regression coverage for nested generated directories under src/
- update the validation spec and generated docs to document the default ignore list

- Issue: #278
- Requirement / feature IDs: REQ-CORE-002, FEAT-CHECK-001

Validation:
- [x] scripts/ci/coverage.sh summary
- [x] cargo run -- validate .

Release notes:
- [x] No release note needed


src/coverage.rs
```diff
@@ -588,6 +593,16 @@ fn collect_files_recursive_by_extension(
     Ok(())
 }
 
+fn should_skip_generated_directory(path: &Path) -> bool {
+    path.file_name()
+        .and_then(|name| name.to_str())
+        .is_some_and(|name| {
+            IGNORED_GENERATED_DIR_NAMES
```
tohboeh5: This basename-only skip rule is broad enough to hide authored code. A repository can legitimately keep real sources under directories like `src/build`, `src/target`, or `tests/coverage`, and after this change strict symbol coverage will silently stop inventorying them. Because the issue also calls out an opt-in path for validating generated artifacts, I think this needs either a configurable ignore list or a narrower heuristic than directory-name matching alone.

tohboeh5:
Skipping generated output directories is the right direction, and the local validation on this branch passed. The blocker is the breadth of the ignore rule: any authored directory literally named `build`, `target`, `dist`, or `coverage` now disappears from strict symbol coverage. I left one inline comment on making that behavior narrower or configurable before this lands.

tohboeh5:
Overall impression: this is a solid, well-scoped fix. The new config surface is documented clearly, the path-normalization rules are explicit about exact repository-relative matches, and the added coverage exercises both the ignore behavior and the “do not over-match nested authored code” case. I did not find a substantive blocker in the changed logic.

tohboeh5:
Overview

PR #288, opened 2026-04-12
feat: discover workspaces from child directories

### LINKED ISSUE #286: [Feature]: Let CLI commands discover syu.yaml by walking parent directories
The CLI currently teaches users to pass the workspace root that directly contains syu.yaml. That is explicit, but it also adds ceremony to everyday use from nested directories, editors, or test folders. Tools like git and cargo reduce this friction by walking parent directories until they find the repository root or manifest.

syu's current help text explicitly says: pass the workspace root that contains syu.yaml. That makes sense for correctness, but it is a usability tax for real repositories.

### What change do you want?
Make commands such as validate, browse, show, search, list, report, and app able to discover the workspace root by walking parent directories until they find syu.yaml.

The command should still allow an explicit workspace path to override discovery, and any discovery result should be shown clearly in output so users know which workspace was selected.

### Specification impact
This likely updates CLI requirements around low-ceremony usage and may need doc updates in getting-started and command help.

### Alternatives considered
Require explicit workspace paths forever. That keeps behavior simple for the implementation, but it pushes incidental directory knowledge onto the user in a way that conflicts with the low-ceremony philosophy.

### Before submitting

tohboeh5:
## Summary
- add shared workspace-root discovery that walks upward to the nearest syu.yaml or default workspace boundary
- update syu app startup output plus CLI/spec help so nested-directory invocation is explicit and low-ceremony
- add regression coverage for validate, browse, list, show, search, report, and app workspace discovery paths

## Testing
- cargo test --test workspace_discovery_command --test help_command --test list_command --test app_command
- cargo test resolve_workspace_root_
- cargo test load_workspace_discovers_parent_workspace_from_child_directory
- cargo test canonical_workspace_root_discovers_parent_workspace_config
- cargo run -- validate docs/syu/features
- cargo run -- list requirement docs/syu/features (sampled first 5 lines)
- bash scripts/ci/coverage.sh summary

Closes #286


src/cli.rs
```diff
@@ -63,8 +63,9 @@ Examples:
   syu list path/to/workspace requirement
 
 Note:
-  Pass the workspace root that contains syu.yaml.
-  The configured spec.root lives inside that workspace; do not pass it directly.";
+  Pass the workspace root or any child directory inside it.
+  syu walks upward until it finds syu.yaml, then resolves the configured spec.root from that workspace.
+  Do not pass the spec.root directory itself.";
```
tohboeh5: The new discovery behavior now accepts `spec.root` paths too, so this warning is no longer accurate. I verified locally that both `cargo run -- list requirement docs/syu` and `cargo run -- list requirement docs/syu/features` succeed on this branch. Keeping “Do not pass the spec.root directory itself.” here will push users away from a workflow the PR intentionally enabled.

tohboeh5:
The upward workspace discovery is a real UX improvement, and the regression coverage hits the right command surfaces. The remaining blocker is the CLI copy: `list --help` still tells users not to pass `spec.root`, but the new implementation deliberately accepts `docs/syu` and deeper paths. I left one inline comment on that mismatch.

src/workspace.rs
```diff
@@ -32,11 +32,30 @@ pub(crate) struct LoadedDocument<T> {
...
         .with_context(|| format!("failed to resolve workspace root `{}`", root.display()))?;
+    let search_root = if resolved.is_dir() {
+        resolved.clone()
+    } else {
+        resolved.parent().unwrap_or(&resolved).to_path_buf()
+    };
+
+    for candidate in search_root.ancestors() {
+        if candidate.join(CONFIG_FILE_NAME).is_file()
+            || resolve_spec_root(candidate, &SyuConfig::default()).is_dir()
```
tohboeh5: Using `docs/syu` as a second workspace marker changes discovery semantics more than the requirement text promises. Any ancestor that happens to contain an unrelated `docs/syu` tree can now win over the actual workspace the user is inside, so `syu` can resolve to the wrong repository root. Unless we can prove that `docs/syu` is unique and authoritative, this should probably keep using `syu.yaml` as the stop condition or validate a much stronger workspace shape before returning.

docs/syu/requirements/core/workspace.yaml
```diff
@@ -51,7 +51,10 @@ requirements:
       browser that shows philosophy, policy, feature, requirement, and error
       counts; allows drilling into linked definitions; and still explains the
       workspace when validation issues exist. When standard input/output are not
-      terminals, `syu` SHOULD fall back to help text instead of crashing.
+      terminals, `syu` SHOULD fall back to help text instead of crashing. When
+      the workspace argument (or default current directory) points inside a
+      workspace, `browse` MUST walk parent directories until it finds
```
tohboeh5, line 78: The requirement text still says these commands walk parent directories until they find `syu.yaml`, but the implementation now also stops at a bare `docs/syu` directory. That mismatch matters because this PR changes user-visible workspace resolution behavior. Please either tighten the implementation back to the documented contract or update the requirement text/tests so the spec stays truthful.

tohboeh5:
Overall impression: the user-facing goal is strong, and I like the effort to make commands work from nested directories instead of forcing people back to the repository root. The blocking concern is that the new discovery rule broadens the workspace marker from `syu.yaml` to any ancestor `docs/syu` directory, which can resolve to the wrong workspace and no longer matches the requirement text. I left inline comments on the implementation/spec mismatch and the false-positive root discovery risk.

tohboeh5:
Overview

The workspace-discovery direction is good and the parent-directory tests are valuable. The remaining issue is root-detection safety: the default-spec-root fallback is currently broad enough to stop at unrelated ancestors, which can send users into the wrong workspace without an obvious error. Once that guard only accepts real syu workspace roots, this will be much more trustworthy.

src/workspace.rs
```diff
@@ -32,11 +32,30 @@ pub(crate) struct LoadedDocument<T> {
...
+        {
+            return Ok(candidate.to_path_buf());
+        }
+    }
+
```
tohboeh5, line 53: This fallback is too broad: any ancestor with a plain `docs/syu` directory now counts as a workspace root, even when that directory belongs to some unrelated parent project. That can make a nested repo silently resolve the wrong workspace and then surface misleading results against the wrong spec tree. Please tighten the default-spec-root detection so it only accepts an actual syu workspace shape, not just a matching directory name.

PR #289, opened 2026-04-12
[REQ-CORE-009][FEAT-INIT-006][FEAT-INIT-007] Add interactive init flow for first-run setup

### LINKED ISSUE #285: [Feature]: Add interactive init for first-run adoption
PHIL-003 and POL-004 both say adoption should stay simple and low-ceremony, but the current onboarding path still assumes that a newcomer understands enough to choose a template, a spec root, naming prefixes, and the layered model before the first useful run.

The current CLI help is clear, but it is still command-driven. New users who do not yet understand philosophy, policy, requirement, and feature layering have to learn the model before they can confidently scaffold a repository.

### What change do you want?
Add an interactive init flow as a first-run guided experience.

The prompt sequence should help the user choose:
- repository shape or starter template
- preferred primary language
- spec root location
- ID prefix
- whether to enable stricter validation defaults immediately or later

The result should still produce plain checked-in files, not hidden state, so the workflow remains transparent and repository-native.

### Specification impact
This likely extends the init feature and the contributor/documentation surface, plus example coverage for the new guided flow.

### Alternatives considered
Keep init non-interactive and improve docs alone. Better docs help, but they still make the user translate a conceptual tutorial into CLI flags before first contact.

### Before submitting

tohboeh5:
## Summary

Add a guided `syu init --interactive` flow that prompts for project name, starter template, `spec.root`, shared ID stem, and stricter validation defaults before scaffolding the workspace.

This also extracts reusable prompt helpers, keeps `syu add` on the shared prompt path, updates the self-spec/docs/generated docs, and adds coverage so the interactive bootstrap stays validated.

- Issue: Closes #285
- Requirement / feature IDs: REQ-CORE-009, FEAT-INIT-006, FEAT-INIT-007

## Validation

## Release notes

- [ ] No release note needed


tohboeh5 (commented):
Temporary review auth check.

src/command/init.rs
```diff
@@ -57,15 +74,21 @@ impl StarterIdPrefixes {
...
-        .unwrap_or_else(|| infer_project_name(&workspace));
-    let spec_root = resolve_init_spec_root(args.spec_root.as_deref())?;
-    let id_prefixes = resolve_init_id_prefixes(args)?;
+    let mut prompt_io = StdioPromptIo;
+    run_init_command_with_prompt_io(args, &mut prompt_io)
+}
 
-    let files = scaffold_files(&project_name, &spec_root, args.template, &id_prefixes);
+fn run_init_command_with_prompt_io(args: &InitArgs, prompt_io: &mut impl PromptIo) -> Result<i32> {
+    let workspace = prepare_workspace_root(&args.workspace)?;
```
tohboeh5, line 87: Creating the workspace before validating that interactive prompting is allowed leaves side effects on a failing command. Repro: `cargo run -- init /tmp/pr289-non-tty-check --interactive </dev/null` exits with the non-TTY error, but still creates `/tmp/pr289-non-tty-check/`. Can we move the terminal check ahead of `prepare_workspace_root` (or otherwise defer filesystem writes) so rejected interactive invocations stay side-effect free?

src/command/prompt.rs
```diff
@@ -0,0 +1,162 @@
...
+        match default {
+            Some(value) => print!("{label} [{value}]: "),
+            None => print!("{label}: "),
+        }
+        io::stdout()
+            .flush()
+            .context("failed to flush interactive prompt")?;
+
+        let mut input = String::new();
+        io::stdin()
```
tohboeh5: EOF is treated as a successful empty response here. In a real terminal, pressing Ctrl+D makes `read_line` return `Ok(0)`, which causes callers like `prompt_required` and `prompt_bool` to loop forever printing retry messages instead of aborting the command. Please treat `bytes_read == 0` as an error so interactive commands can exit cleanly on EOF.

src/command/prompt.rs
```diff
@@ -0,0 +1,162 @@
...
+pub(crate) struct StdioPromptIo;
+
+impl PromptIo for StdioPromptIo {
+    fn is_terminal(&self) -> bool {
+        io::stdin().is_terminal() && io::stdout().is_terminal()
+    }
+
+    fn prompt_line(&mut self, label: &str, default: Option<&str>) -> Result<String> {
+        match default {
+            Some(value) => print!("{label} [{value}]: "),
```
tohboeh5, line 22: `syu init --interactive --format json` currently mixes prompt text and the final JSON payload on stdout, so the advertised JSON output is not actually machine-readable. I reproduced this locally and got the prompt lines plus the JSON object on the same stream. Could we send prompts to stderr, or reject `--interactive` together with `--format json`, so JSON mode keeps a clean contract?

tohboeh5:
Overall this is a strong UX improvement: the guided flow, reusable prompt helpers, and coverage make first-run onboarding much better. I left three inline blockers, though: the command mutates the filesystem before it knows interactive prompting is allowed, EOF during prompting can trap users in retry loops, and `--interactive --format json` no longer produces machine-readable JSON because the prompts are written to stdout. Once those are fixed, this should be close.

src/command/init.rs
```diff
@@ -133,11 +164,120 @@ pub fn run_init_command(args: &InitArgs) -> Result<i32> {
...
+        return Ok(ResolvedInitOptions {
+            project_name,
+            spec_root,
+            template,
+            id_prefixes,
+            strict_validate_defaults: false,
+        });
+    }
+
+    validate_interactive_init_mode(args, prompt_io)?;
```
tohboeh5: This `validate_interactive_init_mode()` call is redundant because `run_init_command_with_prompt_io()` already performs the same validation before it calls `resolve_init_options_with_prompt_io()`. In interactive mode this repeats the TTY / JSON-format checks on the same path, and in non-interactive mode the helper short-circuits anyway. Dropping the second call would keep the control flow easier to follow without changing behavior.

tohboeh5:
Overall impression: this is a substantial and mostly well-executed UX improvement. The prompt flow is discoverable, the docs/test coverage are unusually thorough, and the command fits the project’s “lower the barrier to entry” direction well. The only issue I found in the changed path is a redundant interactive-mode validation call that makes the control flow harder to reason about than it needs to be; I left that inline.

tohboeh5:
Overview

Issue #277, opened 2026-04-12
[Bug]: Stop checking in app/dist and build the browser app in CI instead


tohboeh5:
### What area is affected?
Repository automation or CI

### What happened?
The repository currently tracks the browser application's built output under app/dist. Those files are derived artifacts, not source of truth, so every rebuild risks noisy diffs, accidental stale assets, and merge churn. This also weakens trust in release and preview workflows because contributors can change shipped frontend output without forcing CI to reproduce it.

Concrete evidence from the current repository:
- git ls-files lists app/dist/assets/index-C1DQWMlk.css
- git ls-files lists app/dist/assets/index-DK0RrL4H.js
- git ls-files lists app/dist/assets/syu_app_wasm_bg-C58AZfzD.wasm
- app/.gitignore does not ignore dist/
- the repository already has scripts/ci/check-app-dist-freshness.sh, which shows the project is aware that dist freshness is a repository concern

### What did you expect to happen?
The repository should treat app source files and wasm source files as the only reviewable inputs. CI should build app/dist for previews, release packaging, or deployment, and no checked-in dist tree should be required on main.

### syu version
main

### Environment
Repository audit on Linux in the main syu workspace. The problem is repository state and automation policy rather than a host-specific runtime defect.

### Steps to reproduce
1. Clone the repository.
2. Run git ls-files app/dist.
3. Observe that built CSS, JS, HTML, and wasm artifacts are tracked.
4. Open app/.gitignore and observe that dist/ is not ignored.

### Logs, screenshots, or failing output
```shell
git ls-files app/dist
app/dist/assets/index-C1DQWMlk.css
app/dist/assets/index-DK0RrL4H.js
app/dist/assets/syu_app_wasm-BLC3TU7T.js
app/dist/assets/syu_app_wasm_bg-C58AZfzD.wasm
app/dist/favicon.svg
app/dist/index.html
```

### Before submitting
- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
Closing as already fixed on `main`: `app/dist` is no longer tracked in git, so the repository no longer checks in the browser app build output.

PR #290, opened 2026-04-12
fix: build browser app bundle in CI

tohboeh5:
Summary:
- stop checking the browser app bundle into git and ignore app/dist
- build an embedded browser bundle during Rust builds while keeping CI and release workflows Node-aware
- replace the stale dist freshness flow with a browser app validation flow that refreshes wasm bindings, rebuilds the bundle, and keeps docs/spec coverage aligned

- Issue: #277
- Requirement / feature IDs: REQ-CORE-017, FEAT-APP-001, FEAT-QUALITY-001

Validation:
- [x] bash scripts/ci/check-browser-app-freshness.sh

Release notes:
- [x] No release note needed


.github/workflows/ci.yml
```diff
@@ -257,18 +250,14 @@ jobs:
...
-        working-directory: app
-        run: npm ci
-
-      - name: Verify checked-in browser bundle
-        run: scripts/ci/check-app-dist-freshness.sh
+      - name: Build browser app bundle
+        run: scripts/ci/check-browser-app-freshness.sh
 
       - name: Upload browser app dist
         if: always()
```
tohboeh5: Non-blocking, but this makes CI failures harder to triage: when `scripts/ci/check-browser-app-freshness.sh` exits before producing `app/dist`, this unconditional upload step adds a second `No files were found with the provided path: app/dist` error on top of the real failure. I hit that in the current browser-app log. Gating this on `success()` (or switching `if-no-files-found` away from `error`) would keep the primary failure readable.

build.rs
```diff
@@ -0,0 +1,141 @@
...
+    println!("cargo:rerun-if-changed=build.rs");
+    emit_watch(&app_dir.join("index.html"));
+    emit_watch(&app_dir.join("package.json"));
+    emit_watch(&app_dir.join("package-lock.json"));
+    emit_watch(&app_dir.join("vite.config.ts"));
+    emit_watch(&app_dir.join("tsconfig.json"));
+    emit_watch(&app_dir.join("tsconfig.app.json"));
+    emit_watch(&app_dir.join("tsconfig.node.json"));
+    emit_watch_recursive(&app_dir.join("public"));
+    emit_watch_recursive(&app_dir.join("src"));
```
tohboeh5, line 27: Blocking once the bootstrap issue is fixed: the embedded bundle now depends on `app/wasm/**` and the wasm build script, but the rerun set only watches `app/src/**` plus the Vite config files. Changing the wasm crate or `app/scripts/build-wasm.mjs` will not rerun this build script, so Cargo can reuse a stale embedded app even though the shipped wasm bridge should have changed. Please add `cargo:rerun-if-changed` coverage for the wasm source tree/build script (or generate the bridge inside this build step) so the embedded assets stay in sync.

tohboeh5:
Overall impression: the direction is good. Moving the checked-in browser bundle out of git and teaching CI to rebuild it is the right cleanup, and I like the attempt to centralize the Node setup plus keep a dedicated browser-app freshness script. The current branch still has a blocking bootstrap gap, though: every clean-checkout Cargo build now depends on ignored `app/src/wasm` outputs that are never generated before `build.rs` runs, so precommit, quality, MSRV, and installed-binary smoke all fail from the same root cause. I also left one follow-up note about the `cargo:rerun-if-changed` coverage and one CI ergonomics note around the unconditional artifact upload.

build.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    if let Err(error) =
+        ensure_app_dependencies(&app_dir).and_then(|_| build_browser_bundle(&app_dir, &out_dir))
```
tohboeh5: Blocking: this makes every clean-checkout cargo build fail before any test runs. App.tsx still imports the generated wasm bridge from app/src/wasm, but that tree is entirely gitignored and this build path never runs npm run build:wasm. I reproduced it locally by deleting app/src/wasm and running cargo build; Vite failed to resolve the generated wasm bridge, which matches the current precommit, quality, MSRV, and installed-binary failures. Please generate the bridge here (or stop this path from depending on ignored generated files) before moving the browser bundle into build.rs.

tohboeh5:
Overview

PR #291, opened 2026-04-13
feat: add trace-aware git history lookup (REQ-CORE-024, FEAT-LOG-001)

### LINKED ISSUE #284: [Feature]: Add Git-aware history commands for traced requirements and features
syu can validate that a requirement or feature points at current repository evidence, but it cannot help users inspect the change history around that evidence. That makes maintenance and review weaker than they need to be: once a requirement is linked to files and symbols, teams naturally want to ask which commits changed that implementation, whether a requirement has been stable, and which spec IDs are most active.

This is especially important for repository-native traceability. If the repository is the source of truth, Git history is part of the story.

### What change do you want?
Add Git-aware commands that project specification IDs onto repository history.

A good starting scope would be:
- syu log <id> to show commits touching files or symbols traced from one requirement or feature
- optional filtering by layer kind or path subset
- summaries that explain why each commit was matched
- machine-readable output for audits and release tooling

This can grow later into diff or blame-oriented workflows, but the first step should make history navigable from the current trace graph.

### Specification impact
This likely needs a new requirement and CLI feature. It may also need policy text clarifying that repository history is a valid explainability surface when it can be derived from checked-in traces.

### Alternatives considered
Ask users to manually run git log over traced paths after reading the YAML. That is possible, but it is exactly the kind of repeated manual stitching that a repository-native specification tool should absorb.

### Before submitting

tohboeh5:
## Summary

Add a new `syu log` command that projects requirement and feature traces onto Git history, including text/JSON output, kind/path filters, matched-reason summaries, and self-hosted spec/docs updates.

- Issue: Closes #284
- Requirement / feature IDs: REQ-CORE-024, FEAT-LOG-001

## Validation

## Release notes

- [ ] No release note needed


src/command/log.rs
```diff
@@ -0,0 +1,743 @@
...
+        .collect::<BTreeSet<_>>();
+    if unique_paths.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    let mut command = Command::new("git");
+    command.arg("-C").arg(workspace_root).args([
+        "log",
+        "--relative",
+        "--no-renames",
```
tohboeh5: Using `--no-renames` here means `syu log` stops showing history as soon as a traced file gets renamed or moved. That is a pretty common maintenance path for implementation/test files, so the command will under-report the history exactly in the refactor-heavy cases where people are likely to rely on it. Please either follow each tracked path with `git log --follow` where possible, or add a deliberate fallback/documented limitation instead of silently dropping rename history.

tohboeh5 (commented):
The command shape, error handling, and test coverage here are strong, and the text/JSON outputs feel consistent with the rest of the CLI. I found one history-correctness gap that is worth fixing before merge: rename/move tracking currently gets dropped, which can make long-lived traced files look newer than they really are.

src/command/log.rs
```diff
@@ -0,0 +1,743 @@
...
+
+fn matched_reasons(changed_files: &[String], tracked_paths: &[TrackedPath]) -> Vec<TrackedPath> {
+    let changed = changed_files
+        .iter()
+        .map(String::as_str)
+        .collect::<BTreeSet<_>>();
+
+    tracked_paths
+        .iter()
+        .filter(|tracked| changed.contains(tracked.path.as_str()))
```
tohboeh5: matched_reasons does an exact string match against git log --name-only output. That breaks loadable workspaces with non-canonical trace paths like ./src/../src/history_test.rs: the traced file has commits, but syu log ends up returning no matching commit reasons because the tracked path never normalizes to the name-only output. Please normalize traced paths and/or git output before matching.

src/command/log.rs
```diff
@@ -0,0 +1,743 @@
...
+}
+
+fn build_history_target(
+    lookup: WorkspaceLookup<'_>,
+    workspace_root: &Path,
+    id: &str,
+    kind: HistoryKind,
+    path_filter: Option<&Path>,
+) -> Result<HistoryTarget> {
+    let Some(entity) = lookup.find(id) else {
```
tohboeh5: syu log currently resolves duplicate IDs by taking the first match from lookup.find() and document_path_for_id(). I reproduced two REQ-DUP-001 definitions in different files and the command silently showed only the first copy's title, path, and history. Since duplicate IDs are a validation issue that can still load, this should reject ambiguous IDs (or force disambiguation) instead of guessing.

tohboeh5 (commented):
Nice feature overall: the command is well-scoped, the help/docs/tests are strong, and the happy path feels easy to follow. The remaining gaps are edge-case robustness when the workspace still loads with validation issues: non-canonical trace paths currently lose history, and duplicate IDs are handled ambiguously instead of surfacing that ambiguity to the caller.

src/command/log.rs
```diff
@@ -0,0 +1,743 @@
...
+            .with_context(|| {
+                format!(
+                    "path filter `{}` must stay inside workspace `{}`",
+                    path.display(),
+                    workspace_root.display()
+                )
+            });
+    }
+
+    Ok(path.to_path_buf())
```
tohboeh5: Relative path filters are passed through unchanged here, so `syu log FEAT-APP-001 . --kind implementation --path ./app/src` currently rejects the selection even though `--path app/src` succeeds on the same branch. Because CLI users commonly prefix local paths with `./`, this should normalize dot segments before the later `starts_with` match (or compare normalized relative paths) instead of treating the two spellings as different filters.

tohboeh5:
Overall impression: the command shape is promising, the help/docs/test surface is thorough, and the JSON/text outputs feel consistent with the rest of the CLI. After reviewing the branch and exercising `syu log` locally, I still see four blockers before merge: rename history is still dropped because the git invocation uses `--no-renames`; non-canonical traced paths can miss matches against `git log --name-only`; duplicate IDs are still resolved by guessing the first match instead of surfacing ambiguity; and `--path` does not normalize `./` segments, so `--path ./app/src` currently fails while `--path app/src` succeeds on the same branch. Once those correctness and UX gaps are closed, this will be a very useful trace-driven history feature.

src/command/log.rs
```diff
@@ -0,0 +1,976 @@
...
+    commits.extend(remainder);
+    commits.truncate(limit);
+    Ok(commits)
+}
+
+fn repository_history_order(
+    workspace_root: &Path,
+    candidate_shas: &BTreeSet<String>,
+) -> Result<Vec<String>> {
+    let output = git_command(workspace_root)
```
tohboeh5: `git rev-list HEAD` reads the entire repository history into memory on every `syu log` invocation, even though this code only needs to order a small candidate set. On larger repositories that makes the command scale with total commit count instead of the selected paths; please bound the walk to the candidate SHAs or avoid the extra full-history pass.

tohboeh5:
Overview

Solid feature and good test/doc coverage. The main weakness is performance: the current history ordering scans the whole repository on each invocation, which will hurt as repositories grow. I left one blocking inline comment on that path.

PR #292, opened 2026-04-13
feat: add relation graph command (REQ-CORE-023, FEAT-RELATE-001)

### LINKED ISSUE #283: [Feature]: Add a syu relate command for cross-layer impact and consistency inspection
The current CLI can list, search, browse, and show individual items, but it does not provide one focused command for asking relational questions such as:
- what policies and requirements justify this feature?
- what features and source symbols are downstream of this requirement?
- if I pick one function or one requirement, what else should I inspect to judge whether the spec is coherent?

That gap makes it harder to validate whether the current specification is sensible, not just syntactically valid. It also leaves a hole in the workflow for reviewers who want impact analysis without manually stitching together show, search, and file reads.

### What change do you want?
Add a syu relate command that accepts a requirement ID, feature ID, policy ID, philosophy ID, file path, or source symbol and returns the relevant connected graph in a review-friendly form.

At minimum, it should be able to:
- walk upstream and downstream across philosophy, policy, requirement, and feature layers
- include traced files and symbols when the selected node is implemented
- surface missing links or suspicious gaps that may indicate a weak specification
- emit both text and JSON output for humans and automation

### Specification impact
This likely needs a new CLI feature definition plus requirement text describing relation traversal and explainability expectations.

### Alternatives considered
Use syu show, syu search, and manual grep as a workflow. That works for small cases, but it is not a coherent command for impact analysis or spec-validity review.

### Before submitting

tohboeh5:
## Summary
- add the syu relate command to inspect the connected philosophy, policy, requirement, and feature graph from an ID, path, or trace symbol
- surface adjacent traceability gaps and support both text and JSON output for review workflows
- document the command in the self-hosted spec, generated site docs, and README

Closes #283


src/command/relate.rs
```diff
@@ -0,0 +1,1356 @@
...
+
+    Ok(normalize_relative_path(path).display().to_string())
+}
+
+fn normalize_relative_path(path: &Path) -> PathBuf {
+    let mut normalized = PathBuf::new();
+    for component in path.components() {
+        match component {
+            Component::CurDir => {}
+            other => normalized.push(other.as_os_str()),
```
tohboeh5: `normalize_relative_path` strips `.` segments but still lets `Component::ParentDir` pass through unchanged. That means selectors like `../../some/file.rs` are accepted as path-mode input and then compared literally against traced paths, which makes the command behave like it accepts workspace-escaping selectors even though they cannot resolve correctly. Please either reject `..` components explicitly here or canonicalize the selector against the workspace root so invalid path selectors fail fast with a clear error.

tohboeh5 (commented):
Nice addition overall: the selector model is intuitive, the output shape is easy to consume from both text and JSON, and the tests cover the main happy/error paths well. I found one correctness/UX issue in path normalization that should be tightened before merge. Once that is addressed, this looks like a strong fit for the spec-tracing workflow.

tohboeh5 (commented):
Nice addition overall: the command is well-documented, the CLI integration is clean, and the new tests make the happy path easy to follow. The remaining issues are both about selection semantics: the graph walk currently expands far beyond the selected neighborhood, and root-level path lookup depends on the file still existing on disk, which weakens relate right when users are investigating broken traces.

src/command/relate.rs
```diff
@@ -0,0 +1,1356 @@
...
+        }
+    }
+    normalized
+}
+
+fn expand_related_ids(workspace: &Workspace, mut related: RelatedIds) -> RelatedIds {
```
tohboeh5: Because expand_related_ids keeps re-traversing every shared policy/philosophy edge, cargo run --quiet -- relate run_search_command . --format json on this branch balloons to 21 requirements and 25 features, including unrelated siblings like REQ-CORE-001 and FEAT-ADD-001. That makes the output noisy and makes traversal cost scale with the whole graph instead of the selected neighborhood. Please keep relate scoped to the directly related subgraph.

src/command/relate.rs
```diff
@@ -0,0 +1,1356 @@
...
+        related_ids,
+    })
+}
+
+fn is_path_like(workspace_root: &Path, selector: &str) -> bool {
+    selector.contains(std::path::MAIN_SEPARATOR)
+        || selector.contains('/')
+        || selector.contains('\\')
+        || Path::new(selector).is_absolute()
+        || workspace_root.join(selector).exists()
```
tohboeh5: This path classification depends on workspace_root.join(selector).exists(), so a traced root-level file like README.md stops being selectable as soon as the file is missing. On this branch, syu relate README.md <workspace> works before deleting the traced file, then exits 2 with a did-not-match error after deleting it even though the spec still references that path. Path-vs-symbol detection should come from the trace graph, not current filesystem existence.

src/command/relate.rs
tohboeh5: I reproduced this against the repository on the branch: `cargo run --quiet -- relate run_search_command . --format json` returns 21 requirements and 25 features, which is far wider than the directly linked neighborhood around `run_search_command`. That makes the command hard to trust as an impact-inspection tool, because unrelated siblings dominate the output. Please keep the expansion scoped to the selected item direct relation graph instead of computing the full transitive closure through every shared policy/philosophy edge.

tohboeh5:
Overall impression: the command concept is strong, the selector model is easy to understand, and the text/JSON output format fits the rest of the CLI well. After reviewing the branch and exercising `syu relate` locally, I still see three blockers before merge: parent-directory path selectors are accepted instead of failing fast with a clear workspace-boundary error; root-level path lookup still depends on the file existing on disk instead of the trace graph; and the relation walk expands far beyond the directly related neighborhood (for example, `cargo run --quiet -- relate run_search_command . --format json` currently returns 21 requirements and 25 features). Once the selector semantics are tightened and the graph traversal is scoped down, this will be a very useful impact-inspection command.

src/command/relate.rs
```diff
@@ -0,0 +1,1465 @@
...
+        },
+        direct_matches: DirectMatches {
+            definitions: Vec::new(),
+            traces,
+        },
+        related_ids,
+    })
+}
+
+fn is_path_like(selector: &str) -> bool {
```
tohboeh5, line 281: This heuristic misses valid repo-relative paths like `LICENSE`, `Dockerfile`, and other single-segment files without separators or extensions. Those selectors fall through to symbol lookup and fail even though they are legitimate paths; please classify existing repo-relative files as paths before applying symbol heuristics.

tohboeh5:
Overview

Nice end-to-end command wiring, docs, and tests, but the current path detection is too narrow and misclassifies some valid repository paths. I left one blocking inline comment on that selector heuristic.

PR #293, opened 2026-04-13
feat: add Go trace validation support

### LINKED ISSUE #280: [Feature]: Add Go language support to trace validation and symbol ownership
PHIL-002 says the model should fit repositories regardless of primary programming language, but the built-in language adapters are still narrow. The current implementation in src/language.rs covers Rust, Python, TypeScript/JavaScript, Shell, YAML, JSON, Markdown, and Gitignore, while REQ-CORE-002 only promises trace verification for Rust, Python, and TypeScript/JavaScript.

That leaves basic Go repositories outside the first-class traceability path even though Go is a common target for small, CLI-heavy, repository-native tools.

### What change do you want?
Add a Go language adapter for trace validation and optional symbol ownership checks.

The first version should support at least:
- function detection
- method detection
- type and interface detection
- test symbol discovery for standard Go test naming
- canonical language aliases such as go

### Specification impact
REQ-CORE-002 should expand the supported-language contract, and tests should include a minimal passing Go fixture that exercises implementation traces and test traces.

### Alternatives considered
Ask Go repositories to label their code as plain text or shell-like files. That would discard the language-aware validation that gives syu credibility.

### Before submitting

tohboeh5:
## Summary
- add a first-class Go language adapter for trace validation and extend strict ownership coverage to exported Go symbols and standard Go tests
- extend the passing multi-language fixture, CLI validation tests, and strict coverage fixtures to exercise Go traces end to end
- update the self-hosted validation requirement, config docs, generated site docs, and report to describe Go support

Closes #280


src/coverage.rs
```diff
@@ -547,6 +558,146 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+            ))
+    });
+    targets.dedup();
+
+    Ok(targets)
+}
+
+fn collect_go_public_symbols(contents: &str) -> Vec<String> {
+    let regex = Regex::new(
+        r"(?m)^\s*(?:func\s+(?:\([^)]*\)\s*)?(?P<function>[A-Z][A-Za-z0-9_]*)\s*\(|type\s+(?P<type>[A-Z][A-Za-z0-9_]*)\b)",
```
tohboeh5: collect_go_public_symbols still only matches func and type forms, so strict Go coverage misses other exported API shapes that the Go adapter accepts. I reproduced a false pass with require_symbol_trace_coverage=true by leaving exported var/const declarations uncovered, and the same regex also skips generic functions like func GenericAPI[T any](...). Please widen the inventory here so SYU-coverage-public-001 fires for the full public Go surface.

src/coverage.rs
```diff
@@ -547,6 +558,146 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+            captures
+                .name("function")
+                .or_else(|| captures.name("type"))
+                .map(|symbol| symbol.as_str().to_string())
+        })
+        .collect()
+}
+
+fn collect_go_test_symbols(contents: &str) -> Vec<String> {
+    let regex = Regex::new(r"(?m)^\s*func\s+(?P<name>Test[A-Za-z0-9_]*)\s*\(")
```
tohboeh5: This only inventories Go test functions whose names start with Test. Go also executes Benchmark..., Fuzz..., and Example... entry points from *_test.go files, so strict requirement coverage currently lets those go unowned. I reproduced a false pass by adding an uncovered BenchmarkUncovered to a Go test file; syu validate still passed.

tohboeh5 (commented):
Nice addition overall: the PR wires Go into trace-reference validation, updates the docs/spec fixtures, and adds focused happy-path tests. The remaining problem is that the new Go coverage inventory is still narrower than real Go surface area, so require_symbol_trace_coverage can report false passes for exported APIs and executable test entry points that are not matched by the current regexes.

src/coverage.rs
tohboeh5: I reproduced this with a minimal Go workspace on the branch: `syu validate` only reported the uncovered exported `type PublicThing`, while uncovered exported `var ExportedConfig`, `const ExportedFlag`, and `func GenericAPI[T any]()` all passed without `SYU-coverage-public-001`. That means the Go coverage inventory is still narrower than the public surface that the Go adapter accepts, so strict coverage can report false passes for real APIs.

src/coverage.rs
tohboeh5: The same minimal workspace also let uncovered `BenchmarkUncovered`, `FuzzUncovered`, and `ExampleUncovered` entry points pass cleanly as long as one `Test...` function was traced in the file. Because Go treats all four prefixes as executable test surface, strict requirement coverage is still reporting false passes for uncovered behavior here.

tohboeh5:
Overall impression: the Go adapter is a valuable addition, the spec/docs wiring is coherent, and the new validation tests make the intended happy path easy to follow. After reviewing the branch and exercising `syu validate` against a minimal Go workspace, the remaining blocker is still coverage breadth: strict coverage currently misses exported `var`/`const` declarations, generic exported functions, and non-`Test...` executable test entry points such as `Benchmark...`, `Fuzz...`, and `Example...`. Because those gaps produce false passes under `require_symbol_trace_coverage`, I do not think the branch is ready to merge until the Go inventory matches the full executable/public surface that the Go adapter claims to support.

src/coverage.rs
```diff
@@ -547,6 +558,148 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+            targets.push(CoverageTarget {
+                file: relative.clone(),
+                symbol,
+                kind: CoverageTargetKind::PublicSymbol,
+            });
+        }
+    }
+
+    for path in &test_files {
+        let contents = match fs::read_to_string(path) {
```
tohboeh5: Please do not skip Go files that fail to read. Other discovery paths surface unreadable files as validation issues; swallowing the error here can hide missing symbols and produce a false clean run.

src/coverage.rs
```diff
@@ -547,6 +558,148 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+            ))
+    });
+    targets.dedup();
+
+    Ok(targets)
+}
+
+fn collect_go_public_symbols(contents: &str) -> Vec<String> {
+    let regex = Regex::new(
+        r"(?m)^\s*(?:func\s+(?:\([^)]*\)\s*)?(?P<function>[A-Z][A-Za-z0-9_]*)\s*(?:\[[^\n\r\]]+\])?\s*\(|type\s+(?P<type>[A-Z][A-Za-z0-9_]*)\b|(?:const|var)\s+(?P<value>[A-Z][A-Za-z0-9_]*)\b)",
```
tohboeh5: This regex only matches single-name `const`/`var` declarations. Idiomatic Go also uses grouped `const (...)` and `var (...)` blocks, and the same limitation exists in `src/language.rs`, so those symbols will still be missed by both coverage discovery and trace validation.

tohboeh5:
Overview

Good Go support expansion and accompanying tests, but validation is still incomplete for unreadable files and idiomatic grouped `const` / `var` declarations. I left two blocking inline comments on those gaps.

PR #294, opened 2026-04-13
feat: add Java trace validation support

### LINKED ISSUE #282: [Feature]: Add Java language support to trace validation and symbol ownership
The current built-in adapter set does not give Java repositories a first-class path for symbol-aware trace validation. That undercuts PHIL-002's language-agnostic promise and narrows adoption to the languages already embedded in syu's own repository.

Java is a basic enterprise language that many specification-heavy teams still use. If syu wants to lower adoption barriers while keeping traceability real, Java should not be a second-class citizen.

### What change do you want?
Add a Java language adapter for trace validation and optional symbol ownership checks.

The first version should support at least:
- class, interface, enum, and record detection
- method detection
- field or constant detection where practical
- common JUnit test method discovery
- canonical aliases such as java

### Specification impact
REQ-CORE-002 should expand the supported-language contract, and the repository should gain a minimal Java trace fixture in tests.

### Alternatives considered
Rely on file-level wildcard ownership without symbol-aware checks. That is weaker than what syu already offers for current first-class languages.

### Before submitting

tohboeh5:
## Summary
- add a first-class Java language adapter for trace validation and extend strict ownership coverage to public Java types, members, and common JUnit test methods
- extend the passing multi-language fixture, CLI validation tests, and strict coverage fixtures to exercise Java traces end to end
- update the self-hosted validation requirement, config docs, generated site docs, and report to describe Java support

Closes #282


tohboeh5 (commented):
The Java support fits the existing coverage architecture well, and the fixture/test updates make the feature easy to trust. I found one reliability gap in the JUnit annotation matcher, though: it misses a common JUnit 5 annotation stack shape, which will turn valid traced tests into false coverage misses.

src/coverage.rs
```diff
@@ -547,6 +558,174 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+        if let Some(name) = captures.name("name") {
+            symbols.insert(name.as_str().to_string());
+        }
+    }
+    symbols.into_iter().collect()
+}
+
+fn collect_java_test_symbols(contents: &str) -> Vec<String> {
+    let annotation_regex = Regex::new(
+        r"(?ms)@Test(?:\s*\([^)]*\))?\s*(?:\r?\n\s*)*(?:public|protected|private)?\s*(?:static\s+)?(?:final\s+)?(?:<[^>{}]+>\s*)?(?:void|[\w\[\]<>?,]+)\s+(?P<name>[A-Za-z_][A-Za-z0-9_]*)\s*\(",
```
tohboeh5: This `@Test` matcher only allows whitespace/newlines between `@Test` and the method signature, so it misses the common JUnit 5 shape where extra annotations sit in between (for example `@Test` + `@DisplayName` + `@Tag`). In those cases `collect_java_test_symbols` returns nothing and `require_symbol_trace_coverage` will report a false negative for a perfectly valid traced test. Please allow intermediate annotations here before the visibility/signature portion.

src/coverage.rs
```diff
@@ -547,6 +558,174 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+                    CoverageTargetKind::TestSymbol => 1,
+                },
+            ))
+    });
+    targets.dedup();
+
+    Ok(targets)
+}
+
+fn collect_java_public_symbols(contents: &str) -> Vec<String> {
```
tohboeh5, line 849: collect_java_public_symbols still only inventories members that spell out public. In Java interfaces, methods and constants are public even without that modifier, so strict coverage quietly misses them. I reproduced this on the branch with public interface CoveredApi { void coveredMethod(); String EXPORTED_NAME = "x"; }: syu validate passed even though neither member was traced, so SYU-coverage-public-001 never fired.

src/coverage.rs
```diff
@@ -547,6 +558,174 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+    for captures in field_regex.captures_iter(contents) {
+        if let Some(name) = captures.name("name") {
+            symbols.insert(name.as_str().to_string());
+        }
+    }
+    symbols.into_iter().collect()
+}
+
+fn collect_java_test_symbols(contents: &str) -> Vec<String> {
+    let annotation_regex = Regex::new(
```
tohboeh5, line 894: This regex only matches @Test when the method signature comes immediately after the annotation block. Common JUnit layouts like @Test + @DisplayName + @Tag are skipped, so untracked tests can pass coverage silently. I reproduced it with an untraced untrackedStackedTest() using @Test, @DisplayName, and @Tag; syu validate still passed with require_symbol_trace_coverage: true.

tohboeh5 (commented):
Nice PR overall: the Java adapter wiring, fixtures, and guide updates move coverage support forward in a useful way. The remaining gaps are both in Java symbol discovery: interface members are still treated too narrowly, and common JUnit annotation stacks can let untracked tests slip past require_symbol_trace_coverage.

src/coverage.rs
tohboeh5, line 849: I reproduced this with a minimal Java workspace on the branch: `syu validate` passed even though `CoveredApi` contained uncovered interface members `coveredMethod` and `EXPORTED_NAME`. Because Java interface methods and constants are public even without an explicit `public` modifier, strict coverage is currently missing real public surface here and can report false passes.

src/coverage.rs
tohboeh5: The same minimal workspace also passed cleanly with an uncovered `untrackedStackedTest()` that used `@Test`, `@DisplayName`, and `@Tag` above the method. Because stacked JUnit 5 annotations are a common layout, this matcher is still letting real executable tests bypass `SYU-coverage-test-001` when there is any annotation between `@Test` and the method signature.

tohboeh5:
Overall impression: the Java adapter fits the existing validation model well, the fixture coverage is a good start, and the docs/spec updates make the new language support easy to discover. After reviewing the branch and exercising `syu validate` against a minimal Java workspace, I still see two blockers before merge: interface members are not inventoried unless they spell out `public`, and common JUnit 5 annotation stacks (`@Test` plus annotations like `@DisplayName` / `@Tag`) still let untracked tests pass coverage silently. Because both gaps produce false passes under `require_symbol_trace_coverage`, I do not think the Java support is ready to merge until symbol discovery matches real Java semantics more closely.

src/coverage.rs
```diff
@@ -547,6 +558,257 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+    }
+    for symbol in collect_java_public_interface_members(contents) {
+        symbols.insert(symbol);
+    }
+    symbols.into_iter().collect()
+}
+
+fn collect_java_test_symbols(contents: &str) -> Vec<String> {
+    let annotation_regex = Regex::new(
+        r"(?ms)@Test(?:\s*\([^)]*\))?\s*(?:(?:\r?\n\s*)*@[\w.]+(?:\s*\([^)]*\))?\s*)*(?:\r?\n\s*)*(?:public|protected|private)?\s*(?:static\s+)?(?:final\s+)?(?:<[^>{}]+>\s*)?(?:void|[\w\[\]<>?,]+)\s+(?P<name>[A-Za-z_][A-Za-z0-9_]*)\s*\(",
```
tohboeh5: This only matches bare `@Test`. Fully qualified JUnit annotations like `@org.junit.Test` and `@org.junit.jupiter.api.Test` will still be missed here, so valid Java tests can continue to show up as uncovered.

src/coverage.rs
```diff
@@ -547,6 +558,257 @@ fn discover_typescript_targets(root: &Path) -> Result<Vec<CoverageTarget>, Box<I
...
+            ))
+    });
+    targets.dedup();
+
+    Ok(targets)
+}
+
+fn collect_java_public_symbols(contents: &str) -> Vec<String> {
+    let type_regex = Regex::new(
+        r"(?m)^\s*public\s+(?:sealed\s+|non-sealed\s+|abstract\s+|final\s+)?(?:class|interface|enum|record)\s+(?P<name>[A-Z][A-Za-z0-9_]*)\b",
```
tohboeh5: Java public-symbol discovery is still too narrow here: constructors (which have no return type) and nested `public static` types are both common public APIs, but this pattern does not match them. Those symbols will still be absent from coverage.

tohboeh5:
Overview

Strong first step for Java support, but the discovery regexes are still too narrow for common Java and JUnit patterns. I left two blocking inline comments on missing public-symbol and test-annotation coverage.

PR #295, opened 2026-04-13
feat: add source purity trace policy

### LINKED ISSUE #279: [Feature]: Add a source-purity policy that stops requiring inline spec IDs inside implementation files
The current self-hosted model pushes traceability directly into source files by requiring implementation and test artifacts to mention the owning requirement or feature ID. That keeps ownership explicit, but it also makes everyday code noisier and risks turning implementation files into containers for repository bookkeeping rather than just code.

The repository itself demonstrates this tradeoff today: many Rust source files start with comment headers such as FEAT-CHECK-001 and REQ-CORE-002. The validation rules also explicitly encode this expectation through SYU-trace-id-001. If syu wants adoption to stay low-ceremony, teams should be able to keep source code clean while still preserving auditable traceability.

### What change do you want?
Introduce a first-class philosophy or policy for source purity that says traceability must stay repository-native and explainable without forcing inline spec IDs into implementation files by default. Then update validation and docs so inline IDs become one strategy rather than the required strategy.

The policy should make these constraints explicit:
- source code should stay readable without bookkeeping noise
- traceability still needs to remain reviewable and repository-local
- any alternative must still support audits, refactors, and explainability

### Specification impact
This likely touches philosophy and policy text first, then REQ-CORE-002 and the validation feature/rule set that currently define owning-ID mentions inside traced files.

### Alternatives considered
Keep the current inline-ID rule and only document it better. That would preserve the current implementation, but it does not address the underlying concern that syu may be teaching teams to pollute code in order to satisfy the spec.

### Before submitting

tohboeh5:
## Summary
- stop requiring traced implementation and test files to mention owning spec IDs, and keep `validate --fix` limited to optional documentation breadcrumbs
- add a first-class source-purity policy plus spec and guide updates that describe repository-local traceability without inline bookkeeping by default
- refresh passing fixtures and validation tests to prove clean source files still validate

Closes #279


tests/check_command.rs
```diff
@@ -513,6 +509,27 @@ fn check_command_reports_disabled_validation_toggles_from_config() {
     assert!(stdout.contains("validate.require_symbol_trace_coverage=false"));
 }
 
+#[test]
+// REQ-CORE-002
+fn check_command_accepts_trace_workspaces_without_inline_spec_ids() {
```
tohboeh5, line 514: This new regression test proves `validate` accepts trace maps without inline spec IDs, but the browser-app Playwright suite still hard-codes `SYU-trace-id-001` in `app/tests/browser-app.spec.ts`. Because that suite uses a hand-built payload, the stale issue code will keep passing even though this PR removes the rule from the real validation surface. Please add a browser-app fixture/assertion update here (or in a follow-up tied to this PR) so the UI tests stay aligned with the new rule catalog.

docs/guide/getting-started.md
```diff
@@ -249,34 +249,38 @@ codebase. Each trace entry has two key fields:
...
 fires with a suggestion to add the snippet (or run `syu validate --fix`).
+If checked-in YAML plus the symbol name already gives you enough traceability,
+you can omit `doc_contains` entirely and keep source files free of spec-ID
+bookkeeping.
 
 Requirements should declare tests:
 
 ```yaml
 status: implemented
 tests:
```
tohboeh5, line 277: This YAML snippet uses a much deeper indent for `rust:` than the surrounding examples. It is technically valid YAML, but it reads like a malformed block and makes the new first-run guidance harder to copy with confidence. Can we normalize this example to the same two-space nesting used elsewhere in the guide?

tohboeh5:
Overall impression: the direction is strong. Removing mandatory inline spec-ID bookkeeping fits the repository-local trace model better, the new policy text makes the philosophy shift explicit, and the passing-workspace fixtures/tests show the intended default clearly. The remaining gaps are mostly follow-through: the browser-app duplicate-issue fixture still points at `SYU-trace-id-001`, so the UI test surface is out of sync with the new rule catalog, and one of the getting-started YAML examples now reads more awkwardly than the surrounding docs because of its indentation. Once those are tightened, this looks like a good simplification of the default tracing model.

tohboeh5:
Overview

This source-purity shift is internally consistent. The validation behavior, autofix path, docs, and tests all move together, and I did not find a blocking correctness, security, performance, or UX issue in the current change set.

PR #301, opened 2026-04-13
feat: support sidecar ownership manifests

### LINKED ISSUE #281: [Feature]: Support sidecar ownership manifests so traceability can stay out of source code
Even if syu adopts a stronger source-purity stance, teams still need a concrete repository-native mechanism that tells the validator which functions, files, tests, and public APIs belong to which requirements and features. Right now the strongest built-in mechanism is to embed owning IDs inside the traced artifact itself.

That works, but it couples specification ownership to comment noise in production code. Teams that want clean source files need another way to keep ownership explicit without giving up the auditability that POL-003 is trying to protect.

### What change do you want?
Add support for checked-in sidecar ownership manifests or another repository-native trace index that can map requirement and feature IDs to files and symbols without requiring inline ownership comments in code.

A good first cut would let repositories choose one of these modes:
- inline owning IDs in source files
- sidecar ownership manifests stored near the code
- generated ownership indexes committed under the spec tree

The validator should understand the selected mode and still enforce symbol existence, canonical paths, and reciprocal consistency.

### Specification impact
REQ-CORE-002 and the validation feature/rule set will need updates. The docs should also explain how this mode preserves explainability without reintroducing hidden external metadata.

### Alternatives considered
Use external SaaS metadata or editor-only annotations. That would hide traceability outside the repository, which conflicts with syu's repository-native philosophy.

### Before submitting

tohboeh5:
## Summary
- add validate.trace_ownership_mode with inline and sidecar ownership support
- teach validate/autofix to read and update checked-in sidecar manifests
- refresh the spec, docs, generated site docs, and validation coverage

Closes #281


src/command/check.rs
```diff
@@ -1289,6 +1308,222 @@ fn describe_trace_reference(reference: &TraceReference) -> String {
...
+    manifest: &OwnershipManifest,
+    owner_id: &str,
+    reference: &TraceReference,
+) -> bool {
+    let required_symbols = required_ownership_symbols(reference);
+    manifest
+        .owners
+        .iter()
+        .filter(|entry| entry.id == owner_id)
+        .any(|entry| entry_covers_symbols(entry, &required_symbols))
```
tohboeh5, line 1414: Because this helper just scans matching entries with `.any(...)`, duplicate owner blocks in a sidecar manifest are silently accepted. I reproduced this with a manifest that repeated both `REQ-001` and `FEAT-001` twice, and `syu validate` still passed. That leaves the new format less strict than the main trace YAML and makes ownership ambiguous for readers. Could we reject duplicate `owners.id` entries (and probably duplicate symbols while we are here) when loading the manifest instead of treating them as valid?

src/command/check.rs
```diff
@@ -1289,6 +1308,222 @@ fn describe_trace_reference(reference: &TraceReference) -> String {
...
+    }
+
+    let wildcard_required = required_symbols.len() == 1 && required_symbols[0] == "*";
+    let mut changed = false;
+
+    if let Some(entry) = manifest
+        .owners
+        .iter_mut()
+        .find(|entry| entry.id == owner_id)
+    {
```
tohboeh5, line 1508: `--fix` only mutates the first matching owner entry here. If a manifest already contains duplicate blocks for the same ID, autofix merges missing symbols into the first one and leaves the stale duplicates behind. I verified that with a `REQ-001` manifest split across two entries: `syu validate --fix` passed, but the file still contained both `REQ-001` blocks afterward. Once duplicate owners slip through validation, autofix can no longer normalize the manifest back to one explicit source of truth.

tohboeh5:
Overall this is moving in a good direction. The config surface is small and explicit, the docs/spec updates make the new mode understandable, and the added tests cover the main success and failure paths for sidecar ownership well.

The main concern I ran into is manifest normalization/validation. Right now duplicate owner blocks are accepted, and `--fix` only updates the first matching block instead of restoring one canonical entry per owner. That makes the new sidecar format less strict than the existing trace YAML and leaves room for ambiguous manifests to accumulate over time. I left two inline comments on the specific code paths that seem worth tightening before this lands.

tohboeh5:
Overall impression: this is a thoughtful feature addition. Supporting sidecar ownership manifests is a meaningful step toward the source-purity direction, and the change is backed by solid command coverage plus configuration/docs updates. After reviewing the changed logic and tests, I did not find a substantive blocker in the implementation.

src/command/check.rs
```diff
@@ -1289,6 +1308,236 @@ fn describe_trace_reference(reference: &TraceReference) -> String {
...
+        }
+    };
+
+    if !merge_ownership_entry(&mut manifest, owner_id, reference) {
+        return Ok(());
+    }
+
+    let raw = serde_yaml::to_string(&manifest)?;
+    fs::write(&manifest_path, raw)?;
+    summary.updated_files.insert(
```
tohboeh5: This path is normalized differently from the inline ownership-update path earlier in the file: here we try to strip `root`, but the inline branch inserts `path.to_path_buf()` directly. `updated_files` is currently just a set/count, so it is not a functional bug today, but it does make the summary semantics depend on which ownership mode ran. I would consider routing both paths through the same helper so future reporting stays consistent.

tohboeh5:
Overview

PR #315, opened 2026-04-13
docs: add repository fit guide

### LINKED ISSUE #302: [Feature]: Add a repository-fit guide before the README install flow
Newcomers can reach the README and understand how to install syu, but they still have to infer whether syu is the right fit for their repository. README.md currently moves from a short introduction and the four-layer explanation into installation, and the website/getting-started flow also assumes the reader already decided to adopt syu.

That leaves a gap for teams who are still answering a more basic question: is syu the right tool for spec-as-code traceability in this repo, or should they stay with lighter documentation-only workflows?

### What change do you want?
Add an explicit "Is syu right for this repository?" decision guide before the install section in README.md and mirror that guidance in the website / getting-started entry points.

The guide should explain:
- when syu is a good fit
- when it is probably too heavy
- what trade-offs it makes compared with informal docs-only workflows
- what kind of repositories benefit most from repository-native traceability

### Specification impact
This likely updates README.md, docs/guide/getting-started.md, and website landing copy. It may also need a small documentation requirement or policy note clarifying the intended adoption envelope.

### Alternatives considered
Relying on the existing four-layer explanation is not enough because it explains the model after the reader is already inside the workflow, not how to decide whether they should adopt the tool in the first place.

### Before submitting

tohboeh5:
## Summary
- add a repository-fit guide ahead of the README install flow
- mirror the same adoption guidance in getting started
- surface the fit-check entry point on the docs-site landing page

Closes #302


tohboeh5 (commented):
Overall impression: this is a useful framing improvement for first-time adopters. The added repository-fit guidance makes the trade-offs explicit before people jump into installation, and surfacing the same question from the getting-started guide plus the docs landing page gives readers multiple chances to choose the right path. The copy stays concise while still being specific about when `syu` is and is not worth the extra structure.

After reviewing the documentation and site changes, I did not find a blocking issue in the current change set.

docs/guide/getting-started.md
```diff
@@ -20,10 +20,40 @@ Need a different level of guidance?
...
-it slows down at the first manual editing step so you can see where the
-scaffolded files live, how reciprocal links fit together, and what to fix
-before the first `syu validate .` run.
+If you are still deciding whether to adopt `syu`, use the fit check below before
+installing anything. The rest of this guide assumes `syu` is already installed.
+Unlike the README quick start, it slows down at the first manual editing step so
+you can see where the scaffolded files live, how reciprocal links fit together,
+and what to fix before the first `syu validate .` run.
+
+## Is syu right for this repository?
```
tohboeh5, line 28: Most of this fit-check block is now duplicated from the README almost verbatim. Because both pages target first-time users, the two decision matrices will drift quickly as soon as one list changes. I would rather keep one canonical version and link to it here than maintain the same newcomer guidance in two places.

tohboeh5:
This is a useful improvement to the onboarding path: putting the repository-fit question closer to the first-run guide matches how new adopters actually evaluate a tool like this. The main weakness is maintainability: the fit-check prose is now duplicated between the README and getting-started guide, which makes drift very likely unless one source becomes canonical.

docs/guide/getting-started.md
tohboeh5, line 28: Most of this fit-check block is now duplicated from the README almost verbatim. Because both pages target first-time users, the two decision matrices will drift quickly as soon as one list changes. I would rather keep one canonical version and link to it here than maintain the same newcomer guidance in two places.

tohboeh5:
This is a useful improvement to the onboarding path: putting the repository-fit question closer to the first-run guide matches how new adopters actually evaluate a tool like this. The main weakness is maintainability: the fit-check prose is now duplicated between the README and getting-started guide, which makes drift very likely unless one source becomes canonical.

website/src/pages/index.js
```diff
@@ -27,6 +27,12 @@ const layers = [
 ];
 
 const journeys = [
+  {
+    title: 'Check repository fit',
+    description:
+      'Read the adoption guide first when you are still deciding whether repository-native traceability is worth the extra structure.',
+    to: '/docs/guide/getting-started#is-syu-right-for-this-repository'
```
tohboeh5, line 33: /docs/guide/getting-started does not gain an `is-syu-right-for-this-repository` heading in this PR, so this new card points at a dead fragment. Please either add a matching section heading to `getting-started.md` or point this card at the new README section instead.

tohboeh5:
Overview

This is a useful onboarding improvement: moving the repository-fit question earlier matches how first-time readers evaluate a new tool, and linking it from the docs site is the right discovery surface. The remaining problem is correctness: both new entry points currently miss their target, so readers hit dead or wrong fragments instead of the fit-check section. Once those anchors point at a real destination, this will be a strong docs-path refinement.

PR #316, opened 2026-04-13
docs: add team-scale example workspace

### LINKED ISSUE #303: [Feature]: Add team-scale example workspaces that show phased adoption and multi-contributor growth
The repository ships only three example workspaces today: examples/rust-only, examples/python-only, and examples/polyglot. They are great smoke-test fixtures, but they all stay close to the same minimal starter shape.

That leaves a gap for teams trying to understand what syu looks like after months of use: larger requirement sets, multiple contributors, phased adoption, refactors, and recovery from traceability drift. Without that, newcomers can bootstrap syu but still struggle to picture how it scales in a real repository.

### What change do you want?
Add one or more richer example workspaces that show:
- team-scale growth beyond the single-feature starter shape
- phased adoption in an existing repository
- realistic file splitting / folder reorganization patterns
- recovery workflows for broken reciprocal links or orphaned items

The examples should stay runnable in CI and document why each pattern is useful.

### Specification impact
This likely adds new examples/ content, README/example references, and repository-quality tests that keep the richer examples healthy.

### Alternatives considered
Keeping the current language-focused examples is useful for smoke coverage, but they do not teach maintainers how to structure a larger spec or how to evolve one over time.

### Before submitting

tohboeh5:
## Summary
- add a richer team-scale example workspace with nested requirement and feature documents
- document phased adoption and recovery drills in the example README
- validate the new example in the automated example workspace suite

Closes #303


tohboeh5 (commented):
Overall impression: this is a strong addition to the examples set. The new team-scale workspace demonstrates phased adoption, nested document organization, and cross-area traceability in a way that is much closer to a real long-lived repository than the starter examples, and the extra validation test keeps the example from silently drifting. The README updates also make the example easier to discover for readers looking for a larger reference.

After reviewing the example workspace, validation coverage, and README updates, I did not find a blocking issue in the current change set.

examples/team-scale/syu.yaml
```diff
@@ -0,0 +1,11 @@
+version: 0.0.1-alpha.7
+spec:
+  root: docs/syu
+validate:
+  default_fix: false
+  allow_planned: true
```
tohboeh5, line 8: `allow_planned: true` feels risky in an example that otherwise presents itself as a healthy, validated end state. New adopters tend to copy example config verbatim, so this effectively teaches them to keep migration-mode validation turned on even after the workspace is stable. I would drop the flag here or call out explicitly why this example still needs it.

tohboeh5:
The new example does a good job of showing what a larger, split-out workspace can look like once a repository grows beyond the starter template. My main concern is the example config: shipping relaxed validation in an otherwise good example weakens the message about what a healthy steady-state workspace should look like.

examples/team-scale/syu.yaml
tohboeh5, line 8: `allow_planned: true` feels risky in an example that otherwise presents itself as a healthy, validated end state. New adopters tend to copy example config verbatim, so this effectively teaches them to keep migration-mode validation turned on even after the workspace is stable. I would drop the flag here or call out explicitly why this example still needs it.

tohboeh5:
The new example does a good job of showing what a larger, split-out workspace can look like once a repository grows beyond the starter template. My main concern is the example config: shipping relaxed validation in an otherwise good example weakens the message about what a healthy steady-state workspace should look like.

examples/team-scale/syu.yaml
```diff
@@ -0,0 +1,13 @@
+version: 0.0.1-alpha.7
+spec:
+  root: docs/syu
+validate:
+  default_fix: false
+  # This example models phased adoption, so planned items are intentionally
+  # allowed while the repository grows into a fully strict graph.
+  allow_planned: true
```
tohboeh5, line 8: `allow_planned: true` relaxes delivery validation, but every item in this example is already `implemented` and the test only covers the green path. That means the example teaches a migration-only setting without showing why it is needed. Either remove the flag so this models steady-state defaults, or add an actual planned-item scenario that justifies it.

tohboeh5:
Overview

The new team-scale workspace is valuable and the example does a good job of showing how a larger repository can organize spec material once it grows past the starter layouts. My main concern is that the shipped config normalizes a relaxed validation mode that this example does not actually need, which makes the steady-state recommendation less clear than it should be. Tightening that one piece would make the example much easier for future adopters to trust and copy.

PR #317, opened 2026-04-13
docs: add spec anti-patterns guide

### LINKED ISSUE #304: [Feature]: Add a spec anti-patterns guide for common four-layer design mistakes
The current guides explain the four-layer model and the validation rules, and docs/guide/troubleshooting.md explains specific validation failures. What is missing is a guide for specs that technically validate but are still headed toward a messy structure.

New teams need concrete warnings about anti-patterns such as unstable philosophies, policy layers that just repeat requirements, or requirements that are really implementation notes. Without that guidance, users can stay "green" while still building a spec they will want to rewrite later.

### What change do you want?
Add a dedicated guide for spec anti-patterns and refactoring triggers. It should explain mistakes such as:
- philosophy entries that change every sprint
- policies that duplicate lower layers
- requirements that should be split or promoted
- signals that a document or layer has become too broad
- when to merge, split, or rename spec items

### Specification impact
This is primarily documentation work in docs/guide/, but it may also justify new examples or cross-links from the concepts and troubleshooting guides.

### Alternatives considered
Leaving the advice implicit in the concepts guide and the rule catalog does not help enough, because users need concrete examples of what a bad-but-valid spec looks like and when to refactor it.

### Before submitting

tohboeh5:
## Summary
- add a dedicated guide for bad-but-valid four-layer design mistakes
- cross-link the new guide from concepts, troubleshooting, and the README
- surface the guide from the docs-site landing page

Closes #304


tohboeh5 (commented):
Overall impression: this fills a real documentation gap well. The new anti-patterns guide is practical instead of abstract, the smell/why/refactor structure makes it easy to act on, and the cross-linking from the README, concepts, troubleshooting, and site entry points gives the guide good discoverability without making any single page feel overloaded.

After reviewing the added guide and the surrounding documentation wiring, I did not find a blocking issue in the current change set.

docs/guide/anti-patterns.md
```diff
@@ -0,0 +1,175 @@
...
+contributors must follow, promote it to policy. If it only exists to support
+one behavior, move it down to a requirement.
+
+## Policy anti-patterns
+
+### Policy that just repeats one requirement
+
+**Smell:** the policy and requirement say almost the same thing.
+
+```yaml
```
tohboeh5: The surrounding guide uses real `syu`-shaped YAML everywhere else, so this snippet reads like copy-pasteable config even though `policy:` / `requirement:` are not valid document shapes here. I would either mark this as pseudocode or rewrite it into the actual list-based schema so newcomers do not learn the wrong document form from an anti-pattern example.

tohboeh5:
The guide is directionally strong: calling out bad-but-valid spec shapes is exactly the kind of higher-order help a self-hosted spec system needs. The weak spot is the examples: a few of them look like literal YAML snippets even though they are only conceptual, which can easily teach the wrong shape to readers who skim and copy.

docs/guide/anti-patterns.md
tohboeh5: The surrounding guide uses real `syu`-shaped YAML everywhere else, so this snippet reads like copy-pasteable config even though `policy:` and `requirement:` are not valid document shapes here. I would either mark this as pseudocode or rewrite it into the actual list-based schema so newcomers do not learn the wrong document form from an anti-pattern example.

tohboeh5:
The guide is directionally strong: calling out bad-but-valid spec shapes is exactly the kind of higher-order help a self-hosted spec system needs. The weak spot is the examples: a few of them look like literal YAML snippets even though they are only conceptual, which can easily teach the wrong shape to readers who skim and copy.

docs/guide/anti-patterns.md
```diff
@@ -0,0 +1,177 @@
...
```
tohboeh5: This guide is framed as “bad but valid” spec shapes, but this snippet is not a valid `syu` document shape: it mixes policy and requirement entries in one list and omits the normal top-level keys. Because the surrounding docs use copy-pasteable YAML, please mark these snippets as pseudocode or rewrite them as real schema-shaped examples.

tohboeh5:
Overview

The topic is a strong addition for the repository: calling out spec anti-patterns gives reviewers vocabulary for problems that validation alone cannot catch. The main weakness is that some example snippets currently read like copy-pasteable YAML even though they are not valid `syu` document shapes, which undermines the guides core promise of explaining bad-but-valid patterns. Recasting those examples as explicit pseudocode or valid-schema counterexamples would make the guide much safer and clearer.

PR #319, opened 2026-04-13
ci: name browser app artifacts by run

### LINKED ISSUE #306: [Bug]: Replace the hardcoded pr226 browser-app artifact name in CI
### What area is affected?
Repository automation or CI

### What happened?
The browser-app job uploads app/dist with a hardcoded artifact name, pr226-browser-app-dist, in .github/workflows/ci.yml. When that job fails on any other pull request, the uploaded artifact name does not identify the real PR or branch that produced it.

That makes CI troubleshooting harder because contributors have to guess which artifact belongs to their run instead of seeing an artifact name tied to the actual PR or ref.

The browser-app artifact name should include a dynamic PR number, run identifier, or branch-safe ref so the uploaded bundle is clearly tied to the failing workflow run.

### syu version
main

### Environment
GitHub Actions browser-app job on ubuntu-latest.

### Steps to reproduce
1. Open .github/workflows/ci.yml.
2. Inspect the "Upload browser app dist" step in the browser-app job.
3. Notice that the artifact name is fixed to pr226-browser-app-dist regardless of which PR triggered the run.

The current workflow step is:
- uses: actions/upload-artifact@v4
- with.name: pr226-browser-app-dist

### Before submitting

tohboeh5:
## Summary
- replace the stale hardcoded browser-app artifact name with a run-scoped name
- keep the uploaded bundle easy to tie back to the failing workflow run
- add repository-quality coverage for the artifact naming convention

Closes #306

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,7 +252,8 @@ jobs:
         if: success()
         uses: actions/upload-artifact@v7
         with:
-          name: browser-app-dist
+          # Keep a stable prefix for discovery while tying each artifact to one run.
+          name: browser-app-dist-run-${{ github.run_id }}-attempt-${{ github.run_attempt }}
           path: app/dist
           if-no-files-found: error
 
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -715,7 +715,11 @@ fn repository_ships_browser_app() {
     assert!(ci_workflow.contains("Build browser app bundle"));
     assert!(ci_workflow.contains("scripts/ci/check-browser-app-freshness.sh"));
     assert!(ci_workflow.contains("if: success()"));
-    assert!(ci_workflow.contains("name: browser-app-dist"));
+    assert!(
+        ci_workflow.contains(
+            "browser-app-dist-run-${{ github.run_id }}-attempt-${{ github.run_attempt }}"
+        )
+    );
     assert!(build_script.contains("syu-app-dist"));
     assert!(build_script.contains("npm ci"));
     assert!(build_script.contains("build:wasm"));


tohboeh5 (commented):
Overall impression: this is a clean CI fix. Replacing the hardcoded browser-app artifact name with a run-scoped identifier solves the debugging problem without adding workflow complexity, and the repository-quality assertion gives the change a good regression guard. Using `run_id` plus `run_attempt` is a solid choice because it works across trigger types and reruns.

After reviewing the workflow/test updates, I did not find a blocking issue in the current change set.

.github/workflows/ci.yml
```diff
@@ -268,7 +268,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: pr226-browser-app-dist
+          name: browser-app-dist-run-${{ github.run_id }}-attempt-${{ github.run_attempt }}
```
tohboeh5, line 256: Switching away from the hardcoded `pr226` name fixes the obvious bug, but tying the artifact name to `run_id` and `run_attempt` also removes the only stable identifier downstream tooling can target. Any later job or maintainer script now has to enumerate artifacts and pattern-match the prefix before it can fetch the browser bundle. I would consider keeping a stable prefix contract here, or at least documenting that consumers now need prefix-based lookup.

tohboeh5:
Fixing the stale artifact name is clearly the right direction, and the new name is much more honest than the old hardcoded value. My reservation is about downstream ergonomics: the new contract is better for uniqueness, but worse for discoverability unless other automation is ready to search by prefix instead of a stable artifact name.

tohboeh5:
Overview

No substantive issues found. This change fixes the misleading hardcoded browser-app artifact name cleanly, keeps a stable prefix for discovery, and adds a targeted repository-quality check so the naming contract does not drift again.

PR #320, opened 2026-04-13
fix: improve pre-commit bootstrap diagnostics

### LINKED ISSUE #307: [Feature]: Improve scripts/install-precommit.sh diagnostics for failed hook bootstrap
The contributor bootstrap path depends on scripts/install-precommit.sh, including the devcontainer post-create flow. When that script cannot discover the final pre-commit binary path, it currently falls back through several branches and can fail without much context about which lookup failed.

That makes setup failures harder to debug for contributors who are already in the highest-friction part of the workflow: getting their local environment ready.

### What change do you want?
Improve scripts/install-precommit.sh so it explains what it is trying, what failed, and what the user should do next when pre-commit installation or lookup does not succeed.

A good first cut would add clear error messages for the user-base and pipx lookup branches, plus a troubleshooting hint that matches CONTRIBUTING.md.

### Specification impact
This likely touches scripts/install-precommit.sh, CONTRIBUTING.md, and possibly the contributor workflow docs/spec that describe the local bootstrap flow.

### Alternatives considered
Relying on contributors to rerun the script manually with shell tracing is workable, but it is much rougher than the repo's otherwise guided contributor setup story.

### Before submitting

tohboeh5:
## Summary
- add explicit install and lookup diagnostics to scripts/install-precommit.sh
- document the same PATH, user-base, and pipx checks in CONTRIBUTING
- cover the failure path with an integration test and refresh generated docs

Closes #307


tohboeh5 (commented):
Overall impression: this is a useful contributor-experience improvement. The script now surfaces the exact PATH locations it checked, the failure guidance is actionable instead of generic, and the matching documentation/test updates keep the troubleshooting story coherent. The added bash logic is still straightforward to follow despite the extra diagnostics.

After reviewing the script, tests, and documentation updates, I did not find a blocking issue in the current change set.

CONTRIBUTING.md
```diff
@@ -135,6 +135,12 @@ If you use the hooks, install them once:
 scripts/install-precommit.sh
 ```
 
+If bootstrap fails or the script cannot find the final `pre-commit` binary, run
+`python -m site --user-base` and `pipx environment --value PIPX_BIN_DIR` to see
```
tohboeh5: The troubleshooting advice switches to `python -m site --user-base`, but the installer itself prefers `python3` when that is the only interpreter on PATH. On systems where `python` is absent, this guidance fails at exactly the moment the user is already stuck. I would mirror the script wording here (`python3` or `python`) or tell readers to reuse the interpreter the script detected.

tohboeh5:
The extra diagnostics are genuinely helpful and they move the bootstrap failure mode from mysterious to actionable. The one thing I would tighten is the docs wording: the new troubleshooting command should stay aligned with the script own interpreter detection so contributors do not hit a second failure while following the recovery advice.

scripts/install-precommit.sh
```diff
@@ -22,40 +32,83 @@ install_precommit() {
...
+  local user_base_output
 
   if command -v pre-commit >/dev/null 2>&1; then
     command -v pre-commit
     return 0
   fi
 
-  local user_base
-  user_base="$("$python_bin" -m site --user-base)"
+  if ! user_base_output="$("$python_bin" -m site --user-base 2>&1)"; then
```
tohboeh5: Capturing `2>&1` into `user_base_output` here (and similarly for the `pipx environment` lookup below) can turn benign stderr into part of the path on a successful command. A warning from `sitecustomize`, Python, or pipx would make the computed path invalid and the script would wrongly report that `pre-commit` is missing. Keep stderr out of the success value and only print captured diagnostics on non-zero exit.

CONTRIBUTING.md
```diff
@@ -135,6 +135,14 @@ If you use the hooks, install them once:
 scripts/install-precommit.sh
 ```
 
+If bootstrap fails or the script cannot find the final `pre-commit` binary, run
+the same interpreter check that the script selected:
+`python3 -m site --user-base` (or `python -m site --user-base` when the script
+detected `python` instead of `python3`). Then run
+`pipx environment --value PIPX_BIN_DIR` to see where the binary should live,
```
tohboeh5: This troubleshooting step assumes `pipx` exists, but the script explicitly supports the no-`pipx` path via `python -m pip --user`. On those machines this advice sends contributors into a second failure. Please make the `pipx environment --value PIPX_BIN_DIR` step conditional or explicitly label it as `pipx`-only guidance.

tohboeh5:
Overview

The diagnostics and test coverage are moving in the right direction: this patch makes failed hook bootstrap much more actionable for contributors, and I like that the troubleshooting surface is being documented instead of left as tribal knowledge. The remaining issues are both in edge-case robustness: successful commands should not absorb stderr into path values, and the new CONTRIBUTING advice should not assume `pipx` on the fallback install path the script already supports.

PR #321, opened 2026-04-13
fix: explain app port binding recovery

### LINKED ISSUE #308: [Bug]: Make syu app port binding failures explain how to recover
### What area is affected?
CLI or runtime behavior

### What happened?
When syu app cannot bind its port, the error only reports that binding failed for the selected address and port. It does not tell the user that the port may already be in use, that they can retry with --port, or that they can configure a different default in syu.yaml.

That leaves a first-time user with an error but very little guidance about the next step.

Port binding failures should explain the likely cause and suggest recovery steps such as trying a different --port value or setting app.port in syu.yaml.

### syu version
main

### Environment
Local CLI usage of syu app on Linux/macOS/Windows when the default port is already occupied.

### Steps to reproduce
1. Start another process on port 3000.
2. Run syu app .
3. Observe that the bind failure names the address but does not explain how to recover.

src/command/app.rs currently wraps the bind error with a generic context: failed to bind <bind>:<port>.

### Before submitting

tohboeh5:
## Summary
- expand syu app bind failures with likely causes and recovery steps
- point users to --port for retries and app.port for a different default
- cover the occupied-port case with CLI and unit tests

Closes #308


tohboeh5 (commented):
Overall impression: this is a solid UX improvement for `syu app` startup failures. The new guidance is actionable, the split between `AddrInUse` and other bind failures keeps the message specific, and the added integration coverage exercises the real occupied-port path instead of only unit-level formatting. I also like that the helper stays small and keeps the CLI-facing wording easy to audit.

After reviewing the branch and the surrounding command/test code, I did not find a blocking issue in the current change set.

src/command/app.rs
```diff
@@ -163,6 +163,18 @@ enum StartupLine {
...
 
+fn bind_failure_message(bind: IpAddr, port: u16, error: &std::io::Error) -> String {
+    let likely_cause = if error.kind() == std::io::ErrorKind::AddrInUse {
+        "The selected port is likely already in use."
+    } else {
+        "The selected address or port may be unavailable on this machine."
+    };
+
+    format!(
+        "failed to bind `{bind}:{port}`. {likely_cause} Try `syu app . --port <free-port>` to retry with a different port, or set `app.port` in syu.yaml to change the default. {error}"
```
tohboeh5: The recovery hint hard-codes `syu app . --port <free-port>`. If the user launched `syu app path/to/workspace` from outside the workspace root, this retry command points at the wrong repository and can fail again for a completely different reason. I would preserve the current workspace path in the suggested command instead of forcing `.`.

tohboeh5:
Improving bind-failure diagnostics is absolutely the right UX move, and the new message is much more actionable than the bare socket error. The main weakness is that the suggested retry command loses the caller original workspace path, which can send someone debugging a port issue into an unrelated workspace-resolution problem.

tohboeh5:
Overview

No substantive issues found. This is a good diagnostics improvement with appropriate unit and integration coverage, and the current version appears to preserve the recovery hint context cleanly.

PR #322, opened 2026-04-13
fix: warn before advertising public app binds

### LINKED ISSUE #309: [Bug]: Print the non-loopback security warning before syu app advertises the listening URL
### What area is affected?
CLI or runtime behavior

### What happened?
When syu app binds to a non-loopback address, it prints the listening URL before it prints the warning that workspace data may be reachable from other machines on the network.

That ordering is risky for first-time users because the command advertises the URL before it explains the security implication of exposing the local workspace over the network.

The non-loopback warning should be shown before syu app prints the listening URL or browser-open guidance so users see the exposure warning first.

### syu version
main

### Environment
Local CLI usage of syu app with a non-loopback bind such as --bind 0.0.0.0.

### Steps to reproduce
1. Run syu app --bind 0.0.0.0 .
2. Observe the startup output order.
3. Notice that the listening URL is printed before the warning about network exposure.

src/command/app.rs prints "syu app listening on ..." before it later emits the non_loopback_warning_lines startup messages.

### Before submitting

tohboeh5:
## Summary
- print non-loopback security warnings before advertising the listening URL
- keep ready/open guidance separate from exposure warnings in startup lines
- verify the combined startup log shows the warning before the URL

Closes #309


tohboeh5 (commented):
Overall impression: this is a focused and well-executed security/UX fix. Printing the non-loopback warning before the listening URL matches the intent of the warning much better, the stdout/stderr flushing makes the ordering explicit, and the regression test checks the actual combined output ordering instead of only asserting message presence. The refactor also reads cleanly.

After reviewing the branch and the surrounding command/test code, I did not find a blocking issue in the current change set.

src/command/app.rs
tohboeh5, line 1: This still prints the warning on stderr and the advertised URL on stdout. Flushing helps, but tooling that merges the streams later (redirected logs, CI output, some terminal wrappers) still does not guarantee relative ordering across file descriptors. Since the whole point of the change is to warn before advertising, I would keep both messages on the same stream or assert on a combined log in the tests.

tohboeh5:
The intent here is good and user-facing: move the security warning earlier so the public bind is not advertised before the risk is explained. The remaining weakness is the split streams: because the warning and URL still go to different file descriptors, the ordering guarantee is weaker than the patch and tests imply.

src/command/app.rs
```diff
@@ -194,16 +203,18 @@ pub fn run_app_command(args: &AppArgs) -> Result<i32> {
                 .await
                 .context("local app server exited unexpectedly")
         });
+        print_startup_lines(non_loopback_warning_lines(local_addr.ip()));
```
tohboeh5: Printing the warning on stderr before the URL does not actually guarantee users will see it first. stdout and stderr are separate streams, so terminals and log collectors can still reorder them. If the ordering matters, emit both startup lines through one stream (or one serialized writer) instead of relying on flushes.

tohboeh5:
Overview

Good intent, but the implementation still does not fully guarantee the user-visible ordering across separate stdout/stderr streams. I left one blocking inline note on that ordering assumption; the rest of the change looks reasonable.

PR #323, opened 2026-04-13
fix: repeat parse-error file paths in the app

### LINKED ISSUE #310: [Bug]: Repeat the YAML file path in browser-app parse error banners
### What area is affected?
CLI or runtime behavior

### What happened?
The browser app shows parse errors inside the item header panel, but the parse-error banner itself does not repeat the failing YAML file path. Users have to look elsewhere in the view to reconnect the parse error with the file they should edit.

That makes debugging harder than the CLI flows, which keep the failing file path closer to the error itself.

When currentDocument.parse_error is present, the browser app should show the document path directly inside the parse-error banner or in an equally prominent place next to the error.

### syu version
main

### Environment
Browser app UI served by syu app against a workspace with malformed YAML.

### Steps to reproduce
1. Start syu app . with a workspace that contains an invalid YAML document.
2. Open the failing document in the browser UI.
3. Observe that the parse-error banner shows the parser error text but does not repeat the file path inside the banner itself.

app/src/App.tsx renders currentDocument.parse_error in the warning banner, while the file path is only shown separately in the source panel summary.

### Before submitting

tohboeh5:
## Summary
- repeat the failing YAML path inside browser-app parse-error banners
- add a Playwright regression that injects malformed YAML through the app payload
- rebase the branch onto the latest main

Closes #310

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -905,6 +905,9 @@ function App() {
                 <p className="font-medium">
                   This document could not be parsed into the expected layer model.
                 </p>
+                <p className="mt-2 break-all font-mono text-xs text-amber-50/90">
+                  File: {currentDocument.path}
+                </p>
                 <p className="mt-2 text-xs leading-6 text-amber-50/80">
                   {currentDocument.parse_error}
                 </p>
--- a/app/tests/browser-app.spec.ts
+++ b/app/tests/browser-app.spec.ts
@@ -7,6 +7,11 @@ const usesFailingWorkspace =
   process.env.SYU_APP_E2E_WORKSPACE?.includes("tests/fixtures/workspaces/failing") ?? false;
 
 type AppDataPayload = {
+  source_documents: Array<{
+    section: "philosophy" | "policies" | "features" | "requirements";
+    path: string;
+    content: string;
+  }>;
   validation: {
     issues: Array<{
       code: string;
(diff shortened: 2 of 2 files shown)


tohboeh5 (commented):
Overall impression: this is a clean browser-app UX fix. Repeating the YAML path directly inside the parse-error banner makes the failure much easier to act on, and the Playwright regression covers the exact user-visible problem instead of only the API payload shape. The UI copy and styling fit the existing error treatment well.

After reviewing the branch and the surrounding app/test changes, I did not find a blocking issue in the current change set.

app/src/App.tsx
```diff
@@ -905,6 +905,9 @@ function App() {
                 <p className="font-medium">
                   This document could not be parsed into the expected layer model.
                 </p>
+                <p className="mt-2 break-all font-mono text-xs text-amber-50/90">
+                  File: {currentDocument.path}
```
tohboeh5, line 932: This helps once the user has already opened the broken document, but the navigation list still gives no clue which entry is the one with the parse error. In a multi-file workspace the first-time workflow is still to guess documents until the banner appears. I would consider surfacing the same path or error state in the sidebar, or auto-selecting the first parse-error document, so the file hint becomes discoverable earlier.

tohboeh5:
Repeating the file path inside the parse-error banner is a real UX improvement: it removes one frustrating guess once the user is on the broken document. The remaining issue is discoverability before that point — the patch still relies on the user opening the right document first, so there is another chance to make parse failures more obvious from the navigation surface itself.

tohboeh5:
Overview

No substantive issues found. Repeating the file path in the parse-error banner is a focused UX improvement, it matches the scope of the bug, and the added browser regression coverage looks proportionate to the change.

PR #324, opened 2026-04-13
feat: add source-first trace lookup command

### LINKED ISSUE #311: [Feature]: Add a source-first trace lookup command that starts from files and symbols
The current CLI is strong when I already know a spec ID: I can browse, list, show, and search the spec. What is still missing is the reverse workflow that starts from code or a traced symbol and asks which requirements, features, policies, and philosophies it belongs to.

That gap is especially painful during review and refactoring, where the starting point is often a file or function name instead of a requirement ID.

### What change do you want?
Add a source-first trace lookup command that accepts a repository path and symbol name (or lets the user choose one) and then resolves the linked requirements, features, policies, and philosophies for that source artifact.

A good first cut could support:
- file-only lookup
- file + symbol lookup
- output in both text and JSON
- guidance when the source is unowned or only partially traced

### Specification impact
This likely needs a new CLI feature, docs/help updates, and spec coverage for source-first navigation behavior.

### Alternatives considered
Today users can search by spec text or inspect IDs they already know, but that is still an ID-first workflow and does not solve the common "start from code" review path.

### Before submitting

tohboeh5:
## Summary
- add a new syu trace command for file-first and symbol-first ownership lookups
- surface owned, partial, and unowned guidance in text and JSON output
- cover the command in CLI docs, self-hosted spec, generated docs, and regression tests

Closes #311


tohboeh5 (commented):
Overall impression: this is a strong addition to the CLI. Starting from code and walking back to the traced spec surface is a useful workflow, the owned/partial/unowned status model is easy to understand, and the command/help/README/test coverage makes the feature discoverable from multiple entry points. The validation and path-handling logic also reuse existing patterns instead of inventing a separate model.

After reviewing the branch and the surrounding command/test/docs changes, I did not find a blocking issue in the current change set.

src/command/trace.rs
```diff
@@ -0,0 +1,907 @@
...
+                    "trace file `{}` must stay under workspace `{}`",
+                    file.display(),
+                    workspace.root.display()
+                )
+            })?
+            .to_path_buf()
+    } else {
+        file.to_path_buf()
+    };
+    let normalized = normalize_relative_path(&relative);
```
tohboeh5, line 148: `normalize_relative_path` preserves leading `..` segments, so a call like `syu trace ../other/file.rs` will normalize and keep going instead of failing fast as a non-workspace path. That breaks the new “repository-relative source file” contract and turns an invalid selector into a misleading `unowned` result. Please reject relative paths that escape the workspace after normalization, the same way other commands guard repository-relative inputs.

tohboeh5:
Overall impression: the command is a strong addition in spirit. Starting from code instead of a spec ID fits the project well, the text/JSON surface is easy to understand, and the spec/docs/test updates make the feature discoverable. The blocker I found is path validation: the new lookup still accepts `../...` selectors that escape the workspace, so an invalid source path can quietly report `unowned` instead of failing as out-of-bounds input. Tightening the repository-relative path check would make the command much safer and more trustworthy.

src/command/trace.rs
tohboeh5, line 148: `normalize_relative_path()` preserves leading `..`, so `syu trace ../other/file.rs` is accepted and reported as `unowned` instead of being rejected as outside the workspace. That breaks the new repository-relative contract and can mislead users into thinking the lookup succeeded. Reject normalized paths that still escape the workspace and add an integration test for `../...`.

tohboeh5:
Overview

The source-first lookup command is a valuable addition and the output/doc surface is already in good shape. The remaining defect is at the workspace boundary: paths that still escape the repo after normalization are accepted too far and then misreported as ordinary misses, which weakens the contract the command is trying to establish. Tightening that guard would make the UX much more trustworthy.

PR #325, opened 2026-04-13
feat: add VS Code extension for syu diagnostics

### LINKED ISSUE #312: [Feature]: Build a VS Code extension for syu navigation and validation diagnostics
The repository already acknowledges VS Code and Codespaces as important contributor environments, but syu still requires users to leave the editor and switch to terminal/browser flows for most spec navigation and validation feedback.

That means a common workflow is missing: seeing traceability and spec diagnostics directly where people edit code and YAML.

### What change do you want?
Add a VS Code extension that can:
- surface validation diagnostics in the editor
- link source files to requirements/features
- jump from spec IDs to documents and traced files
- open syu-specific views without requiring the user to manually run terminal commands for every lookup

### Specification impact
This likely needs a new extension package, documentation, and a stable protocol between the extension and syu's lookup/validation engine.

### Alternatives considered
The current terminal and browser UI flows are helpful, but they still force editor users to context-switch away from the place where they are making changes.

### Before submitting

tohboeh5:
## Summary
- add a checked-in VS Code extension under `editors/vscode/` for diagnostics, trace lookup, and spec navigation
- document the extension in the README, guide, homepage, and self-hosted spec docs
- cover the new editor surface with Node tests and repository quality assertions

Closes #312


editors/vscode/src/model.js
```diff
@@ -0,0 +1,616 @@
...
+  const traceLocation = parseTraceLocation(issue.location)
+  const subjectItem = itemFromIssueSubject(issue, model)
+  const relativePath =
+    traceLocation ||
+    (looksLikeWorkspaceRelativeFile(issue.location)
+      ? normalizeRelativePath(issue.location)
+      : null) ||
+    subjectItem?.documentPath ||
+    'syu.yaml'
+  const targetPath = path.join(workspaceRoot, toSystemPath(relativePath))
```
tohboeh5: Absolute coverage/read/walk locations from the current CLI contract can already be real absolute paths. Joining every path-like `issue.location` onto `workspaceRoot` breaks Windows in particular (`C:\repo\C:\tmp\file.rs`). Please detect absolute paths first and pass them through unchanged before applying workspace-relative normalization.

editors/vscode/src/model.js
```diff
@@ -0,0 +1,616 @@
...
+    return { line: 0, startCharacter: 0, endCharacter: 0 }
+  }
+}
+
+function findIssueRange(contents, issue, subjectItem) {
+  const lines = contents.split(/\r?\n/u)
+  const searchTerms = []
+
+  if (subjectItem && isFieldName(issue.location)) {
+    searchTerms.push(`${issue.location}:`)
```
tohboeh5: This search order can highlight the wrong item in multi-entry YAML files. When `issue.location` is a generic field like `status`, the code looks for `status:` before it anchors on `id: <subject>`, so the first matching field in the document wins even if it belongs to another requirement/feature. Please scope the field lookup under the matched item block, or anchor on the item first and then search within that range.

tohboeh5:
Overall impression: the extension is a strong and welcome addition. The command wiring, docs surface, and repository/spec updates make the feature easy to discover, and shipping model-level tests alongside the new editor integration is the right shape for maintainability.

The remaining issues are both in diagnostics targeting, which is one of the core UX promises of the extension. Right now multi-item YAML files can highlight the wrong field, and absolute file locations are mis-resolved on Windows because they are treated as workspace-relative. Once those targeting bugs are fixed, this will be a much more trustworthy first version of the VS Code integration.

editors/vscode/src/model.js
tohboeh5: `issue.location` can already be an absolute path from CLI diagnostics. Joining it unconditionally onto `workspaceRoot` mangles absolute paths (`C:\repo\C:\tmp\file.rs` on Windows, `/repo/tmp/file.rs` on POSIX), so the extension can open or highlight the wrong file. Detect absolute paths first and pass them through unchanged, then add a regression test.

editors/vscode/src/model.js
```diff
@@ -0,0 +1,616 @@
...
+  } catch {
+    return { line: 0, startCharacter: 0, endCharacter: 0 }
+  }
+}
+
+function findIssueRange(contents, issue, subjectItem) {
+  const lines = contents.split(/\r?\n/u)
+  const searchTerms = []
+
+  if (subjectItem && isFieldName(issue.location)) {
```
tohboeh5: This searches for the field name before anchoring on the target item ID, so in multi-entry YAML files it can highlight the same field on the wrong item. Please anchor on `id: ${subjectItem.id}` first and then search within that item block, with a regression test covering duplicate field names.

tohboeh5:
Overview

This is a strong first extension: the command surface, repository wiring, and diagnostics story all fit the project well, and the docs/spec/test coverage is thoughtful. The main issue is reliability in diagnostics-driven navigation: two path/field-resolution edge cases can send users to the wrong file or wrong YAML entry, which would undermine trust in the extension quickly. Once those cases are anchored correctly, this looks like a compelling editor integration.

PR #327, opened 2026-04-15
fix: warn before advertising non-loopback app url

### LINKED ISSUE #309: [Bug]: Print the non-loopback security warning before syu app advertises the listening URL

tohboeh5:
## Summary
- print the non-loopback exposure warning before the app advertises its listening URL
- flush startup lines as they are emitted so the warning is visible before later guidance
- cover the new startup ordering in app command tests

Closes #309


src/command/app.rs
```diff
@@ -222,6 +210,9 @@ pub fn run_app_command(args: &AppArgs) -> Result<i32> {
 fn startup_lines(local_addr: SocketAddr) -> Vec<StartupLine> {
```
tohboeh5: This is still relying on stderr for the warning and stdout for the URL, so the relative order is only best-effort. A human running the command interactively will usually see the warning first, but merged logs and redirected output can still show the listening URL before the warning. If the guarantee matters, emit the whole startup sequence on one stream.

tohboeh5:
The change moves in the right direction by trying to keep the warning ahead of the public bind announcement. My main concern is the same ordering contract as in the earlier app-startup work: splitting the startup sequence across stdout and stderr makes the guarantee weaker than the user-facing wording suggests.

src/command/app.rs
```diff
@@ -194,22 +194,10 @@ pub fn run_app_command(args: &AppArgs) -> Result<i32> {
                 .await
                 .context("local app server exited unexpectedly")
         });
-        println!("syu app listening on http://{local_addr}");
         tokio::task::spawn_blocking(move || wait_for_ready(local_addr))
```
tohboeh5, line 192: This still waits for `wait_for_ready(local_addr)` to finish before printing the non-loopback warning. That means the security warning is silent during startup delays or readiness hangs, even though the process has already bound to a public interface. If the intent is to warn users as soon as we know the bind is public, can we emit the warning lines before the readiness probe and only keep the `listening`/`ready` messages behind the probe?

tohboeh5:
Overall impression: this heads in the right direction because it keeps the warning ahead of the advertised URL output and adds coverage for the stdout/stderr split. The remaining concern is that the warning is still delayed until after the readiness probe completes, so a slow or stuck startup can leave users with no early signal that they chose a public bind. I left one inline comment on that startup ordering edge case.

tohboeh5:
Overview

This is a sensible security/UX fix. Moving the non-loopback warning ahead of the advertised URL makes the risk harder to miss, and the extracted startup-line helper keeps the behavior easy to test. I did not find a substantive diff-local issue in the current patch.

Issue #336, opened 2026-04-15
[Bug]: Make browser search keyboard shortcuts easier to discover


tohboeh5:
### What area is affected?
CLI or runtime behavior

### What happened?
The browser app supports useful keyboard shortcuts for search navigation, but the shortcut help is easy to miss for first-time users.

The current hint sits in small, low-contrast helper text below the search input. Users who immediately start typing can miss that the app supports ArrowUp, ArrowDown, Enter, and Escape navigation.

### What did you expect to happen?
The keyboard shortcut affordance should be more discoverable, for example with:
- a more prominent hint treatment
- a hover/focus tooltip
- or a small info affordance near the search field that explains the shortcuts

### syu version
main

### Environment
Browser app UI

### Steps to reproduce
1. Run `syu app .`
2. Open the browser app and focus on the search input
3. Notice that the keyboard shortcut help sits below the input in low-emphasis styling
4. A first-time user can easily miss the available keyboard navigation

### Logs, screenshots, or failing output
- `app/src/App.tsx:700-713` renders the shortcut hint text below the search field
- the help text uses `text-xs text-slate-400`, which gives it low visual emphasis compared to the rest of the search UI

### Before submitting
- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
This has already landed on main via #339, so I am closing the issue to match the shipped implementation.

PR #339, opened 2026-04-15
fix: surface search keyboard shortcuts

tohboeh5:
## Summary

Make the browser-app search shortcuts easier to notice by turning the helper text into a more prominent keyboard-shortcuts callout and covering the new copy in Playwright.

- Issue: #336

## Validation

- [x] `scripts/ci/check-app-dist-freshness.sh` (stopped with the expected regenerated `app/dist` diff before commit)
- [x] `npm --prefix app run test:e2e`

## Release notes

- [x] No release note needed


app/src/App.tsx
```diff
@@ -697,20 +697,28 @@ function App() {
                 className="w-full rounded-2xl border border-white/10 bg-slate-900/60 py-2 pl-9 pr-4 text-sm text-slate-100 placeholder-slate-500 focus:border-sky-400/60 focus:outline-none focus:ring-1 focus:ring-sky-400/40"
               />
             </div>
-            <p
+            <div
```
tohboeh5, line 722: Because the search input still points `aria-describedby` at this container, moving from a short inline hint to a full multi-sentence callout makes screen readers announce a lot more text every time the field receives focus. That is a UX regression for keyboard / assistive-tech users even though the visual panel is nicer. Consider keeping a short dedicated description for the input and making the richer panel purely visual.

tohboeh5:
Overall impression: the new shortcut callout is visually clearer and easier to notice than the old one-line hint, and the updated browser test does a good job of locking the intended copy in place. The downside is accessibility verbosity: because the searchbox still uses `aria-describedby`, the richer panel now becomes the full spoken description on every focus, which makes the field noticeably noisier for screen-reader users. Separating the accessible description from the visual callout would keep the UX improvement without regressing assistive-tech flow.

app/src/App.tsx
tohboeh5, line 1: Because this panel is still referenced from `aria-describedby` on the search box, expanding it into a multi-paragraph callout makes the field's accessible description much longer every time focus lands on the input. I'd keep a short assistive-tech description separate from the richer visual panel so the shortcuts stay discoverable without making the search field verbose for screen-reader users.

tohboeh5:
Overview

This direction improves shortcut discoverability and the added test coverage makes the UI intent clear. The main issue I found is accessibility-related: the richer shortcut panel is still wired into the search box via `aria-describedby`, so assistive-tech users now get a much longer description every time the field receives focus. I would keep the prominent visual treatment, but split out a shorter AT-specific description before merging.

Issue #334, opened 2026-04-15
[Bug]: Make the browser app search result limit explicit before truncation


tohboeh5:
### What area is affected?
CLI or runtime behavior

### What happened?
The browser app search silently truncates results to the first 20 items, and the UI only explains that limit when the rendered result list length is exactly 20.

That makes common searches ambiguous: when a user sees 20 results, they have to infer whether that is the complete set or just the visible slice. When they see fewer than 20, they get no hint that the UI has a hard cap at all.

### What did you expect to happen?
The search UI should make the result limit explicit before truncation becomes confusing. For example:
- always show a short note about the current limit
- expose a “show more” / pagination path
- or move the limit into the input help text so users understand the boundary up front

### syu version
main

### Environment
Browser app UI

### Steps to reproduce
1. Run `syu app .`
2. Search for a common term such as `core` or `check`
3. Observe that the result list is sliced to 20 items
4. Notice that the UI only mentions truncation when `searchResults.length === 20`

### Logs, screenshots, or failing output
- `app/src/App.tsx:744-747` shows the truncation note only when the visible list length is exactly 20
- the search placeholder at `app/src/App.tsx:665` does not mention the result limit

### Before submitting
- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
This has already landed on main via #340, so I am closing the issue to match the shipped implementation.

PR #340, opened 2026-04-15
fix: clarify browser search result limit

tohboeh5:
## Summary

Make the browser-app search result limit explicit before truncation by surfacing the 20-match cap in the search placeholder and helper copy, then tightening the truncation message itself.

- Issue: #334

## Validation

- [x] `scripts/ci/check-app-dist-freshness.sh` (stopped with the expected regenerated `app/dist` diff before commit)
- [x] `npm --prefix app run test:e2e`

## Release notes

- [x] No release note needed


app/src/App.tsx
```diff
@@ -743,7 +747,7 @@ function App() {
                 )}
                 {searchResults.length === 20 && (
                   <p className="px-2 py-1 text-[11px] text-slate-500">
-                    Showing first 20 results — refine your query for fewer matches.
+                    Showing the first 20 matches — refine your query for fewer results.
```
tohboeh5: This copy still keys off `searchResults.length === 20`, which cannot tell the difference between “there are more than 20 matches” and “there are exactly 20 matches.” In the exact-20 case the UI now claims truncation even though nothing was dropped. Please derive this message from the untrimmed match count so the new limit guidance stays truthful.

tohboeh5:
Overall impression: this is a useful UX clarification, and I like that the placeholder, visible helper copy, and Playwright coverage all move together. The remaining issue is correctness: the new truncation message still infers “showing the first 20” from the trimmed result length, so users can be told results were cut off even when the query actually has exactly 20 matches. Once the message is driven by the full match count instead of the displayed slice length, this should be a clean improvement.

app/src/App.tsx
tohboeh5, line 1: This copy says results were truncated whenever the rendered list reaches 20, but the list is already sliced to 20 matches earlier. Queries with exactly 20 total matches will now show the warning even though nothing was hidden. Can we gate this on the unsliced match count (or a `hasMoreResults` boolean) instead of the displayed length?

tohboeh5:
Overview

This is a solid UX pass on the search limit messaging: the placeholder, helper copy, and truncation notice all push in the same direction and the Playwright updates cover the new strings well. The blocking issue is correctness rather than copy: the warning is still keyed off the displayed array length, so a query with exactly 20 matches will be reported as truncated. Once the UI can distinguish "exactly 20" from "20 out of more", this looks ready.

Issue #333, opened 2026-04-15
[Bug]: Explain browser trace metrics for first-time users


tohboeh5:
### What area is affected?
CLI or runtime behavior

### What happened?
The browser app shows “requirement traces” and “feature traces” as `validated/declared`, but it does not explain what those terms mean to a first-time user.

That creates friction right in the first screen of the app: a newcomer can see the numbers, but not why they matter or how to interpret them.

### What did you expect to happen?
The trace metrics should include lightweight context, such as a tooltip, inline hint, or jump-to-validation affordance that explains:
- what “declared” means
- what “validated” means
- why a gap between them matters

### syu version
main

### Environment
Browser app UI

### Steps to reproduce
1. Run `syu app .`
2. Open the browser app
3. Look at the workspace metrics cards for requirement traces and feature traces
4. Notice that the UI shows ratios but does not explain the trace terms themselves

### Logs, screenshots, or failing output
- `app/src/App.tsx:625-637` renders the trace ratio cards
- the cards only show `validated / declared` with no explanatory hint or affordance

### Before submitting
- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
This has already landed on main via #341, so I am closing the issue to match the shipped implementation.

PR #341, opened 2026-04-16
fix: explain browser trace metrics for first-time users

tohboeh5:
## Summary\n- explain the requirement/feature trace metrics inline in the browser workspace sidebar\n- reuse the existing info-hint UI so the metrics stay lightweight but discoverable\n- cover the new hints in Playwright while keeping the browser build artifacts in sync\n\n## Testing\n- npm --prefix app run check\n- npm --prefix app run build\n- npm --prefix app run test:e2e\n- cargo test\n\nCloses #333

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -628,6 +628,11 @@ function App() {
                 note="validated / declared"
                 tone="sky"
                 ratio={requirementTraceRatio}
+                hint={{
+                  label: "Requirement traces",
+                  description:
+                    "Declared traces are the requirement test references written in the spec. Validated traces are the declared references that syu could confirm in the current workspace. A gap means some declared requirement traces are stale or unresolved.",
+                }}
               />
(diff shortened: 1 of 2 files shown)


app/dist/index.html
tohboeh5, line 1: The source change in `app/src/App.tsx` looks fine, but the checked-in browser bundle is still stale in CI. Please regenerate `app/dist` (for example with `npm --prefix app run build`) and commit the updated artifacts so the shipped app actually includes the new metric hints.

tohboeh5:
Overview

The new metric hints are a good newcomer-facing improvement, and reusing the existing InfoHint pattern keeps the UI consistent. The blocker here is operational: CI is still reporting the checked-in browser bundle as stale, so the deployed app would not actually contain the source change yet. After regenerating and committing `app/dist`, the feature direction looks good.

PR #346, opened 2026-04-16
feat: add a go-only example workspace [REQ-CORE-002]

### LINKED ISSUE #330: [Feature]: Add a go-only example workspace for unsupported-language adoption
The repository tells newcomers to study the shipped examples, but every checked-in example uses a language set with full first-class support.

That leaves teams evaluating `syu` for Go-style repositories without a minimal, runnable example that shows how unsupported-language adoption works today. They can read about the concept, but they cannot inspect a working repository that demonstrates the trace patterns and limitations in practice.

### What change do you want?
Add a minimal `examples/go-only/` workspace that demonstrates how to use `syu` in a repository whose implementation language does not rely on `doc_contains`.

- one small Go source file and one `_test.go` file
- a matching four-layer spec under `docs/syu/`
- passing `syu validate .` output
- a short README that explains which traceability capabilities work today and which are still future enhancements

### Specification impact
This extends the examples/documentation surface and should update the places that point newcomers toward starter examples.

### Alternatives considered
Relying on language-support implementation issues alone still leaves current-day adopters without a concrete example repository they can copy from.

### Before submitting

### Evidence
- `docs/guide/getting-started.md:396-405` points newcomers to the checked-in examples
- `examples/` currently ships `rust-only`, `python-only`, and `polyglot` examples only
- `README.md:273-280` only advertises Rust/Python/Polyglot starter templates

tohboeh5:
## Summary
- add a minimal Go language adapter so syu can validate explicit Go symbol traces in .go files
- ship a checked-in examples/go-only workspace and cover it in the example workspace test suite
- update README, guides, self-spec, and generated reference docs to reflect Go trace support and the new example

## Testing
- scripts/ci/check-generated-docs-freshness.sh
- cargo test

Closes #330


docs/syu/requirements/core/validation.yaml
tohboeh5, line 1: This PR updates the spec/docs to say strict symbol-trace coverage now supports Go, but `src/coverage.rs` still only discovers Rust, Python, and TypeScript targets. Without a Go discovery path there, enabling `require_symbol_trace_coverage` will not actually enforce ownership for exported Go symbols or `Test*` functions yet.

tohboeh5:
Overview

The Go example workspace and language-adapter work are both heading in a useful direction, and I like that the docs try to make the unsupported-language adoption path more concrete. The blocking gap is that the PR now documents Go as covered by strict symbol-trace coverage, but `src/coverage.rs` still only discovers Rust/Python/TypeScript targets. Until Go coverage discovery is implemented, the docs/spec claim is ahead of the actual validator behavior.

PR #357, opened 2026-04-16

tohboeh5:
## Summary
- make the README quick start promise match its real prerequisites
- state that the README-only path only depends on the local four-layer refresher
- keep the concepts guide as optional deeper reading instead of a hidden prerequisite

- Issue: #255

## Validation


README.md
```diff
@@ -32,8 +32,8 @@ verified.
 Pick the newcomer path that matches what you need next:
 
 - **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .` and only need a short layer refresher before the first
-  commands.
+  to `syu validate .`. It only assumes the [Why four layers?](#why-four-layers)
+  refresher below, and the deeper concepts guide stays optional.
 - **Tutorial**: follow [`docs/guide/tutorial.md`](docs/guide/tutorial.md) when you
```
tohboeh5, line 45: Routing `new to syu` readers to the full tutorial still skips the existing `getting-started.md` guide, which is the shorter narrated onboarding path the repo already maintains. If this chooser is meant to reduce confusion, I would surface Getting started here and reserve Tutorial for readers who specifically want the longer repository story.

tohboeh5:
Overview

The chooser copy is moving in a useful direction: it tries to explain who each path is for instead of assuming readers already know the docs set. The main thing still missing is the actual narrated first-run lane. Right now the chooser points brand-new users at the full tutorial while the dedicated Getting started guide remains hidden in the secondary link list, which weakens the simplification this PR is trying to achieve. I left that inline.

Issue #264, opened 2026-04-12
[Feature]: Make the README path chooser explain which guide fits which reader


tohboeh5:
## What problem are you trying to solve?

The README Choose your path section lists Quick start, Tutorial, and Troubleshooting, but it does not tell a newcomer how to choose between them based on prior knowledge or learning style.

Today the section explains what each path contains, but not who it is for. A first-time reader still has to guess whether they should stay in the README, jump to the tutorial, or read concepts first.

Evidence:
- README.md lines 30-40 list the three entry points
- the chooser does not mention prior familiarity with the four-layer model, time cost, or whether the reader wants a narrative walkthrough versus a reference flow

## What change do you want?

Make the chooser more explicit about which path fits which reader.

For example:
1. call out whether the path assumes the reader already understands philosophy, policy, requirements, and features
2. mention approximate time or depth for each path
3. distinguish reference-style quick setup from narrative tutorial learning more clearly

## Specification impact

This touches PHIL-003, POL-004, POL-005, and FEAT-DOCS-001 because it changes how the repository lowers adoption friction for new users.

## Alternatives considered

- Leaving the chooser as-is: shortest edit, but it keeps the decision criteria implicit.
- Replacing the chooser with a single canonical path: simpler, but it gives up the benefit of offering multiple entry points for different readers.

## Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

tohboeh5:
Closing as already fixed on `main`: the README path chooser now explains which guide fits which reader, including prior familiarity and time/depth cues for Getting started, Quick start, Tutorial, and the other entry paths.

PR #358, opened 2026-04-16
docs: clarify README path chooser audiences

tohboeh5:
## Summary
- make the README path chooser explain who each onboarding path is for
- add depth and time-shape cues so readers can pick a path with less guesswork
- keep the existing README, tutorial, and troubleshooting entry points intact

- Issue: #264

## Validation


README.md
```diff
@@ -31,14 +31,16 @@ verified.
 
 Pick the newcomer path that matches what you need next:
 
-- **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .` and only need a short layer refresher before the first
-  commands.
+- **Quick start**: stay in this README when you already grasp the four-layer
```
tohboeh5: This promise feels a little too strong for the flow below. The quick start still asks the reader to edit reciprocal links manually and reason about adjacent layers, so saying it only assumes the `Why four layers?` refresher risks under-preparing first-time users. I would either soften this wording or add a direct pointer to `getting-started.md` right here so the escape hatch appears before the manual YAML step.

tohboeh5:
Overview

I like the intent to make the README chooser more explicit about audience and depth. That is the right problem to solve. The remaining issue is expectation-setting: the new Quick start wording now sounds more self-contained than the actual flow is, even though the reader still hits manual reciprocal-link editing a few steps later. I left one inline note on tightening that promise so the README does not oversell how much context the quick path requires.

PR #359, opened 2026-04-16
docs: split README and getting-started roles

tohboeh5:
## Summary
- position the README quick start as the compact command card
- point readers to getting-started when they want the narrated first-run path
- describe getting-started as the canonical walkthrough for first-time setup

- Issue: #256

## Validation


README.md
```diff
@@ -163,8 +163,13 @@ below. The rest of this section assumes you installed the published CLI.
 
 ## Quick start
 
-Stay in this README for the shortest install-to-validate path. If you skipped
-[`docs/guide/concepts.md`](docs/guide/concepts.md), use the
+Treat this section as the compact command card. Stay in this README when you
```
tohboeh5, line 215: I like the README / guide role split, but this fork still happens after the reader has already chosen the README path. If the goal is to help newcomers pick the right depth up front, I would move the `README quick card vs narrated getting-started walkthrough` distinction into the top-level `Choose your path` section instead of relying on readers to discover it only once they are already inside Quick start.

tohboeh5:
Overview

The direction here is good: separating the compact README flow from the slower narrated guide is exactly the kind of onboarding refinement this repository needs. The wording inside Quick start is also clear. The remaining UX gap is placement. The split only becomes visible after someone has already landed in the README quick path, so first-time readers can still make the wrong choice at the top-level chooser. Moving that distinction earlier would make the improvement much more effective.

PR #360, opened 2026-04-16
docs: add repository-fit guide

### LINKED ISSUE #302: [Feature]: Add a repository-fit guide before the README install flow

tohboeh5:
## Summary
- add a repository-fit decision guide before the README install flow
- mirror the adoption guidance in the getting-started guide
- surface the same decision point on the website landing page

## Validation

Closes #302
Spec impact: documentation-only update for README, getting-started, and website landing copy.


docs/guide/getting-started.md
```diff
@@ -25,6 +25,40 @@ it slows down at the first manual editing step so you can see where the
 scaffolded files live, how reciprocal links fit together, and what to fix
 before the first `syu validate .` run.
 
+## Is syu right for this repository?
```
tohboeh5, line 28: This section now duplicates the same repository-fit decision tree that the README also adds. Because both copies are user-facing entry points, they are very likely to drift the next time the project tightens its onboarding language. I would keep one canonical fit-check section (probably in the README) and link to it here instead of maintaining two near-identical versions.

tohboeh5:
Overview

This is a helpful onboarding improvement in spirit. Asking readers whether `syu` is a good fit before dropping them into commands matches how people evaluate a new tool, and the docs-site homepage update makes the idea easier to discover. The remaining issue is maintainability: the repository-fit rubric is now duplicated across multiple entry points, so the wording will drift unless one source becomes canonical. I left one inline comment on consolidating that.

PR #362, opened 2026-04-16
docs: add spec anti-patterns guide

### LINKED ISSUE #304: [Feature]: Add a spec anti-patterns guide for common four-layer design mistakes

tohboeh5:
## Summary
- add a dedicated spec anti-patterns guide for common four-layer design mistakes
- cross-link it from the README, concepts guide, and troubleshooting guide
- extend repository quality checks to keep the new guide and links present

## Validation

Spec: FEAT-DOCS-001
Closes #304


docs/guide/spec-antipatterns.md
```diff
@@ -0,0 +1,210 @@
...
+
+## 1. Philosophy that changes every sprint
+
+Philosophy should protect values that survive multiple releases. If a
+philosophy entry changes whenever the current delivery plan changes, it is
+usually living in the wrong layer.
+
+**Bad smell:**
+
+```yaml
```
tohboeh5: These YAML blocks still look close enough to real document snippets that readers can copy them as-is, even though they omit the surrounding fields that the rest of the docs teach as normal document shape. I would mark them explicitly as conceptual fragments or switch to pseudocode-style bullets so the guide teaches the anti-pattern without accidentally implying this is valid starter YAML.

tohboeh5:
Overview

The guide is aiming at a real gap: `syu validate .` can stay green while the layer model still drifts, and giving contributors review vocabulary for those cases is valuable. The examples and sectioning also make the page easy to skim. My main concern is presentation: several of the anti-pattern examples still read like copy-pasteable YAML rather than clearly illustrative fragments, so the page risks teaching malformed shapes while trying to explain good structure. With that clarified, this would be a useful addition.

PR #364, opened 2026-04-16
test: derive duplicate browser-app issue code from payload

### LINKED ISSUE #300: [Bug]: Browser-app duplicate-issue Playwright fixture still targets removed SYU-trace-id-001 rule
### What area is affected?
Repository automation or CI

### What happened?
PR #295 removes `SYU-trace-id-001` from the validation surface, but `app/tests/browser-app.spec.ts` still hard-codes that issue code in the duplicate-issue selection tests. Because those Playwright tests drive a hand-built payload instead of the live validation output, CI can stay green even though the browser-app fixture is now out of sync with the actual rule catalog.

That leaves the UI regression suite less trustworthy right after a specification change that alters the issue list shown in the app.

When a validation rule is removed or renamed, the browser-app Playwright suite should either derive its duplicate-issue fixture from the current app payload or update its hard-coded issue code in the same change. The UI regression tests should fail when the real validation surface and the browser-app fixture drift apart.

### syu version
main

### Environment
Linux, Rust 1.94.1, Node 24.14.0 locally; GitHub Actions browser-app job on ubuntu-latest with Node 25.

### Steps to reproduce
1. Check out PR #295 (`feat/issue-279-source-purity`) or apply the same removal of `SYU-trace-id-001` from `src/command/check.rs` and the validation spec docs.
2. Inspect `app/tests/browser-app.spec.ts`.
3. Notice that the duplicate-issue tests still look for `SYU-trace-id-001` at lines 135/138/170/172/176.
4. Run the browser-app test suite; it can still pass because the tests use a hand-built payload rather than the live validation output from the changed rule catalog.

```text
app/tests/browser-app.spec.ts:135:  const duplicateIssueRows = page.getByRole("button", { name: /SYU-trace-id-001/i });
app/tests/browser-app.spec.ts:138:    .getByRole("heading", { level: 3, name: "SYU-trace-id-001" })
app/tests/browser-app.spec.ts:170:  const reorderedPayload = swapDuplicateIssues(payload, "SYU-trace-id-001");
app/tests/browser-app.spec.ts:172:  const duplicateIssueRows = page.getByRole("button", { name: /SYU-trace-id-001/i });
app/tests/browser-app.spec.ts:176:    .getByRole("heading", { level: 3, name: "SYU-trace-id-001" })
```

### Before submitting

tohboeh5:
## Summary\n- derive the duplicate issue code from the live app payload instead of hard-coding a rule ID\n- keep the existing duplicate-selection assertions focused on the two relevant validation messages\n- preserve refresh-reordering coverage while making the fixture resilient to rule renames\n\n## Linked issue\n- Closes #300\n\n## Validation\n- [x] bash scripts/ci/quality-gates.sh\n- [x] npm --prefix app run test:e2e\n


app/tests/browser-app.spec.ts
```diff
@@ -19,6 +19,23 @@ type AppDataPayload = {
   };
 };
 
+function duplicateIssueCodeForMessages(payload: AppDataPayload, messages: string[]): string {
+  const matchingCodes = payload.validation.issues
+    .filter((issue) => messages.includes(issue.message))
+    .map((issue) => issue.code);
+
+  if (matchingCodes.length !== messages.length) {
```
tohboeh5: This helper removes the hard-coded rule ID nicely, but the same two fixture message strings are still duplicated in both tests below. Extracting them to a shared constant would keep the new lookup logic just as readable while making future fixture-message churn cheaper to update.

tohboeh5:
Overview

This is a good resilience improvement. Deriving the duplicate issue code from the live payload is the right fix for the removed `SYU-trace-id-001` dependency, and the helper keeps the intent of the tests easy to follow. The only thing I would still tighten is maintainability inside the test file: the duplicate message strings are repeated in both call sites, so future fixture-message churn will still need synchronized edits in multiple places. Pulling those strings into one shared constant would make the test less brittle overall.

PR #365, opened 2026-04-16
docs: add example versus template onboarding matrix

### LINKED ISSUE #355: [Feature]: Add an example-versus-template matrix for newcomer path selection
The repository now has both starter templates and checked-in example workspaces, but that newcomer surface is spread across README, getting-started, and CLI help.

Users can see that examples exist, yet still have to infer which ones are scaffoldable with `syu init` and which ones are reference-only repositories to copy from.

### What change do you want?
Add a single example/template comparison page that makes the newcomer path explicit.

- language/example/template matrix in docs
- links from README and getting-started
- clear wording about when to prefer `syu init` versus when to study or copy a checked-in example

### Specification impact
This is a docs/UX improvement and should stay aligned with example and init-template features.

### Alternatives considered
The information already exists in fragments, but the current split makes the first decision point harder than it needs to be.

### Before submitting

tohboeh5:
## Summary\n- add a newcomer-facing guide that compares starter templates with checked-in examples\n- link that guide from README and getting-started at the points where people choose a path\n- extend repository docs coverage tests so the new guide stays wired into onboarding\n\n## Linked issue\n- Closes #355\n\n## Validation\n- [x] bash scripts/ci/quality-gates.sh\n- [x] npm --prefix website run build\n


README.md
```diff
@@ -257,6 +257,10 @@ starter layout:
 syu init . --spec-root spec/contracts --template rust-only
 ```
 
+Not sure whether you should start from a template or inspect a checked-in
```
tohboeh5, line 301: The new page title and filename use "examples and templates", but the links added in this PR call it the "examples-versus-templates guide". That mismatch will make the doc harder to search and refer to consistently. Please pick one name and use it everywhere.

tohboeh5:
Overview

This is a useful onboarding addition: the decision table is concrete, the new guide is well scoped, and the docs wiring keeps the choice visible from the main newcomer paths. The main issue I found is naming consistency. The new document is titled and filed as "examples and templates", but the link text added around the docs set still calls it the "examples-versus-templates guide", which makes the page harder to search and reference. Once that wording is made consistent, this looks like a strong docs improvement.

PR #367, opened 2026-04-16
ci: add merge queue checks manifest

tohboeh5:
## Summary
- add a checked-in merge queue required-checks manifest
- document the maintainer update path for merge queue check renames
- assert the manifest stays aligned with workflow-level merge queue jobs

## Testing
- cargo fmt --all


.github/merge-queue-checks.json
```diff
@@ -0,0 +1,18 @@
+{
+  "version": 1,
+  "required_checks": [
+    { "context": "precommit", "workflow": "ci", "job_id": "precommit" },
+    { "context": "quality", "workflow": "ci", "job_id": "quality" },
+    { "context": "coverage", "workflow": "ci", "job_id": "coverage" },
+    { "context": "actionlint", "workflow": "ci", "job_id": "actionlint" },
+    { "context": "dependency-audit", "workflow": "ci", "job_id": "dependency-audit" },
+    { "context": "npm-audit", "workflow": "ci", "job_id": "npm-audit" },
+    { "context": "dependency-review", "workflow": "ci", "job_id": "dependency-review" },
```
tohboeh5: `dependency-review` is listed as a required merge-queue check here, but the actual workflow job only runs on `pull_request` events. That means a merge-group run will wait forever for this check even though it can never execute there. Please remove it from the manifest or make the workflow runnable on `merge_group` before relying on this file.

tests/repository_quality.rs
```diff
@@ -599,6 +601,14 @@ fn repository_declares_dependency_hygiene_and_ci_caching() {
     assert!(codeql_workflow.contains("github/codeql-action/init@v4"));
     assert!(codeql_workflow.contains("github/codeql-action/autobuild@v4"));
     assert!(codeql_workflow.contains("github/codeql-action/analyze@v4"));
+    assert!(merge_queue_checks.contains("\"version\": 1"));
```
tohboeh5: These assertions only prove that certain strings exist in the source files; they do not verify that every manifest entry corresponds to a job that can actually run on `merge_group`. That gap is why the `dependency-review` mismatch slips through. I would strengthen this test so it validates functional alignment rather than substring presence alone.

tohboeh5:
Overview

The checked-in merge-queue manifest is a good idea in principle, and I like the intent to make queue expectations reviewable instead of tribal. The current patch still has a critical correctness gap, though: it lists `dependency-review` as a required merge-queue check even though that workflow job is gated to `pull_request` events, so merge-group runs can never satisfy it. The accompanying test coverage also checks only for string presence, which does not protect against this kind of drift. I would fix the manifest/workflow mismatch and harden the validation before merging.

PR #368, opened 2026-04-16
feat: add starter template discovery command

tohboeh5:
## Summary
- add `syu templates` with text and JSON output for starter discovery
- document the new discovery flow in the spec, README, and getting-started guide
- trace the new command through tests and generated docs

## Testing
- cargo fmt --all
- cargo run -- templates
- cargo run -- templates --format json
- python3 scripts/generate-site-docs.py
- cargo run -- report . --output docs/generated/syu-report.md


src/command/templates.rs
```diff
@@ -0,0 +1,139 @@
...
+            name: "polyglot",
+            description: "Mixed-language starter that keeps the same four layers while naming the first spec around a polyglot repository.",
+            relationship: TemplateRelationship::TemplateAndExample,
+            related_example: Some("examples/polyglot"),
+        },
+    }
+}
+
+fn print_text_catalog(templates: &[TemplateCatalogEntry]) {
+    for template in templates {
```
tohboeh5, line 88: The text formatter prints four tab-separated fields but no header row, so first-time users have to guess what each column means (`starter-only`, `template-and-example`, the optional path, etc.). For a discovery command, that makes the human-readable output harder to use than it needs to be. Please add column headers or switch to a more self-describing text layout.

src/command/templates.rs
```diff
@@ -0,0 +1,139 @@
...
+        StarterTemplate::Polyglot,
+    ]
+    .into_iter()
+    .map(template_catalog_entry)
+    .collect()
+}
+
+fn template_catalog_entry(template: StarterTemplate) -> TemplateCatalogEntry {
+    match template {
+        StarterTemplate::Generic => TemplateCatalogEntry {
```
tohboeh5, line 59: These descriptions are doing double duty as CLI output and docs copy, but they do not describe the template set in a consistent way. `Generic` foregrounds the four-layer shape, while the language-specific entries switch to different phrasing and imply different file coverage. Tightening them into one shared structure would make template comparison much easier for users.

tohboeh5:
Overview

This is a solid feature direction. The new `syu templates` command fits the project well, the JSON shape is clean, and the tests/docs updates make the command easy to discover. The main issue I found is UX in the text output: the command currently prints four unlabeled tab-separated columns, which forces first-time users to infer what each field means. I also think the template descriptions should use a more consistent structure so the available starters are easier to compare at a glance. With those usability refinements, this would be in good shape.

PR #388, opened 2026-04-18
feat: clarify which README path fits which reader

tohboeh5:
## Summary\n- clarify who each README entry path is for, including prior knowledge and time/depth expectations\n- make Quick start vs Getting started vs Tutorial easier to choose without guessing\n- keep the repository quality assertions aligned with the updated README wording\n\nCloses #264


tohboeh5:
## Overview
This makes the README chooser much more concrete for newcomers: the revised copy now tells readers how much guidance to expect and roughly how long each path takes. I did not find a blocking issue in the diff, and the accompanying repository-quality assertions keep the wording from drifting again.

Issue #374, opened 2026-04-17
[Feature]: Link the migration guide in the README path chooser and guide list


tohboeh5:
### What problem are you trying to solve?

`docs/guide/migration.md` exists, but the main README pathways do not point to it. Users upgrading between alpha releases or migrating an existing workspace have to discover the migration guide indirectly instead of seeing it as a documented entry path alongside Quick start, Tutorial, and Troubleshooting.

Evidence:
- The README path chooser does not mention upgrading or migration scenarios.
- The README "Keep the detailed guides close" list omits `docs/guide/migration.md`.
- The guide itself exists, so this is a discoverability gap rather than a missing-document problem.

### What change do you want?

Add an upgrade/migration option to the README path chooser and include `docs/guide/migration.md` in the main guide list. Update the migration guide header so it explicitly says this is the place to start when upgrading an existing `syu` workspace between alpha releases.

### Specification impact

This is mainly a documentation-pathway change, but it likely touches the docs feature/policy language about entry-point clarity and maintainability.

### Alternatives considered

Relying on the file existing under `docs/guide/` is technically sufficient, but it under-serves existing users compared with the care already given to first-run newcomer paths.

### Before submitting

- [x] I searched for existing issues and discussions first.
- [ ] I can help validate the change against a real project story.

tohboeh5:
Closing as completed on main. README.md now links docs/guide/migration.md both in the path chooser and in the main guide list, which covers the discoverability gap described here.

PR #389, opened 2026-04-18
feat: link the migration guide from the README chooser

tohboeh5:
## Summary\n- add a migration/upgrade entry to the README path chooser\n- include the migration guide in the main README guide list\n- keep the repository quality assertions aligned with the updated guide list\n\nCloses #374


tohboeh5:
## Overview
Adding the migration guide to the newcomer chooser closes a real onboarding gap for existing `syu` users who are upgrading between alpha releases. The change is small, stays consistent with the README structure, and I did not find a blocking issue in the diff.

PR #391, opened 2026-04-18
fix: make validate quiet mode suppress success summaries

tohboeh5:
## Summary
- make successful text runs with --quiet emit only the validation status line
- keep the full summary for non-quiet output and for failure cases
- update the CLI help text and regression coverage for the quiet contract

## Testing
- cargo test --test check_command check_command_quiet_suppresses_success_summary_and_next_step_guidance
- cargo run -- validate .
- cargo run -- validate . --quiet

### DIFF
--- a/src/command/check.rs
+++ b/src/command/check.rs
@@ -771,6 +771,7 @@ fn render_text_report(
 ) -> String {
     let mut output = String::new();
     let status = if overall_success { "passed" } else { "failed" };
+    let quiet_success = quiet && overall_success && result.issues.is_empty();
     let filtered_suffix = if filtered_view.is_some() {
         " (filtered view)"
     } else {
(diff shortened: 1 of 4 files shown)


src/command/check.rs
```diff
@@ -779,63 +780,65 @@ fn render_text_report(
...
-        &mut output,
-        "definitions: philosophies={} policies={} requirements={} features={}",
-        result.definition_counts.philosophies,
-        result.definition_counts.policies,
-        result.definition_counts.requirements,
-        result.definition_counts.features
-    )
-    .expect("writing to String must succeed");
-    if let Some(summary) = text_summary {
+    if !quiet_success {
```
tohboeh5, line 798: This behavior change looks intentional, but it now contradicts the user-facing guide. `docs/guide/getting-started.md` still says `--quiet` keeps the validation summary while only hiding the next-step block, so merging this as-is makes the guide wrong for first-time users. Please update the guide (and any mirrored docs) in the same PR.

tohboeh5:
## Overview
The implementation is tight and the focused test update matches the intended CLI behavior, but I left one blocking inline comment. The runtime now suppresses the success summary on clean `--quiet` runs, while the getting-started guide still tells users that the summary remains visible, so the documentation needs to move with the behavior change.

PR #392, opened 2026-04-18
fix: add next-step guidance for validation failures

tohboeh5:
## Summary
- add a failure-only next-step block to text validation output
- point users to show, filtered validate reruns, and the browser UI after validation errors
- cover the new guidance in the validation command regression test

## Testing
- cargo test --test check_command check_command_reports_missing_definition_links
- cargo run -- validate .
- cargo run -- validate tests/fixtures/workspaces/failing

### DIFF
--- a/src/command/check.rs
+++ b/src/command/check.rs
@@ -892,8 +892,9 @@ fn render_text_report(
         }
     }
 
+    let workspace_arg = shell_quote_path(workspace_arg);
+
     if overall_success && result.issues.is_empty() && !quiet {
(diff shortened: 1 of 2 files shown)


tohboeh5:
## Overview
The new failure-tailored next-step block is a solid UX improvement: it gives users an immediate recovery path instead of stopping at the raw diagnostics. The suggestions are scoped to the kinds of follow-up actions people actually take after a failed validation run, and I did not find a blocking issue in the diff.

PR #405, opened 2026-04-18
docs: align getting-started concepts tip

### LINKED ISSUE #404: [Feature]: Align the getting-started concepts tip with the README chooser
The README chooser now clearly says some readers can stay in the README quick start when the in-page refresher is enough. `docs/guide/getting-started.md` still opens with a tip that says “New to syu? Read syu concepts first,” which reads more like a blanket prerequisite than a conditional suggestion.

That tension makes the newcomer path less crisp than the README now promises.

## What change do you want?
Align the getting-started intro with the README chooser by making the concepts guide recommendation conditional and explicit about who actually needs it.

For example, the guide could say that readers who already understand the four layers can continue directly, while readers who want the deeper mental model should open `concepts.md` first.

## Specification impact
Docs-only, but it affects the project’s low-friction onboarding story and the guidance hierarchy between README, concepts, and getting-started.

## Alternatives considered
Leaving the current tip is technically safe, but it nudges every newcomer toward extra reading even after the README introduced a shorter path.

tohboeh5:
## Summary
- make the getting-started concepts tip conditional instead of sounding like a blanket prerequisite
- keep the guide aligned with the README's low-friction newcomer chooser

- Closes #404

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -2,9 +2,10 @@
 
 <!-- FEAT-DOCS-001 -->
 
-:::tip New to syu?
-New to `syu`? Read [syu concepts](./concepts.md) first to understand the four
-layers before you start editing YAML.
+:::tip New to `syu`?
+If you already understand the four-layer model, you can continue straight to the
+guided setup below. If you want the fuller mental model before you start editing
+YAML, read [syu concepts](./concepts.md) first.
 :::
 
 Need a different level of guidance?
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -3,9 +3,12 @@
 <!-- FEAT-DOCS-001 -->
 
 :::tip New to `syu`?
-If you already understand the four-layer model, you can continue straight to the
-guided setup below. If you want the fuller mental model before you start editing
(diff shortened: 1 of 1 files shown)


docs/guide/getting-started.md
```diff
@@ -2,9 +2,10 @@
 
 <!-- FEAT-DOCS-001 -->
 
-:::tip New to syu?
-New to `syu`? Read [syu concepts](./concepts.md) first to understand the four
-layers before you start editing YAML.
+:::tip New to `syu`?
+If you already understand the four-layer model, you can continue straight to the
```
tohboeh5: This is more conditional than before, but it still funnels readers who already know the four layers into the guided page. The README chooser frames quick start as the low-friction path for exactly that audience, so the tip still leaves the two entry points slightly out of sync. Please either mention the README quick start here as the fast path or narrow this sentence so it does not sound like the guide is the default continuation for returning readers.

tohboeh5:
## Overview

The direction is good: making the concepts link optional is exactly the right UX move for newcomers. I left one blocking inline note because this tip still nudges already-oriented readers deeper into the guide instead of reflecting the README's quick-start vs guided-walkthrough split, so the repo's two first-run entry points still do not tell the same story.

PR #406, opened 2026-04-18
docs: scope early alpha migration support

### LINKED ISSUE #403: [Feature]: Backfill or explicitly scope the undocumented alpha.1-alpha.4 migration path
`docs/guide/migration.md` currently has a table row for `alpha.1–alpha.4` that says `not yet documented`. That leaves early adopters without a clear answer about whether those versions are intentionally unsupported, too old to care about, or simply missing migration notes.

A migration page that starts with an undocumented range creates uncertainty about how much upgrade help the project actually promises.

## What change do you want?
Either backfill the migration notes for the early alpha releases or explicitly scope the migration guide so readers understand why `alpha.1–alpha.4` are omitted and what they should do instead.

## Specification impact
Documentation-only, but it affects the migration/upgrade story that the README now advertises directly.

## Alternatives considered
The current placeholder is honest, but it still leaves the user's next step ambiguous.

tohboeh5:
## Summary
- explain why alpha.1-alpha.4 do not have step-by-step migration notes
- give early adopters a concrete manual upgrade path toward the supported alpha.5+ layout
- replace the ambiguous compatibility-table placeholder with explicit scope

- Closes #403

### DIFF
--- a/docs/guide/migration.md
+++ b/docs/guide/migration.md
@@ -121,9 +121,9 @@ This guide only has release-by-release notes starting at `alpha.5`. The earlier
 layout stabilized, so this repository does **not** maintain step-by-step upgrade
 instructions for those versions. If you are upgrading from one of those early
 alphas, treat `alpha.5` as the first supported landing point: compare your
-workspace against a freshly generated `alpha.7` scaffold, make the required
-`spec.root` and validation-config updates, then run `syu validate .` until the
-workspace is green.
+workspace against a freshly generated scaffold from the version you are
+upgrading to, make the required `spec.root` and validation-config updates, then
+run `syu validate .` until the workspace is green.
 
 | syu version | `spec.root` default | `require_reciprocal_links` | `report.output` |
 |---|---|---|---|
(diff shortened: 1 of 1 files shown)


docs/guide/migration.md
```diff
@@ -116,9 +116,18 @@ for the full contribution workflow.
 
 ## Version compatibility summary
 
+This guide only has release-by-release notes starting at `alpha.5`. The earlier
+`alpha.1`-`alpha.4` builds shipped before the current migration notes and docs
+layout stabilized, so this repository does **not** maintain step-by-step upgrade
+instructions for those versions. If you are upgrading from one of those early
+alphas, treat `alpha.5` as the first supported landing point: compare your
+workspace against a freshly generated `alpha.7` scaffold, make the required
```
tohboeh5: Hard-coding `alpha.7` here will go stale on the next release, and then the manual-upgrade path will tell early adopters to compare against an outdated scaffold. Please point people at a freshly generated scaffold from *their installed/current version* instead of baking the current release number into the prose.

tohboeh5:
## Overview

The scope clarification is helpful, and replacing the placeholder compatibility row with a concrete manual path is a big improvement. I left one blocking inline note because anchoring that advice to `alpha.7` makes the docs self-expire as soon as the next release ships.

PR #407, opened 2026-04-18
docs: add a docs-first trace adapter example

### LINKED ISSUE #402: [Feature]: Add a pattern-based adapter example for docs-first repositories
The trace adapter docs cover more than just Rust, Python, and TypeScript/JavaScript. Built-in adapters also exist for Shell, YAML, JSON, Markdown, and Gitignore, but the checked-in examples are still heavily code-centric (`rust-only`, `python-only`, `polyglot`) plus the special `go-only` workaround.

That leaves a gap for repositories that mostly want documentation/configuration traceability or pattern-based file ownership rather than rich code-symbol inspection.

## What change do you want?
Add a small checked-in example for one of the pattern-based adapters (for example Shell + Markdown, or YAML + Markdown) that demonstrates:
- file/symbol mappings for non-rich adapters
- when `doc_contains` is unavailable
- how `symbols: ["*"]` or file-level ownership should be used carefully

## Specification impact
This is mostly an examples/docs change, but it reinforces the repository's language-agnostic product story.

## Alternatives considered
The capability matrix explains the limitations, but a runnable example would make the trade-offs much easier to understand.

tohboeh5:
## Summary
- add a checked-in docs-first example workspace for shell, markdown, and YAML pattern-based adapters
- link the trace adapter guide to the new runnable example
- cover the new example in repository-quality and example-workspace validation

- Closes #402

### DIFF
--- a/docs/guide/trace-adapter-support.md
+++ b/docs/guide/trace-adapter-support.md
@@ -58,6 +58,10 @@ they do **not** participate in the repository-wide strict ownership scan.
   `file` + `symbols` (or `symbols: ["*"]` when one file intentionally belongs
   to one item), but do not expect doc-comment inspection or strict ownership
   inventory.
+- Want a runnable reference for that lighter mapping? Start with the
+  [`examples/docs-first` workspace on GitHub](https://github.com/ugoite/syu/tree/main/examples/docs-first),
+  which demonstrates a shell symbol trace, markdown-backed requirement
+  evidence, and cautious wildcard YAML ownership in one small workspace.
 
 Even in rich-inspection languages, wildcard traces cannot use `doc_contains`
 because `symbols: ["*"]` does not point to one inspectable symbol.
--- a/README.md
+++ b/README.md
@@ -582,6 +582,7 @@ missing files.
 
 The repository ships working example projects:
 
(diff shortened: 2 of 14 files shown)


docs/guide/trace-adapter-support.md
```diff
@@ -58,6 +58,10 @@ they do **not** participate in the repository-wide strict ownership scan.
   `file` + `symbols` (or `symbols: ["*"]` when one file intentionally belongs
   to one item), but do not expect doc-comment inspection or strict ownership
   inventory.
+- Want a runnable reference for that lighter mapping? Start with the
```
tohboeh5, line 61: This makes the example reachable from the trace-adapter guide, but the repository's example-vs-template guide is still where we explain *which* checked-in examples exist and how to choose among them. Without a matching update there, `docs-first` stays hard to discover from the guide that newcomers are already sent to when deciding between templates and examples.

tohboeh5:
## Overview

I like the example itself — it fills a real gap around docs-first, pattern-based adapters and gives the trace-adapter guide something runnable to point at. I left one blocking inline note because the discoverability work is incomplete: the new workspace is linked here, but the dedicated examples/templates guide still won't surface it to readers making that choice.

PR #408, opened 2026-04-18
docs: cross-link unsupported-language guidance to go-only

### LINKED ISSUE #397: [Feature]: Cross-link unsupported-language docs to the go-only example
The repository now has a strong `examples/go-only` workaround story for unsupported languages, but the other unsupported-language docs do not consistently point to it. `docs/guide/trace-adapter-support.md`, `docs/guide/configuration.md`, and `docs/guide/troubleshooting.md` all explain current limitations, yet a reader can still miss the concrete checked-in workaround example.

That keeps unsupported-language adoption more theoretical than it needs to be.

## What change do you want?
Cross-link the unsupported-language guidance pages to the `go-only` example (and any future unsupported-language examples) wherever they explain current adapter limitations or workarounds.

## Specification impact
Mostly docs/guide changes, potentially with an examples-and-templates cross-link policy if the project wants this to stay consistent.

## Alternatives considered
The current guides are technically accurate, but they make users reconstruct the workaround themselves instead of pointing to the checked-in example repository shape.

tohboeh5:
## Summary
- link unsupported-language guidance pages to the checked-in go-only workaround example
- make the migration path from unsupported languages more concrete for readers

- Closes #397

### DIFF
--- a/docs/guide/trace-adapter-support.md
+++ b/docs/guide/trace-adapter-support.md
@@ -67,3 +67,7 @@ because `symbols: ["*"]` does not point to one inspectable symbol.
 Unsupported adapters such as `csharp` still raise `SYU-trace-language-001`.
 Keep those repositories connected through the spec layers first, and only add
 language-specific code traces once adapter support lands.
+If you need a checked-in workaround shape today, study the
+[`examples/go-only` workspace on GitHub](https://github.com/ugoite/syu/tree/main/examples/go-only),
+which keeps real Go files in the repository while validated trace evidence stays
+anchored in a supported markdown file.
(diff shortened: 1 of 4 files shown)


docs/guide/trace-adapter-support.md
```diff
@@ -67,3 +67,7 @@ because `symbols: ["*"]` does not point to one inspectable symbol.
 Unsupported adapters such as `csharp` still raise `SYU-trace-language-001`.
 Keep those repositories connected through the spec layers first, and only add
 language-specific code traces once adapter support lands.
+If you need a checked-in workaround shape today, study the
```
tohboeh5, line 70: This adds the workaround link in advanced support docs, but the newcomer-facing getting-started guide still has an 'unsupported implementation languages' section with no pointer to the example. That means first-time adopters still miss the concrete workaround unless they happen to open one of the more specialized docs pages. Please wire the same example into `docs/guide/getting-started.md` as part of this cross-linking pass.

tohboeh5:
## Overview

Pointing unsupported-language readers at a concrete checked-in example is the right move, and the wording here is clear. I left one blocking inline note because the linking is still incomplete across user journeys: the advanced guides get the example, but the main getting-started flow does not.

PR #411, opened 2026-04-18
docs: clarify which examples match starter templates

### LINKED ISSUE #375: [Feature]: Clarify the relationship between example workspaces and starter templates

The README and example workspaces talk about starter templates and example repositories as separate concepts, but they do not clearly explain which examples correspond to `syu init --template` outputs and which are reference-only workspaces. Newcomers can see both surfaces without understanding how they connect.

Evidence:
- The Quick start flow mentions `syu init --template ...` but does not point to the corresponding example workspace as a concrete reference.
- The README example-workspaces section lists examples without saying whether they are scaffoldable templates, learning references, or both.
- Example README files focus on the workspace itself, not on whether it matches an init template.

### What change do you want?

Clarify in the README and example README files which examples correspond to starter templates. For example:
- add a short sentence in the README examples section that explains which entries match `syu init --template` outputs
- add one line in each example README saying whether it is the direct output of a starter template or a reference-only workspace
- link the Quick start template walkthrough back to the matching example folder

### Specification impact

This is mainly a documentation/examples change, but it may touch docs feature language around discoverability and onboarding.

### Alternatives considered

A full examples-versus-templates matrix could help later, but the immediate low-cost improvement is to connect the existing docs and examples more explicitly.

### Before submitting

tohboeh5:
## Summary
- explain in the README which checked-in examples map directly to starter templates
- label each example README as template-backed or reference-only

- Closes #375

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,6 +587,10 @@ The repository ships working example projects:
 - [`examples/python-only`](examples/python-only)
 - [`examples/polyglot`](examples/polyglot)
 
+`rust-only`, `python-only`, and `polyglot` match `syu init --template ...`
+starters directly. `go-only` is a reference-only workaround example for an
+unsupported implementation language.
+
 Each one is validated in the automated test suite. If you are deciding between a
 checked-in example and a scaffold template, start with
 [`docs/guide/examples-and-templates.md`](docs/guide/examples-and-templates.md).
--- a/README.md
+++ b/README.md
@@ -586,10 +586,12 @@ The repository ships working example projects:
 - [`examples/rust-only`](examples/rust-only)
 - [`examples/python-only`](examples/python-only)
 - [`examples/polyglot`](examples/polyglot)
+- [`examples/team-scale`](examples/team-scale)
 
 `rust-only`, `python-only`, and `polyglot` match `syu init --template ...`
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -587,6 +587,10 @@ The repository ships working example projects:
 - [`examples/python-only`](examples/python-only)
 - [`examples/polyglot`](examples/polyglot)
 
+`rust-only`, `python-only`, and `polyglot` match `syu init --template ...`
+starters directly. `go-only` is a reference-only workaround example for an
+unsupported implementation language.
+
```
tohboeh5, line 610: This paragraph clarifies the template-backed examples, but it still doesn't tell readers what to make of `team-scale`, which is another checked-in example already present in the repository. As written, the README now distinguishes `go-only` from the starter-backed examples but still leaves one reference-only workspace unlabeled, so the map is not complete yet.

tohboeh5:
## Overview

I like the intent here — explicitly labeling template-backed vs reference-only examples makes the repository easier to scan. I left one blocking inline note because the classification is still incomplete: `team-scale` remains an unlabeled checked-in example, so readers still do not get a full picture from the README alone.

PR #412, opened 2026-04-18
chore: encode Node.js majors in package tooling

### LINKED ISSUE #398: [Feature]: Encode the documented Node.js version split in repo tooling
`CONTRIBUTING.md` documents a deliberate Node.js split: Node 25 for the browser app and Node 20 for the docs site. Today that strategy lives almost entirely in prose. `app/package.json` and `website/package.json` do not give contributors much tooling help to select the right major before they run installs or local checks.

That increases setup friction and makes it easier to end up with a confusing local mismatch that CI would have caught later.

## What change do you want?
Encode the documented Node.js expectations in repository tooling, for example with `engines`, `.nvmrc`, Volta metadata, or another checked-in mechanism that works for both `app/` and `website/`.

## Specification impact
Contributor workflow docs and possibly repository automation docs should describe whichever mechanism becomes the source of truth.

## Alternatives considered
Leaving the guidance in prose only keeps the repo simpler, but it pushes version selection mistakes onto every new contributor.

tohboeh5:
## Summary
- add checked-in .nvmrc files for the app and docs site
- encode each surface expected Node.js major in package.json engines
- point CONTRIBUTING.md at those files as the source of truth

- Closes #398


app/package.json
```diff
@@ -27,5 +27,8 @@
     "vite": "^8.0.8",
     "vite-plus": "^0.1.18"
   },
+  "engines": {
```
tohboeh5, line 30: Once these version numbers live in `.nvmrc`, `package.json#engines`, CONTRIBUTING, and the workflow files, we need an automated guard that they stay aligned. Right now this PR introduces multiple copies of the same Node major without any repository-quality assertion tying them together, so the newly documented 'source of truth' can still drift silently on the next bump.

tohboeh5:
## Overview

Encoding the expected Node majors in checked-in files is the right direction and makes local setup less guessy. I left one blocking inline note because the change adds several copies of the version numbers but no automated consistency check, so it increases the chance of silent drift instead of fully solving it.

PR #413, opened 2026-04-18
docs: clarify that quick start stays in the README

### LINKED ISSUE #255: [Feature]: Clarify the README Quick start prerequisites so it matches the shortest-path promise

tohboeh5:
## Summary
- make the README chooser say the quick-start path is self-contained
- explicitly mark concepts.md as optional extra context instead of a hidden prerequisite

- Closes #255

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ Pick the newcomer path that matches what you need next:
   shortest install-to-validate path, the in-page four-layer refresher is the
   fastest path through first workspace setup explained step by step.
 - **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .` and only need a short layer refresher before the first
-  commands.
+  to `syu validate .`, want the README itself to stay self-contained, and only
+  need a short layer refresher before the first commands.
 - **Tutorial**: follow [`docs/guide/tutorial.md`](docs/guide/tutorial.md) when you
   want a realistic end-to-end repository story instead of a short scaffold flow.
 - **Trace adapter matrix**: open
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,15 +177,16 @@ below. The rest of this section assumes you installed the published CLI.
 Stay in this README for the shortest install-to-validate path. If you skipped
 [`docs/guide/concepts.md`](docs/guide/concepts.md), use the
 [Why four layers?](#why-four-layers) section above as the refresher on
-`philosophy`, `policy`, `requirements`, and `features`.
+`philosophy`, `policy`, `requirements`, and `features`. There is no other
```
tohboeh5, line 190: This makes the README's quick-start path self-contained, but `docs/guide/getting-started.md` still opens by telling newcomers to read concepts first. That leaves the repo's two primary newcomer entry points disagreeing on whether there is any required pre-read, so the messaging is still inconsistent across surfaces. Please update the guide in the same PR or scope this claim more narrowly.

tohboeh5:
## Overview

The README copy is clearer after this change, especially the explicit 'no other required pre-read' sentence in the quick-start section. I left one blocking inline note because the docs-site entry path still says something different, so the cross-surface onboarding story is not actually consistent yet.

PR #414, opened 2026-04-18
docs: separate quick start from the guided walkthrough

### LINKED ISSUE #256: [Feature]: Split the README quick start and getting-started guide into clearly different first-run paths

tohboeh5:
## Summary
- sharpen the chooser copy so README quick start and getting-started serve different newcomer needs
- describe the README path as the compact command-card flow and the guide as the narrated walkthrough

- Closes #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,10 @@ verified.
 
 Pick the newcomer path that matches what you need next:
 
-- **Getting started**: if you are using `syu` for the first time and want the
-  shortest install-to-validate path, the in-page four-layer refresher is the
-  fastest path through first workspace setup explained step by step.
-- **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .` and only need a short layer refresher before the first
-  commands.
+- **Getting started**: if you are using `syu` for the first time, the in-page four-layer refresher is the hand-off into the guided walkthrough that keeps the first workspace setup explained step by step, including why the manual YAML edits matter before validation.
+- **Quick start**: stay in this README when you want the shortest path from
+  install to `syu validate .`, prefer a compact command card, and only need a
+  short layer refresher before the first commands.
 - **Tutorial**: follow [`docs/guide/tutorial.md`](docs/guide/tutorial.md) when you
   want a realistic end-to-end repository story instead of a short scaffold flow.
 - **Trace adapter matrix**: open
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -31,12 +31,10 @@ verified.
 
 Pick the newcomer path that matches what you need next:
 
-- **Getting started**: if you are using `syu` for the first time and want the
-  shortest install-to-validate path, the in-page four-layer refresher is the
-  fastest path through first workspace setup explained step by step.
-- **Quick start**: stay in this README when you want the shortest path from install
-  to `syu validate .` and only need a short layer refresher before the first
-  commands.
+- **Getting started**: if you are using `syu` for the first time, the in-page four-layer refresher is the hand-off into the guided walkthrough that keeps the first workspace setup explained step by step, including why the manual YAML edits matter before validation.
```
tohboeh5: The wording split is better here, but the hosted docs entry point still only exposes the guided 'Get started' path. Until the docs-site home also offers the compact README-first option, newcomers arriving through the site still won't see the distinction this copy is trying to establish between the two onboarding modes.

tohboeh5:
## Overview

I agree with the goal here: quick start and getting started should feel like two intentionally different paths instead of near-duplicates. I left one blocking inline note because the distinction only exists in the checked-in docs right now — the docs-site landing page still funnels everyone into the guided path, so the user experience is inconsistent depending on where people enter.

PR #425, opened 2026-04-18
docs: refresh the go-only example for built-in adapter support

### LINKED ISSUE #415: [Bug]: Refresh the go-only example after built-in Go adapter support landed
### What area is affected?

### What happened?

`examples/go-only/README.md` still says the repository needs a workaround "before Go gets a built-in trace adapter" and that direct Go tracing is future work. That is no longer true on `main`: the checked-in trace adapter guide now lists `go` as a built-in adapter with pattern-based symbol validation and strict ownership coverage.

This makes one of the most visible example READMEs actively contradict the current product surface.

The `go-only` example should explain the current Go support accurately, or the example should be repurposed so it no longer claims to be standing in for a missing adapter.

### syu version

main

### Environment

Repository checkout; compared `examples/go-only/README.md` with `docs/guide/trace-adapter-support.md`

### Steps to reproduce

1. Open `examples/go-only/README.md`.
2. Read the opening paragraphs and the "Current-day workaround trace anchors" section.
3. Open `docs/guide/trace-adapter-support.md`.
4. Compare the `Go` row and the strict coverage section with the example README copy.

- `examples/go-only/README.md` says "before Go gets a built-in trace adapter"
- the same file says "Direct Go tracing is future work"
- `docs/guide/trace-adapter-support.md` says `Go | ... | ✅ Pattern-based symbol matching | ❌ | ✅`

### Before submitting

tohboeh5:
## Summary\n- switch the go-only example from markdown workaround traces to direct Go adapter mappings\n- update the example README so it describes current Go support accurately\n\n## Linked issue or specification\n- Closes #415\n

### DIFF
--- a/examples/go-only/README.md
+++ b/examples/go-only/README.md
@@ -54,7 +54,6 @@ traceability: requirements=1/1 traces validated; features=1/1 traces validated
   real `.go` files instead of a markdown workaround.
 - **Pattern-based matching is enough here** — `syu` validates that the named Go
   symbols exist, even though the adapter does not inspect doc comments.
-- **Strict coverage can grow later** — this example keeps the smallest possible
-  shape, but the Go adapter also participates in
-  `validate.require_symbol_trace_coverage` when a repository wants stricter
-  ownership checks.
+- **Strict coverage is available elsewhere** — the Go adapter participates in
+  `validate.require_symbol_trace_coverage`, but this example keeps the minimal
+  `go/` layout and is not demonstrating the stricter ownership-inventory path.
--- a/examples/go-only/README.md
+++ b/examples/go-only/README.md
@@ -1,12 +1,12 @@
 # go-only example
 
-This example demonstrates a minimal Go-first workspace for repositories that
-want to adopt `syu` before Go gets a built-in trace adapter.
+This example demonstrates a minimal Go-first workspace using the built-in Go
+trace adapter.
 
 It contains one philosophy, one policy, one requirement, and one feature, plus
-one Go source file and one Go test file. The current workaround keeps the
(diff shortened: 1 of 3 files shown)


examples/go-only/README.md
```diff
@@ -39,21 +39,22 @@ definitions: philosophies=1 policies=1 requirements=1 features=1
...
-  `implemented`, but the validated `tests:` and `implementations:` mappings point
-  at this README because `go` is not a supported trace language yet.
-- **Direct Go tracing is future work** — once `syu` gains a Go adapter, the same
-  spec IDs can move from these markdown-backed anchors to the real `.go` files.
+- **Go files are traced directly** — the requirement and feature point at the
+  real `.go` files instead of a markdown workaround.
+- **Pattern-based matching is enough here** — `syu` validates that the named Go
+  symbols exist, even though the adapter does not inspect doc comments.
+- **Strict coverage can grow later** — this example keeps the smallest possible
+  shape, but the Go adapter also participates in
```
tohboeh5: This wording is a little too strong for this example. The Go adapter does participate in strict `validate.require_symbol_trace_coverage`, but the strict inventory docs currently describe `src/` + test locations, while this example keeps the code under `go/`. As written, readers can easily infer that turning strict coverage on in this example will work unchanged.

Could we either drop the strict-coverage claim here or add one sentence clarifying that this example still uses the minimal layout and is not demonstrating the stricter ownership-inventory path?

tohboeh5:
## Overview

I like the direction here: moving the Go-only example from a markdown workaround to the built-in Go adapter makes the example much more honest and much easier to trust. The main thing I flagged is the new strict-coverage wording in the README, which currently reads stronger than the example layout really demonstrates.

PR #427, opened 2026-04-18
docs: align the website package version with the CLI release

### LINKED ISSUE #417: [Bug]: Keep website package metadata aligned with the current syu version
### What area is affected?

### What happened?

`website/package.json` still reports `"version": "0.0.1-alpha.5"` while the current checked-in CLI/docs tree is `0.0.1-alpha.7`. That version drift is easy to miss during local docs-site work, but it makes the docs package metadata look stale even when the content has moved on.

Because the docs site is part of the checked-in contributor workflow, stale package metadata makes the automation surface feel less trustworthy than the CLI and app packages.

The docs-site package metadata should stay aligned with the repository's current shipped version, or it should be derived from the same source as the Rust crate version.

### syu version

main

### Environment

Repository checkout; inspected `website/package.json` and `Cargo.toml`

### Steps to reproduce

1. Open `website/package.json`.
2. Compare its `version` field with the root crate version in `Cargo.toml` / `Cargo.lock`.

- `website/package.json`: `0.0.1-alpha.5`
- root crate version: `0.0.1-alpha.7`

### Before submitting

tohboeh5:
## Summary
- update website/package.json so the docs-site package version matches the current repository release

- Closes #417

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -2,7 +2,7 @@
   "name": "syu-docs",
   "description": "FEAT-DOCS-002 Docusaurus site for the checked-in syu docs tree",
   "private": true,
-  "version": "0.0.1-alpha.5",
+  "version": "0.0.1-alpha.7",
   "scripts": {
     "generate": "python3 ../scripts/generate-site-docs.py",
     "prestart": "npm run generate",
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "syu-docs",
-  "version": "0.0.1-alpha.5",
+  "version": "0.0.1-alpha.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "syu-docs",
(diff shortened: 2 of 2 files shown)


website/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "syu-docs",
   "description": "FEAT-DOCS-002 Docusaurus site for the checked-in syu docs tree",
   "private": true,
-  "version": "0.0.1-alpha.5",
+  "version": "0.0.1-alpha.7",
```
tohboeh5, line 5: Updating `package.json` alone still leaves the checked-in lockfile at `0.0.1-alpha.5` (the root package version is duplicated in `website/package-lock.json`). After merge, the docs-site metadata is still internally inconsistent. Could you bump the lockfile root version in the same change?

tohboeh5:
## Overview

Good catch on the website package version drift. I left one follow-up because the checked-in lockfile still carries the old root package version, so the metadata would remain inconsistent after this lands.

PR #428, opened 2026-04-18
docs: pin the docs-site npm release in package metadata

### LINKED ISSUE #418: [Feature]: Add explicit package-manager metadata for the docs site

The browser app pins `packageManager` in `app/package.json`, but the docs site does not pin an npm version at all. That means contributors get stronger tooling guidance in `app/` than in `website/`, even though the docs site is also a checked-in Node surface with its own build pipeline.

This inconsistency makes docs-site setup more trial-and-error than it needs to be.

## What change do you want?

Add the same kind of package-manager guidance for `website/` that the browser app already has. That could be a `packageManager` field, Volta metadata, or another checked-in mechanism that keeps the docs-site npm expectations explicit.

## Specification impact

Contributor workflow docs and docs-site automation guidance should point at whichever package-manager metadata becomes the source of truth.

## Alternatives considered

Leaving the docs site without package-manager metadata keeps one file smaller, but it keeps the contributor experience inconsistent across the repository's two Node surfaces.

## Before submitting

tohboeh5:
## Summary
- add explicit package-manager metadata to website/package.json
- point contributor guidance at the checked-in packageManager pin for both Node surfaces

- Closes #418


CONTRIBUTING.md
```diff
@@ -151,7 +151,8 @@ When you work locally, match the Node major to the surface you are changing:
 
 If you switch between both in one shell session, use a version manager such as
 `nvm`, `fnm`, or `Volta` so the browser app and docs site each run on the same
-major that CI expects.
+major that CI expects. Both checked-in Node surfaces also pin the expected npm
```
tohboeh5: Adding `packageManager` to the docs site makes sense, but this sentence reads as if repository automation already honors that pin. I could not find any Corepack/packageManager consumption in CI or local setup scripts; they still run plain `npm ci`. Could we soften the wording or wire automation up to the pinned npm release so the docs do not overstate enforcement?

tohboeh5:
## Overview

Adding explicit package-manager metadata to the docs site is a sensible improvement, and I like that the contributor guide now mentions it. My concern is that the new wording sounds stronger than the current automation behavior: I could not find CI or setup paths that actually consume the pin yet.

PR #429, opened 2026-04-18
docs: add a migration journey to the docs-site home

### LINKED ISSUE #419: [Feature]: Add a migration journey to the docs-site landing page

The docs-site landing page (`website/src/pages/index.js`) has a "Common journeys" section, but it does not include any migration or upgrade entry point even though the repository already ships a dedicated migration guide. That makes upgrade users invisible on the hosted landing page unless they already know the guide exists.

The README newcomer chooser and the docs site should both help returning users find the upgrade path quickly.

## What change do you want?

Add a migration / upgrade journey to the docs-site landing page so readers can jump straight to `docs/guide/migration.md` from the homepage.

## Specification impact

This is a docs-site UX change under the documentation feature layer.

## Alternatives considered

Leaving migration buried in the sidebar works for readers who are already exploring the docs tree, but it misses the landing-page audience that is explicitly choosing a next step.

## Before submitting

tohboeh5:
## Summary
- add a migration journey card to the docs-site landing page common journeys section
- point homepage readers directly at the checked-in migration guide

- Closes #419

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -50,12 +50,6 @@ const journeys = [
     to: '/docs/guide/getting-started'
   },
   {
-    title: 'Upgrade an existing workspace',
-    description:
-      'Jump straight to the migration guide when a new alpha release changes config, validation defaults, or guide structure.',
-    to: '/docs/guide/migration'
-  },
-  {
     title: 'Follow a full tutorial',
     description: 'Build a realistic four-layer example from scratch when you want the full repository story.',
     to: '/docs/guide/tutorial'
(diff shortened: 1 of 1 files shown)


website/src/pages/index.js
```diff
@@ -49,6 +49,12 @@ const journeys = [
     description: 'Scaffold a project, fill in the layered spec, and run validate without guessing the layout.',
     to: '/docs/guide/getting-started'
   },
+  {
+    title: 'Upgrade an existing workspace',
```
tohboeh5: The card itself makes sense, but putting it directly after `Start a workspace` interrupts the newcomer flow. The migration guide is for upgrading an existing workspace between alpha releases, not for the next step after first-time scaffolding. Could we move this closer to `Adopt an existing repository` or `Troubleshoot a broken workspace` so new and existing-user journeys stay grouped?

tohboeh5:
## Overview

The migration guide is worth surfacing from the docs-site home, especially while alpha releases can still reshape defaults. My main concern is information architecture: the new card is useful, but its current placement interrupts the newcomer flow instead of grouping existing-workspace journeys together.

PR #431, opened 2026-04-18
docs: add a browser UI journey to the docs-site home

### LINKED ISSUE #421: [Feature]: Add a visual-explorer journey to the docs-site landing page

The hosted docs landing page has no visual-explorer/browser-app journey, even though the repository ships a dedicated `syu app` guide and positions the browser UI as an onboarding path elsewhere. Readers who start on the docs site do not get the same visual-first entry point that the broader product story is moving toward.

That leaves one of the most approachable surfaces under-advertised on the site homepage.

## What change do you want?

Add a browser-app / visual-explorer journey card to the docs-site landing page that points to `docs/guide/app.md`.

## Specification impact

Docs-site UX only.

## Alternatives considered

Keeping the app guide in the sidebar works, but it hides the visual-first path from homepage readers who are explicitly looking for a next step.

## Before submitting

tohboeh5:
## Summary
- add a browser UI journey card to the docs-site landing page
- point homepage readers directly at the syu app guide

- Closes #421

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -50,6 +50,12 @@ const journeys = [
     to: '/docs/guide/getting-started'
   },
   {
+    title: 'Explore the browser UI',
+    description:
+      'Take the visual-first path through `syu app` when you want tabs, trace links, and validation context in one browser view.',
+    to: '/docs/guide/app'
+  },
+  {
     title: 'Follow a full tutorial',
     description: 'Build a realistic four-layer example from scratch when you want the full repository story.',
     to: '/docs/guide/tutorial'


website/src/pages/index.js
```diff
@@ -49,6 +49,12 @@ const journeys = [
     description: 'Scaffold a project, fill in the layered spec, and run validate without guessing the layout.',
     to: '/docs/guide/getting-started'
   },
+  {
+    title: 'Explore the browser UI',
```
tohboeh5, line 53: I like surfacing the browser UI path, but this increases the mismatch between the docs-site home and the checked-in README chooser. The hosted landing page now offers a visual-first path while the README still only exposes text-first paths, so onboarding depends on which entry point people happen to hit. Could we land this together with the README chooser update or otherwise acknowledge the asymmetry?

tohboeh5:
## Overview

I agree with giving visual-first users a direct route into the browser UI guide. The issue I called out is consistency: this improves the hosted-docs entry point, but the checked-in README chooser still lacks the same path, so the onboarding story keeps diverging by surface.

PR #432, opened 2026-04-18
docs: link the app guide back to both onboarding homes

### LINKED ISSUE #422: [Feature]: Let the app guide point back to both the docs-site home and the README chooser

`docs/guide/app.md` currently tells readers to return to the site landing page for text-first entry points, but it does not point back to the GitHub README chooser at all. That leaves repo-native readers and site-native readers with slightly different navigation models instead of one clearly connected onboarding map.

When guides are consumed both on GitHub and on the docs site, the app guide should acknowledge both entry surfaces explicitly.

## What change do you want?

Update the app guide intro so it links back to both the hosted-docs landing page and the GitHub README chooser, with clear wording about which one is better for each reader.

## Specification impact

Documentation UX only.

## Alternatives considered

A site-only backlink is simpler, but it makes the GitHub-first onboarding story harder to follow from the checked-in guide.

## Before submitting

tohboeh5:
## Summary
- add a site-home backlink to the app guide intro
- keep the README chooser link while clarifying which surface each entry point serves

- Closes #422

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -6,9 +6,11 @@
 
 If you started from the README newcomer chooser, this guide is the visual-first
 path back into the same onboarding flow. Return to the
+[`site landing page`](/) when you want the hosted-docs home with task-oriented
+journeys, or jump back to the
 [`README chooser on GitHub`](https://github.com/ugoite/syu/blob/main/README.md#choose-your-path)
-when you want the text-first install, tutorial, migration, or troubleshooting
-entry points instead.
+when you want the checked-in text-first install, tutorial, migration, or
+troubleshooting entry points instead.
 
 ```bash
 syu app .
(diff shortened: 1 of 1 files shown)


docs/guide/app.md
```diff
@@ -6,9 +6,11 @@
 
 If you started from the README newcomer chooser, this guide is the visual-first
 path back into the same onboarding flow. Return to the
+[`site landing page`](/) when you want the hosted-docs home with task-oriented
```
tohboeh5: The wording is better, but the `/` target only works in the hosted docs. This same checked-in Markdown is also read on GitHub from the README chooser flow, and there `/` sends people to the repository root instead of the docs-site landing page. Could we use an absolute docs-site URL here or otherwise avoid a link that changes meaning between the two surfaces?

tohboeh5:
## Overview

The extra backlinking is a real improvement: the app guide should absolutely help readers return to the broader onboarding surfaces. I left one portability note because the new root-relative link behaves differently in hosted docs versus the checked-in GitHub Markdown.

PR #433, opened 2026-04-18
docs: expose quick start from the docs-site hero

### LINKED ISSUE #423: [Feature]: Expose the compact quick-start path from the docs-site hero

The docs-site hero CTA sends everyone to the guided getting-started flow. That is helpful for first-time readers, but it hides the compact README-first quick-start path that GitHub readers can already choose.

As a result, the hosted landing page nudges every newcomer toward the longer walkthrough, even when they only want the minimal command-card path.

## What change do you want?

Add a homepage CTA or nearby chooser that exposes the compact README quick-start path alongside the existing guided "Get started" button.

## Specification impact

Docs-site UX only, but it affects the low-friction onboarding promise in the docs feature layer.

## Alternatives considered

Keeping only the getting-started CTA is simpler, but it makes the hosted-docs entry point more prescriptive than the GitHub README chooser.

## Before submitting

tohboeh5:
## Summary
- add a README quick-start CTA to the docs-site hero
- keep the guided getting-started CTA while exposing the compact command-card path

- Closes #423

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -101,6 +101,12 @@ export default function Home() {
               Get started
             </Link>
             <Link
+              className="button button--secondary button--lg"
+              to="https://github.com/ugoite/syu/blob/main/README.md#quick-start"
+            >
+              Quick start in the README
+            </Link>
+            <Link
               className="button button--outline button--lg siteHeroOutlineButton"
               to="/docs/generated/site-spec"
             >
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -101,10 +101,10 @@ export default function Home() {
               Get started
             </Link>
             <Link
-              className="button button--secondary button--lg"
+              className="button button--outline button--lg siteHeroOutlineButton"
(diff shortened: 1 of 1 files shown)


website/src/pages/index.js
```diff
@@ -100,6 +100,12 @@ export default function Home() {
             <Link className="button button--secondary button--lg" to="/docs/guide/getting-started">
               Get started
             </Link>
+            <Link
```
tohboeh5, line 103: Adding the quick-start CTA is useful, but giving it the same visual weight as `Get started` removes the audience cue the README chooser provides. First-time users can now jump straight into the compact path without seeing why they might want the fuller getting-started flow. Could we demote this button or add a short hint that distinguishes the two paths?

tohboeh5:
## Overview

Exposing the compact quick-start path from the site hero makes sense, and I like the goal of giving experienced users a faster route. I left one UX concern because the new button has the same weight as `Get started`, which makes it harder for first-time users to tell which path is actually safer for them.

PR #434, opened 2026-04-18
docs: add a failing-workspace recovery flow to the app guide

### LINKED ISSUE #424: [Feature]: Document the browser app recovery flow for failing workspaces

The browser UI guide explains the happy-path layout with passing-workspace screenshots, but it does not walk readers through the failing-workspace experience where the validation panel is actually most valuable. New users who open `syu app` because validation is already red do not get a docs example of how to read the issue summary, follow trace links, or recover from a failing state.

That leaves one of the most important real-world UI flows under-documented.

## What change do you want?

Add a failing-workspace section to `docs/guide/app.md` that shows what changes when validation issues are present and how the browser UI helps users recover.

## Specification impact

Documentation UX only.

## Alternatives considered

The current guide is useful for clean workspaces, but it does not teach the recovery flow that many first-time users care about most.

## Before submitting

tohboeh5:
## Summary
- document how to use the browser UI when a workspace starts with validation issues
- spell out the recovery flow from issue count to View ID jumps and refresh behavior

- Closes #424

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -161,6 +161,27 @@ Use the validation flow in this order:
 2. Read the selected issue message, location, suggestion, and rule reference.
 3. Click `View <ID>` to open the affected spec item in the main content area.
 
+### Recovering from a failing workspace
+
+When the workspace starts with validation issues already present, the same UI
+becomes a recovery checklist instead of a passive summary:
+
+1. Check the **issues** count in the sidebar first so you know whether you are
+   looking at one blocker or a longer queue.
+2. Open the validation row with the highest-severity problem you can fix next.
+3. Use the selected issue's **suggestion** and **rule reference** to understand
+   whether the problem is a missing backlink, a missing trace, or a broken file
+   reference.
+4. Click `View <ID>` to jump from the issue directly to the affected item, then
+   compare the YAML, links, and traces in the detail panel against the problem
+   description.
+5. After you edit the workspace, keep the tab visible or switch back to it so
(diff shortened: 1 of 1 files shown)


docs/guide/app.md
```diff
@@ -161,6 +161,27 @@ Use the validation flow in this order:
...
+
+1. Check the **issues** count in the sidebar first so you know whether you are
+   looking at one blocker or a longer queue.
+2. Open the validation row with the highest-severity problem you can fix next.
+3. Use the selected issue's **suggestion** and **rule reference** to understand
+   whether the problem is a missing backlink, a missing trace, or a broken file
+   reference.
+4. Click `View <ID>` to jump from the issue directly to the affected item, then
+   compare the YAML, links, and traces in the detail panel against the problem
+   description.
```
tohboeh5, line 177: This recovery flow should probably mention the stale-snapshot banner as well. Earlier in the guide we document that the app keeps the last successful snapshot visible if a refresh fails, so after fixing the workspace the issue count may stay red until refresh recovers. Without that reminder, users can read step 5 as if the count should always drop immediately and assume the app is wrong when it does not.

tohboeh5:
## Overview

The recovery checklist is a useful addition and fits the app guide well; it gives failing-workspace users a much more concrete sequence than the earlier generic validation notes. I left one follow-up because the new flow should still acknowledge the stale-snapshot banner and refresh-failure behavior documented elsewhere in the same guide.

PR #435, opened 2026-04-19
fix: make validate --quiet suppress success summaries

### LINKED ISSUE #380: [Bug]: Make `syu validate --quiet` suppress the validation summary as well as next steps
### What area is affected?

CLI or runtime behavior

### What happened?

`--quiet` currently suppresses the post-success "What to do next" block, but it still prints the workspace path, definition counts, and traceability summary. In CI and scripting contexts, that makes `--quiet` much noisier than the flag name suggests.

A `--quiet` validation run should produce truly minimal output, or at least clearly document the remaining output contract.

### syu version

main

### Environment

Linux, repository root checkout, Rust CLI from source

### Steps to reproduce

1. Run `cargo run -- validate . --quiet`
2. Observe that the validation summary still prints even though the flag is named `--quiet`
3. Compare with the expectation of minimal/no human-oriented summary output for scripting

The current behavior still prints workspace-level summary lines and trace counts; only the next-step guidance is suppressed.

### Before submitting

tohboeh5:
## Summary
- suppress the workspace summary, counts, and next-step text when validate --quiet succeeds
- keep the pass/fail headline and existing exit-code behavior unchanged
- document the quieter success contract and cover it with CLI tests

Closes #380

### DIFF
--- a/src/command/check.rs
+++ b/src/command/check.rs
@@ -855,7 +855,7 @@ fn render_text_report(
         .expect("writing to String must succeed");
     }
 
-    if let Some(filtered_view) = filtered_view {
+    if !quiet_success && let Some(filtered_view) = filtered_view {
         writeln!(&mut output, "filters: {}", filtered_view.describe_filters())
             .expect("writing to String must succeed");
         writeln!(
@@ -874,7 +874,8 @@ fn render_text_report(
                 writeln!(&mut output, "{line}").expect("writing to String must succeed");
             }
         }
-    } else if let Some(filtered_view) = filtered_view
+    } else if !quiet_success
+        && let Some(filtered_view) = filtered_view
         && filtered_view.total_issue_count > 0
     {
         writeln!(&mut output).expect("writing to String must succeed");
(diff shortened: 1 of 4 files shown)


src/command/check.rs
```diff
@@ -794,63 +795,65 @@ fn render_text_report(
...
-        &mut output,
-        "definitions: philosophies={} policies={} requirements={} features={}",
-        result.definition_counts.philosophies,
-        result.definition_counts.policies,
-        result.definition_counts.requirements,
-        result.definition_counts.features
-    )
-    .expect("writing to String must succeed");
-    if let Some(summary) = text_summary {
+    if !quiet_success {
```
tohboeh5, line 798: `quiet_success` only guards the top summary block, so a successful run like `syu validate . --quiet --severity error` still prints the filtered-view footer and the "no issues matched the active filters" message below. That breaks the new "suppress the text summary" contract for the common quiet+filter CI path.

tests/check_command.rs
```diff
@@ -232,7 +232,7 @@ fn check_command_prints_workspace_aware_next_steps_for_explicit_paths() {
 
 #[test]
 // REQ-CORE-001
-fn check_command_quiet_suppresses_next_step_guidance() {
+fn check_command_quiet_suppresses_success_summary_and_next_step_guidance() {
```
tohboeh5, line 235: This test only covers the plain passing path. A `--quiet` case combined with `--severity` or `--genre` would catch the footer leak above, because that successful filtered run still emits extra lines today.

tohboeh5:
Overall impression: the headline behavior is moving in the right direction, and the help/docs updates make the new `--quiet` contract easy to understand.

The main gap is that the implementation only suppresses the top success summary. In successful filtered runs, the filtered-view footer still prints, so `--quiet` is not actually quiet in one of the most script-friendly combinations. I also think the test coverage should include a quiet+filter case, because that is exactly where the regression is hiding.

PR #436, opened 2026-04-19
fix: add next-step guidance for validate failures

### LINKED ISSUE #382: [Bug]: Show targeted next-step guidance after validation failures
### What area is affected?

CLI or runtime behavior

### What happened?

`syu validate` prints a useful "What to do next" block after success, but failure cases drop straight from the issue list to process exit. First-time users who hit trace or graph errors get no hint about which follow-up command will help them inspect or narrow the problem.

Validation failures should also include lightweight next-step guidance, such as how to inspect one item with `syu show`, how to filter validation output by genre/severity, or how to open the browser UI for deeper investigation.

### syu version

main

### Environment

Linux, repository root checkout, Rust CLI from source

### Steps to reproduce

1. Introduce a validation failure in a workspace (for example, a missing reciprocal link or broken trace)
2. Run `cargo run -- validate .`
3. Observe that the command prints issues but no failure-mode guidance about what to inspect first

The success path already prints follow-up guidance, so the gap is specifically in the failure path UX.

### Before submitting

tohboeh5:
## Summary
- add a failure-mode next-step block to text validate output
- point users at show, severity/genre reruns, and the browser UI
- cover the failure guidance in CLI tests

Closes #382


src/command/check.rs
```diff
@@ -916,6 +917,29 @@ fn render_text_report(
             "  syu show <ID> {workspace_arg}    inspect a single spec item in detail"
         )
         .expect("writing to String must succeed");
+    } else if !overall_success {
```
tohboeh5: The success path respects `--quiet`, but this new failure branch does not. As written, `syu validate --quiet` suppresses follow-up guidance on success and then starts printing it again on failure, which makes the flag behave inconsistently across the two outcomes.

src/command/check.rs
```diff
@@ -916,6 +917,29 @@ fn render_text_report(
...
+        writeln!(&mut output).expect("writing to String must succeed");
+        writeln!(&mut output, "What to inspect next:").expect("writing to String must succeed");
+        writeln!(
+            &mut output,
+            "  syu show <ID> {workspace_arg}           inspect one requirement, feature, policy, or philosophy named above"
```
tohboeh5: This wording becomes misleading in filtered views. When the active filter hides every issue, the output explicitly says that no issues matched, so there is nothing "named above" for the user to inspect.

tohboeh5:
Overall impression: the new next-step block is a good UX improvement for the common failure case, and I like that it points users toward both narrower reruns and the browser UI.

The two places I would tighten before merge are consistency and wording. `--quiet` should stay quiet on failures too, and the `show <ID>` hint should avoid saying something is "named above" when a filtered view can intentionally show nothing above it.

PR #437, opened 2026-04-19
feat: add a templates discovery command

### LINKED ISSUE #356: [Feature]: Add a syu templates command for starter-template discovery
The CLI can bootstrap starter templates, but there is no dedicated command that tells users which templates exist or how they differ from the checked-in examples.

That makes discovery depend on help text and documentation reading instead of a small command users can run directly from the terminal.

### What change do you want?
Add a small CLI command such as `syu templates` that lists starter templates and, ideally, related checked-in examples.

- available template names and short descriptions
- whether each template maps to a checked-in example, an example-only path, or both
- output formats that match the rest of the CLI (plain text first, JSON optional if it fits the existing command family)

### Specification impact
This likely needs a new CLI feature entry plus docs and tests for the command surface.

### Alternatives considered
Relying on `syu init --help` keeps the command surface smaller, but it also hides an important discovery workflow inside a longer scaffold-oriented help page.

### Before submitting

tohboeh5:
## Summary
- add a templates command that lists the built-in starter templates
- document the new discovery flow in the README, getting-started guide, and self-hosted spec
- cover the command in CLI and repository-quality tests

Closes #356


src/command/templates.rs
```diff
@@ -0,0 +1,139 @@
...
+            "{}",
+            serde_json::to_string_pretty(&JsonTemplatesOutput { templates })
+                .expect("serializing templates output to JSON should succeed")
+        ),
+    }
+
+    Ok(0)
+}
+
+fn starter_template_catalog() -> Vec<TemplateCatalogEntry> {
```
tohboeh5: Could we derive this catalog from one shared metadata source instead of hard-coding the template list and descriptions a second time here? `init.rs` already owns the authoritative template branches, so every new starter now has to update both files or the discovery command can drift away from the actual scaffold behavior.

tohboeh5:
Overall impression: this is a nice UX addition. The command is easy to discover, the docs are updated in the right places, and the text/JSON coverage makes the behavior clear.

My only real concern is long-term drift: the template metadata now lives in the scaffolding code and in the discovery command, so future starter-template work has two authoritative places to keep in sync. If that gets centralized, the rest of the change looks solid.

PR #438, opened 2026-04-19
feat: distinguish non-instrumented coverage states

### LINKED ISSUE #350: [Feature]: Distinguish non-instrumented traces from zero coverage in spec summaries
The new requirement/feature coverage summary uses `n/a` for multiple situations: files outside Rust line instrumentation, features with no Rust implementation files, and requirements that only point at non-Rust evidence.

That makes the report harder to read because reviewers cannot distinguish "this path is not line-instrumented" from "this path should have data but does not."

### What change do you want?
Make the spec-coverage summary distinguish non-instrumented traces from zero or missing coverage.

- separate labels such as `not instrumented` / `no Rust files` / `0.0%`
- wording that keeps CI reviewers from over-reading `n/a`
- stable formatting for the uploaded Markdown artifact and step summary

### Specification impact
This extends the repository quality/reporting surface and may need docs for the new report semantics.

### Alternatives considered
Keeping a single `n/a` state is simpler, but it throws away context that reviewers need when the report spans mixed-language evidence.

### Before submitting

tohboeh5:
## Summary
- distinguish missing refs, missing Rust files, and non-instrumented paths in the spec coverage summary
- keep percentage output only for genuinely instrumented coverage data
- update the summary script without changing the existing report structure

Closes #350

### DIFF
--- a/scripts/ci/write-spec-coverage-summary.py
+++ b/scripts/ci/write-spec-coverage-summary.py
@@ -45,7 +45,7 @@ def run_syu_json(repo_root: Path, *args: str) -> dict:
 
 def percent_string(covered: int, total: int) -> str:
     if total == 0:
-        return "n/a"
+        return "0.0% (0/0)"
     return f"{covered * 100.0 / total:.1f}% ({covered}/{total})"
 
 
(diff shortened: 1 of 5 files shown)


scripts/ci/write-spec-coverage-summary.py
```diff
@@ -82,16 +82,38 @@ def load_spec_items(
...
         if stats is None:
             continue
+        instrumented_paths += 1
         covered += stats[0]
         total += stats[1]
-    return covered, total
+    return covered, total, instrumented_paths
+
+
+def coverage_label(
```
tohboeh5, line 101: The new labels are much more informative, but this logic is subtle enough that I would really like a focused regression test around it. Right now the behavior change lives entirely in the script, so a future refactor could collapse `no Rust files` / `not instrumented` / `0.0% (0/0)` back together without any test failing.

tohboeh5:
Overall impression: the state split is useful and the output is much easier to reason about than the old catch-all `n/a`. The implementation also stays nicely local to the coverage-summary script.

My one reservation is maintainability: these user-facing labels encode a few distinct edge cases now, but there is no dedicated regression test pinning them down. I would be more comfortable merging it once those states are locked in with a focused test or golden-output check.

PR #451, opened 2026-04-19
docs: consolidate README browser-app guidance

### LINKED ISSUE #445: [Feature]: Consolidate the README browser-app guidance into one canonical path

The README now explains the browser app in multiple places: command reference, a dedicated browser-app overview, and a separate development section. Each section is useful on its own, but together they make the entry story feel fragmented for new readers deciding whether they want `syu app`, screenshots, or contributor build steps.

That redundancy makes it harder to tell which section is the canonical first stop.

## What change do you want?

Consolidate the README browser-app story into a clearer hierarchy. For example:
- keep one newcomer-oriented overview that explains when to use `syu app`
- move deeper implementation/build details into the development section only
- cross-link the command reference and the dedicated guide instead of partially repeating the same explanation

## Specification impact

This is mainly a documentation and UX cleanup, but it should stay aligned with the browser-app feature docs and contributor workflow requirements.

## Alternatives considered

Leaving the duplication in place works, but it keeps the browser story more verbose than the rest of the README's path-oriented onboarding.

## Before submitting

tohboeh5:
## Summary
- keep the newcomer-facing browser-app overview in one README section instead of repeating command details
- move source/bundle and contributor build details into the development section where they belong
- keep the command reference and browser UI guide cross-linked so readers can choose the right depth

Closes #445

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,13 +486,18 @@ with `scripts/ci/check-generated-docs-freshness.sh`.
 
 ## Browser app
 
-Start with [`syu app`](#syu-app) when you want richer browser-based exploration
-than the terminal-first `syu browse` flow. For a visual tour with annotated
-screenshots, see the [browser UI guide](docs/guide/app.md).
+Use [`syu app`](#syu-app) when you want a local browser UI for searching the
+four-layer graph, following links across related documents, and sharing the
+same live workspace data that `syu browse` exposes in the terminal.
 
-If you are deciding whether to use the browser UI or only need the CLI command,
-this section is the overview. The implementation and contributor build details
-live under [Browser app development](#browser-app-development).
+```bash
+syu app .
+syu app . --bind 127.0.0.1 --port 3000
+```
+
+For a visual tour with annotated screenshots, see the
+[browser UI guide](docs/guide/app.md). Contributor-only implementation and
+build details live under [Browser app development](#browser-app-development).
 
 ## VS Code extension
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -484,24 +486,13 @@ with `scripts/ci/check-generated-docs-freshness.sh`.
...
-
-When contributors change browser app sources or build inputs, they should run
-`scripts/ci/check-browser-app-freshness.sh`. That flow regenerates the local
-`app/src/wasm` bridge, typechecks the browser app, and builds a fresh local
-`app/dist/` artifact the same way CI does before uploading it.
-
-Want the visual tour first? See the [browser UI guide with annotated
-screenshots](docs/guide/app.md).
+If you are deciding whether to use the browser UI or only need the CLI command,
+this section is the overview. The implementation and contributor build details
```
tohboeh5: This wording is a bit circular for a newcomer: the section says it is "the overview", but it mostly redirects away instead of actually summarizing what the browser app is and when someone should prefer it. Either add 1-2 sentences of real overview content here or rewrite this line as a simple pointer to `syu app`, the browser UI guide, and the contributor-facing build section.

tohboeh5:
Overall impression: separating user-facing browser-app guidance from contributor build details is the right direction, and moving the technical build steps lower in the README makes the top-level flow easier to scan. The weak spot is the new bridge sentence in the `Browser app` section: it talks about the section as an overview without actually giving the reader an overview, so the section reads more like meta-navigation than guidance. Tightening that wording or adding a short product-level summary would make the restructure much easier to follow.

PR #452, opened 2026-04-19
docs: align README JavaScript setup with npm ci flows

### LINKED ISSUE #442: [Bug]: Align README JavaScript setup snippets with the documented npm ci flows
### What area is affected?

### What happened?

The main README still recommends `npm install` in the browser-app and docs-site development sections even though contributor guidance elsewhere has moved toward explicit `npm ci` flows and separate Node-major expectations for each surface.

That leaves new contributors with two different stories: the landing docs suggest mutable installs, while the stricter contributor flow expects lockfile-driven setup and surface-specific runtime policy.

The README's JavaScript setup snippets should match the repository's documented contributor policy, including when to use `npm ci` versus `npm install` and which Node major each surface expects.

### syu version

main

### Environment

Repository checkout, README development sections for `app/` and `website/`.

### Steps to reproduce

1. Open `README.md`.
2. Jump to the browser-app and docs-site development/setup snippets.
3. Compare the `npm install` instructions there with the stricter contributor guidance in `CONTRIBUTING.md`.
4. Notice that the README does not reflect the current lockfile-oriented flow or the split Node-major policy.

Current README snippets include:

```shell
cd app
npm install
...
cd website
npm install
```

### Before submitting

tohboeh5:
## Summary
- switch the README browser-app and docs-site setup snippets from `npm install` to `npm ci`
- call out the current Node majors that CI uses for the app and docs-site surfaces
- keep the landing README aligned with the stricter contributor workflow docs

Closes #442


README.md
```diff
@@ -658,9 +658,12 @@ If you prefer to install `pre-commit` manually, `pipx install pre-commit` or
 
 ### Browser app development
```
tohboeh5, line 678: `CI currently validates ...` reads like an observational note, but for contributors this is effectively the expected local runtime too: the checked-in `engines` fields and `.nvmrc` files pin those majors. I would phrase this as an instruction (`Use Node 25 ...`) rather than a CI fact, otherwise readers can easily treat the version as optional and only discover the mismatch once installs or builds fail.

tohboeh5:
Overall impression: switching the README to `npm ci` and documenting the two Node majors is directionally correct, because it makes the checked-in setup path closer to what CI actually runs. The remaining weakness is tone and contract clarity: the new wording presents the Node versions as something CI happens to validate instead of the versions contributors are actually expected to use locally, even though the repository already encodes those expectations in `package.json` and `.nvmrc`.

Issue #439, opened 2026-04-19
[Bug]: Make repeated docs-site installs idempotent in reused worktrees


tohboeh5:
### What area is affected?

Repository automation or CI

### What happened?

Repeated docs-site validation in a reused checkout can fail before the build even starts because `npm --prefix website ci` trips over an existing `website/node_modules` tree with `ENOTEMPTY`. I hit this multiple times while validating docs-only branches back to back in the same repository checkout.

That makes the documented validation flow feel non-idempotent: running the same command twice in a healthy worktree can require manual cleanup that the docs never mention.

### What did you expect to happen?

The documented docs-site validation path should be safe to re-run in the same checkout without manual `node_modules` cleanup, or the repository should provide a cleanup-aware wrapper that makes the happy path deterministic.

### syu version

main

### Environment

Linux, repository checkout with repeated `npm --prefix website ci` / `npm --prefix website run build` runs across multiple branches and worktrees.

### Steps to reproduce

1. Start from a checkout where `website/node_modules` already exists from a prior docs-site build.
2. Run `npm --prefix website ci` again after switching branches or re-running validation.
3. Observe intermittent `ENOTEMPTY` failures while npm removes existing package directories.

### Logs, screenshots, or failing output

```shell
npm error code ENOTEMPTY
npm error syscall rmdir
npm error path /workspaces/.../website/node_modules/@docusaurus
npm error ENOTEMPTY: directory not empty, rmdir '.../website/node_modules/@docusaurus'
```

### Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

PR #454, opened 2026-04-19
ci: make docs-site installs idempotent

tohboeh5:
## Summary\n- add a cleanup-aware docs-site dependency install script for reused worktrees\n- route the shared docs-site GitHub action and contributor docs through that deterministic install path\n- update spec traces and repository-quality assertions for the new automation\n\n## Testing\n- bash scripts/ci/install-docs-site-deps.sh\n- bash scripts/ci/install-docs-site-deps.sh\n- npm --prefix website run build\n- cargo test -q --test repository_quality\n- bash scripts/ci/quality-gates.sh\n- cargo run --quiet -- validate .\n\nCloses #439


scripts/ci/install-docs-site-deps.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
tohboeh5, line 1: Because this is a standalone script with a shebang and it now becomes the documented entrypoint for docs-site setup, please also mark it executable in git. Right now contributors can follow the README by invoking it through `bash`, but `./scripts/ci/install-docs-site-deps.sh` still fails with `Permission denied`, which is inconsistent with the rest of `scripts/ci/`.

README.md
```diff
@@ -674,11 +674,13 @@ checked-in `docs/` tree directly, and the published site is available at
 `https://ugoite.github.io/syu/`.
 
 ```bash
-cd website
-npm install
+bash scripts/ci/install-docs-site-deps.sh
 npm run start
```
tohboeh5: This command no longer matches the repository layout. `npm run start` only works from `website/`, so after replacing `cd website` with the wrapper script the next line should also switch to `npm --prefix website run start` (or add `cd website` back). As written, the documented flow fails from the repo root.

tohboeh5:
Overall impression: making docs-site installs idempotent is a useful fix, and I like that the PR carries the change through the workflow, docs, self-hosted spec, and repository-quality coverage instead of hiding it in one script. The remaining problems are around polish and newcomer ergonomics: the new helper should be executable like the other `scripts/ci/*` entrypoints, and the README command sequence now mixes a repo-root wrapper with a `website/`-relative `npm run start`, so the documented local dev flow breaks unless the reader already knows the package layout.

Issue #347, opened 2026-04-16
[Feature]: Add a go-only starter template to syu init


tohboeh5:
### What problem are you trying to solve?
The repository now ships `examples/go-only`, but `syu init --template ...` still only exposes the older starter set.

That creates a newcomer gap: the docs can point to a working Go example, but the CLI cannot scaffold the same shape directly. Users who discover the example still have to copy files by hand instead of starting from the flow the rest of the docs recommend.

### What change do you want?
Add a `go-only` starter template to `syu init`, or add a closely equivalent template name that produces the same minimal Go workspace shape as the checked-in example.

A good first scope would cover:
- `syu init --template go-only`
- help text and docs that mention the new template
- parity with the checked-in `examples/go-only` layout so docs and scaffolds stop drifting apart

### Specification impact
This likely touches contributor/onboarding docs plus the init feature/spec references.

### Alternatives considered
Keeping Go as an example-only path still helps readers, but it keeps the CLI's newcomer path inconsistent with the repository's documented examples.

### Before submitting
- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

PR #455, opened 2026-04-19
feat: add a go-only init template

tohboeh5:
## Summary\n- add `go-only` to the init/template catalog and scaffold matching Go starter files\n- update docs and self-hosted spec text so the Go example is a template-backed path\n- extend CLI and repository tests to cover the new template end-to-end\n\n## Testing\n- cargo test -q --test init_command --test templates_command --test help_command --test repository_quality\n- cargo run --quiet -- templates\n- cargo run --quiet -- init /tmp/syu-go-only-demo --template go-only\n- cargo run --quiet -- validate /tmp/syu-go-only-demo\n- npm --prefix website ci\n- npm --prefix website run build\n- bash scripts/ci/quality-gates.sh\n- cargo run --quiet -- validate .\n\nCloses #347


src/command/init.rs
```diff
@@ -509,7 +516,27 @@ fn scaffold_files(
...
         ),
-    ]
+    ];
+    files.extend(starter_source_files(template));
+    files
+}
+
+fn starter_source_files(template: StarterTemplate) -> Vec<(String, String)> {
+    match template {
+        StarterTemplate::GoOnly => vec![
```
tohboeh5, line 576: If this is meant to be a real `Go-first` starter, it still stops short of a runnable Go scaffold because it never creates a `go.mod`. The generated `go/app.go` and `go/app_test.go` files are enough for `syu validate`, but they still leave common first-step tooling (`go test`, editor language tooling, module-aware imports) in a half-bootstrapped state. I would either add a minimal module file here or tone the docs/template copy down from “Go-first starter” to “Go traceability starter”.

tohboeh5:
Overall impression: adding a built-in go-only template is directionally strong, especially because it closes the gap between the checked-in example and `syu init --template ...` discovery. The main weakness is that the scaffold currently creates traced Go source and test files without also creating the module metadata a Go-first repository normally needs, so the template is specification-ready but not quite turnkey from the language-tooling side.

PR #456, opened 2026-04-19
docs: centralize install snippet release references

### LINKED ISSUE #401: [Feature]: Generate install snippet release numbers from one shared source
The checked-in install docs repeat the current alpha tag (`v0.0.1-alpha.7`) in many places across `README.md` and `docs/guide/getting-started.md`. That repetition makes release-number drift likely whenever the next alpha ships.

Even when the commands are otherwise correct, a stale release number makes first-run docs feel unreliable.

## What change do you want?
Generate the versioned install snippets from one shared source, or otherwise centralize the current published release tag so README and guide snippets stay aligned.

## Specification impact
This likely touches docs generation, contributor docs, and maybe the release automation that already knows the published version.

## Alternatives considered
Manual edits are workable while the docs are small, but the same release tag is already duplicated enough that drift is becoming an attractive nuisance.

tohboeh5:
## Summary
- reuse the verified README release variable across shortcut installer examples
- turn getting-started into a canonical pointer to the maintained README install flows
- lock the new guide structure with repository quality assertions

Closes #401


README.md
```diff
@@ -149,32 +149,31 @@ documentation version. Use `SYU_VERSION=alpha`, `stable`, or an explicit
 version selector when you want that installer to fetch a different published
 package track after it starts.
 
+Reuse the `RELEASE` value from the verified flow above, or set it once in your
```
tohboeh5: This section is now the target of deep links from getting-started.md, but it still assumes the reader already saw the earlier `RELEASE=...` definition. All of the shortcut examples below use `${RELEASE}` without defining it here, so a reader who jumps straight to this anchor gets a non-functional snippet. Please make the section self-contained by adding an inline `RELEASE=...` example before the first shortcut block (or by defining it in that first block).

tohboeh5:
Overall impression: centralizing the installer snippets in the README is the right maintenance move, and linking getting-started back to one canonical source will reduce future drift. The remaining UX gap is that the new shortcut section is no longer self-contained: readers who arrive through a deep link do not see a concrete `RELEASE=...` definition before the `${RELEASE}` examples, so the shortest path now depends on reading the page strictly top-to-bottom.

PR #458, opened 2026-04-19
ci: enforce pinned npm versions

### LINKED ISSUE #441: [Feature]: Enforce packageManager pins with real setup automation

The repository now records npm versions in `packageManager`, but the pins are still passive metadata: the documented local flows and CI paths keep calling plain `npm`, and there is no Corepack/setup step that actually enforces the recorded package-manager release.

That means contributors can believe the repository is actively pinning npm behavior when the current automation only hints at the intended version.

## What change do you want?

Make the `packageManager` pins executable instead of informational. A good first scope would be:
- enable Corepack (or an equivalent setup step) in CI and contributor docs
- fail early when the wrong npm release is used
- keep the browser-app and docs-site surfaces aligned with whatever enforcement model the repository adopts

## Specification impact

This likely touches contributor workflow docs, repository automation, and the repository-quality requirements around deterministic tooling.

## Alternatives considered

Leaving the field as passive metadata is better than nothing, but it still lets actual installs drift away from the declared toolchain.

## Before submitting

tohboeh5:
## Summary
- add a shared script that installs or checks the exact npm version from packageManager
- use that script in CI, the docs-site build action, the browser freshness script, and the devcontainer bootstrap
- document the local pinned-npm flow and lock it into the self-spec plus repository-quality assertions

Closes #441


tests/repository_quality.rs
```diff
@@ -897,11 +906,14 @@ fn repository_ships_browser_app() {
     let app_playwright = read_file("app/tests/browser-app.spec.ts");
     let app_wasm = read_file("app/wasm/src/lib.rs");
```
tohboeh5, line 926: These new assertions make the pinned-npm story look comprehensive, but they still leave one real npm-install path uncovered: `build.rs` can auto-install `app/` dependencies during Rust builds, and that code path still does not call `scripts/ci/pinned-npm.sh`. In other words, CI and the documented helper flows may now align npm correctly while the embedded-app build path still runs with whatever global npm happens to be present. I would either wire the same pinning check into `build.rs` (or the script it shells out to) or narrow the docs/spec wording so it does not imply repository-wide enforcement yet.

tohboeh5:
Overall impression: pinning npm to the checked-in `packageManager` value is the right direction, and I like that this PR threads the new helper through CI, contributor setup, browser-app freshness checks, and the self-hosted spec/tests. The remaining gap is scope: the repository still has at least one npm-install path outside the newly pinned surfaces (`build.rs` for the embedded browser app), so the current docs/spec wording reads stronger than the actual enforcement boundary.

PR #476, opened 2026-04-19
feat: make app refresh polling adaptive (FEAT-APP-001)

### LINKED ISSUE #465: [Feature]: Make syu app refresh polling adaptive instead of a fixed 2s loop

The browser app refresh loop polls `/api/version` every two seconds with a fixed `setInterval`, and it does so for every visible tab without any adaptive backoff once the workspace is stable.

That is simple, but it creates a steady stream of requests for long-lived sessions and gives maintainers no way to tune the refresh cost for larger workspaces or slower environments.

## What change do you want?

Make syu app refresh polling adaptive instead of a hard-coded 2s loop. A good first scope would be:

- replace the fixed interval with a backoff-aware or event-aware refresh strategy
- keep the current refresh-when-visible behavior for active troubleshooting sessions
- consider exposing the polling cadence through configuration or an app flag if that still fits the project\'s simplicity goals

## Specification impact

This affects browser-app UX and performance expectations for the live workspace snapshot experience.

## Alternatives considered

Keeping the fixed 2s loop is easy to reason about, but it pays the same refresh cost in every repository regardless of size, stability, or how long the page stays open.

## Before submitting

tohboeh5:
## Summary
- replace the fixed 2s refresh interval with an adaptive timeout-based poller
- back off when the workspace stays stable and reset to the fast cadence after updates, failures, or tab focus changes
- keep the existing visible-tab refresh behavior for active troubleshooting sessions

- Closes #465

## Validation
- [x] `scripts/ci/validate-app.sh --e2e`

## Release notes
- [x] No release note needed

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -180,6 +180,9 @@ function App() {
       Math.min(REFRESH_POLL_MAX_MS, REFRESH_POLL_MIN_MS * 2 ** stablePollCount);
 
     const schedulePoll = (delay: number) => {
+      if (cancelled) {
+        return;
+      }
       if (timeoutId != null) {
         window.clearTimeout(timeoutId);
       }
@@ -227,6 +230,9 @@ function App() {
     };
 
     const handleVisibilityChange = () => {
+      if (cancelled) {
+        return;
+      }
       if (!document.hidden) {
         stablePollCount = 0;
         schedulePoll(REFRESH_POLL_MIN_MS);
@@ -241,6 +247,7 @@ function App() {
       document.removeEventListener("visibilitychange", handleVisibilityChange);
       if (timeoutId != null) {
         window.clearTimeout(timeoutId);
+        timeoutId = null;
       }
     };
   }, [isRefreshing, loadWorkspace, snapshotVersion]);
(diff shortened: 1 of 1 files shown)


app/src/App.tsx
```diff
@@ -186,20 +207,41 @@ function App() {
...
           // eslint-disable-next-line no-console
           console.error("Failed to poll app version for refresh", pollError);
           setRefreshError(formatRefreshFailure("check for workspace updates", pollError));
         }
+        schedulePoll(REFRESH_POLL_MIN_MS);
       }
-    }, 2000);
+    };
+
+    const handleVisibilityChange = () => {
```
tohboeh5, line 232: `handleVisibilityChange()` can still schedule a fresh timeout during cleanup because it never checks `cancelled`. If a visibilitychange event lands after cleanup sets `cancelled = true` but before the listener is removed, `schedulePoll()` installs a new timer and cleanup clears only the previous one. Please guard this handler (or reorder cleanup) so the effect cannot leave an orphaned timeout behind.

tohboeh5:
Overall impression: adaptive refresh polling is a meaningful UX and performance improvement over a fixed interval, and switching to timeout-based scheduling gives the app better control over backoff. I did find one cleanup race around `visibilitychange` rescheduling, so I left a line comment there before this is merged.

PR #477, opened 2026-04-19
docs: add a csharp fallback example

### LINKED ISSUE #460: [Feature]: Add a real unsupported-language example instead of reusing go-only

The docs still need a story for repositories whose language is genuinely unsupported, but the current workaround example is `go-only` even though Go now has a built-in adapter.

That leaves the repository without a checked-in example for the real unsupported-language case that the docs are trying to point people toward.

## What change do you want?

Add a truly unsupported-language example (for example a C#-backed or other non-supported repository) and update the docs to point at that example instead of reusing `go-only`.

A good first scope would be:

- keep `examples/go-only` focused on native Go support
- add one small reference workspace that demonstrates the fallback pattern for a still-unsupported language
- update `docs/guide/trace-adapter-support.md` and `docs/guide/examples-and-templates.md` to point at the new example

## Specification impact

This mostly affects examples and onboarding guidance, but it also sharpens how the repository explains supported vs unsupported trace-adapter promises.

## Alternatives considered

Keeping `go-only` as the workaround example was understandable before native Go support landed, but it is now teaching the wrong mental model.

## Before submitting

tohboeh5:
## Summary
- add a checked-in csharp-fallback example for unsupported-language adoption
- point unsupported-language docs to the new example and keep go-only focused on native Go support
- extend repository tests so the example stays validated

- Closing keyword issue: Closes #460
- Requirement / feature IDs:

## Validation
- [x] cargo run -- validate .

## Release notes
- [x] No release note needed

### DIFF
--- a/examples/csharp-fallback/syu.yaml
+++ b/examples/csharp-fallback/syu.yaml
@@ -1,0 +1,12 @@
+version: 0.0.1-alpha.7
+spec:
+  root: docs/syu
+validate:
+  default_fix: false
+  allow_planned: true
+  require_symbol_trace_coverage: false
+runtimes:
+  python:
+    command: auto
+  node:
+    command: auto
--- a/README.md
+++ b/README.md
@@ -621,6 +621,7 @@ missing files.
 
 The repository ships working example projects:
 
+- [`examples/csharp-fallback`](examples/csharp-fallback)
 - [`examples/docs-first`](examples/docs-first)
 - [`examples/go-only`](examples/go-only)
 - [`examples/rust-only`](examples/rust-only)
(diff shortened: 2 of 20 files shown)


examples/csharp-fallback/syu.yaml
```diff
@@ -0,0 +1,12 @@
+version: 0.0.1-alpha.7
```
tohboeh5, line 1: The new example hard-codes the released syu version, but `tests/repository_quality.rs` only checks the synced version for `examples/rust-only/syu.yaml`. Please add the same freshness assertion for `examples/csharp-fallback/syu.yaml`, otherwise this example can silently drift after the next release even though the README promises every checked-in example stays validated.

tohboeh5:
Overview review:

This is a strong documentation-oriented example: the new workspace tells an honest unsupported-language story, the guide updates point readers to the right fallback, and the example validates cleanly locally.

The main gap I found is maintainability: the repository quality test was extended to assert the new requirement and README anchors, but it still does not guard the examples `syu.yaml` version from drifting across future releases. That weakens the “checked-in examples stay trustworthy” promise over time, so I left one inline comment on that point.

Aside from that, the scope is focused and the replacement of `go-only` as the unsupported-language workaround reads much clearer for newcomers.

examples/csharp-fallback/syu.yaml
tohboeh5: Added matching freshness coverage for examples/csharp-fallback/syu.yaml in tests/repository_quality.rs so the example version now stays pinned to the current release alongside rust-only.

PR #479, opened 2026-04-19
feat: add app search layer filters

### LINKED ISSUE #464: [Feature]: Add section or kind filters to the syu app search box

The browser app search box currently hard-caps results at 20 items and tells users to refine their query, but it does not offer any section/kind filters to narrow a broad match set from inside the same UI.

That makes large repositories harder to explore because the app knows which layer each result belongs to, yet the search workflow still forces people to keep guessing narrower free-text queries.

## What change do you want?

Add section/kind-aware filtering to the browser app search. A good first scope would be:

- let users filter search results to one layer (philosophy / policy / requirement / feature)
- keep the existing keyboard search flow intact
- preserve the current capped-result messaging as a fallback for genuinely broad searches

## Specification impact

This affects the browser UI's ability to deliver the same clear, low-friction navigation story that the CLI already provides across the four layers.

## Alternatives considered

The current refine-your-query message works for small workspaces, but it does not scale well once a broad term appears across many items and layers.

## Before submitting

tohboeh5:
## Summary
- add layer filter buttons to the browser app search so broad queries can be narrowed without losing keyboard navigation
- keep the existing capped-result behavior while tailoring empty and overflow states to the active filter
- extend Playwright coverage for the new filter controls and filtered result flow

Closes #464

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -33,8 +33,11 @@ type SearchResult = {
   kind: SectionKind;
 };
 
+type SearchFilter = "all" | SectionKind;
+
 const SECTION_ORDER: SectionKind[] = ["philosophy", "policies", "requirements", "features"];
 const SEARCH_RESULTS_LIST_ID = "spec-search-results-list";
+const SEARCH_FILTER_GROUP_ID = "spec-search-filter-group";
 const REFRESH_POLL_MIN_MS = 2_000;
 const REFRESH_POLL_MAX_MS = 10_000;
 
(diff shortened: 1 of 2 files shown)


tohboeh5:
Overview review:

This adds a genuinely helpful search affordance for larger workspaces: the filter chips are easy to understand, the implementation stays small, and the browser regression test exercised the new path successfully.

The main issue I found is keyboard UX continuity. After selecting a layer filter, pressing `Escape` clears the query but leaves the filter active, so the next search is still silently scoped even though the box looks reset. I left one inline comment on that behavior because it is exactly the kind of subtle state leak that can make search feel unreliable to first-time users.

Other than that, the UI direction is strong and the capped-result guidance reads better with the new filtering story in place.

app/src/App.tsx
```diff
@@ -819,14 +833,42 @@ function App() {
...
+            >
+              {SEARCH_FILTER_OPTIONS.map((option) => {
+                const active = searchFilter === option.value;
+                return (
+                  <button
+                    key={option.value}
+                    type="button"
+                    aria-pressed={active}
+                    onClick={() => {
+                      setSearchFilter(option.value);
```
tohboeh5, line 854: This setter makes the layer filter sticky, but the existing keyboard clear path (`Escape`) still only resets the text query. After a user clicks a layer once, the next search stays silently scoped even though the box looks cleared, which weakens the current keyboard-search workflow. Please either reset `searchFilter` on `Escape` too or keep the active scope visible after clearing the query.

PR #480, opened 2026-04-19
feat: support go doc_contains traces

### LINKED ISSUE #348: [Feature]: Support doc_contains inspection for Go trace mappings
Go trace mappings can now prove explicit symbol ownership, but they still cannot validate `doc_contains` snippets the way Rust, Python, and TypeScript can.

That leaves Go adopters with an awkward middle state: they can point to symbols, but they cannot express or verify the richer documentation promises that existing trace workflows already rely on in other languages.

### What change do you want?
Add rich Go symbol inspection so `doc_contains` can be validated for traced Go functions, methods, and related declarations.

- extracting Go doc comments for traced functions and methods
- validating `doc_contains` against that extracted comment text
- keeping the current explicit-ID requirement aligned with the richer inspection result

### Specification impact
This likely extends the validation/language-adapter spec and any docs that describe which adapters support documentation assertions.

### Alternatives considered
Keeping Go limited to symbol existence is a workable baseline, but it makes the adapter meaningfully less expressive than the primary languages `syu` already supports.

### Before submitting

tohboeh5:
## Summary
- add rich Go symbol inspection so doc_contains works for Go trace mappings
- teach syu validate --fix to insert missing Go doc-comment snippets above traced declarations
- update the docs and tests to reflect Go's new trace-inspection support

Closes #348

### DIFF
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -1155,6 +1155,81 @@ mod external;
     }
 
     #[test]
+    fn go_inspection_covers_types_values_blocks_and_empty_docs() {
+        let type_source = "// REQ-TYPE\n type sample struct{}\n";
+        let type_inspected = inspect_symbol(
+            "go",
+            &SyuConfig::default(),
+            std::path::Path::new("go/sample.go"),
+            type_source,
+            "sample",
+        )
+        .expect("inspection should succeed")
+        .expect("type should exist");
+        assert!(type_inspected.docs.contains("REQ-TYPE"));
+
+        let value_source = "// REQ-VALUE\nconst sample = 1\n";
+        let value_inspected = inspect_symbol(
+            "go",
+            &SyuConfig::default(),
+            std::path::Path::new("go/sample.go"),
+            value_source,
+            "sample",
+        )
+        .expect("inspection should succeed")
+        .expect("value should exist");
+        assert!(value_inspected.docs.contains("REQ-VALUE"));
+
+        let block_source = "const (\n\tother = 1\n\t// REQ-BLOCK\n\tsample = 2\n)\n";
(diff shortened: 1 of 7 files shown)


src/inspect.rs
```diff
@@ -320,6 +333,43 @@ pub(crate) fn inspect_typescript_file(
     Ok(symbols)
 }
 
+fn inspect_go_symbol(contents: &str, symbol: &str) -> Option<GoSymbolInspection> {
+    let escaped = regex::escape(symbol);
+    let function_regex = Regex::new(&format!(r"^\s*func\s*(?:\([^)]+\)\s*)?{escaped}\b")).ok()?;
+    let type_regex = Regex::new(&format!(r"^\s*type\s+{escaped}\b")).ok()?;
```
tohboeh5, line 339: The new Go inspector adds several fresh branches (`type`, single-line `var`/`const`, grouped value blocks, and block-comment collection), but the added tests only cover a function plus one method. That matches the current CI state: the coverage gate is still failing at 99.93%. Please extend the Go inspection tests to exercise the newly introduced declaration forms, otherwise this PR cannot clear the repositorys 100% line-coverage contract.

tohboeh5:
Overview review:

The direction is good: bringing `doc_contains` to Go closes a real capability gap, the implementation hooks are in the right places, and the core inspection tests pass locally.

The blocking problem is test completeness. CI is currently red on both `precommit` and `coverage`, and the concrete reason is that the new Go inspector introduces multiple declaration-path branches without covering them all in tests. I left one inline comment on that point because the repository keeps a strict 100%% line-coverage contract.

Once the Go-specific tests cover the additional declaration forms, the documentation updates here should be much easier to trust and merge.

PR #481, opened 2026-04-19
fix: share wasm build target dirs

### LINKED ISSUE #440: [Bug]: Respect shared target directories during browser Wasm builds in temp worktrees
### What area is affected?

Repository automation or CI

### What happened?

The browser Wasm rebuild path writes its Cargo target output inside the worktree by default. In a `/tmp` worktree, that can fail with `No space left on device` even when the repository's normal shared target directory still has room.

I hit this while validating a branch from a temporary worktree: the build failed inside `app/wasm/target/...` until I overrode `CARGO_TARGET_DIR` to point at a roomier shared location.

The browser Wasm build should respect a shared target/output directory by default (or at least document and honor an override consistently) so worktree-heavy contributor flows do not fail just because the worktree lives on a smaller filesystem.

### syu version

main

### Environment

Linux, temporary git worktree under `/tmp`, Cargo-driven validation that triggers the browser Wasm rebuild.

### Steps to reproduce

1. Create a git worktree under a space-constrained path such as `/tmp`.
2. Run a Cargo workflow that triggers the embedded browser app build.
3. Observe that the Wasm build writes under `app/wasm/target` inside the worktree.
4. Watch the build fail when that filesystem runs out of space.

```shell
error: couldn't create a temp dir: No space left on device (os error 28)
... path "/tmp/.../app/wasm/target/wasm32-unknown-unknown/..."
```

### Before submitting

tohboeh5:
## Summary
- default browser Wasm builds to a shared target directory rooted at the git common dir when Cargo does not provide one
- keep honoring explicit CARGO_TARGET_DIR overrides while leaving worktree-local fallback behavior intact outside git worktrees
- avoid deleting the shared Wasm target after successful builds so repeated worktree validations can reuse it

Closes #440

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -301,7 +301,11 @@ fn default_wasm_target_dir_from_common_dir(
     configured_target_dir: Option<&Path>,
 ) -> PathBuf {
     if let Some(configured_target_dir) = configured_target_dir {
-        return configured_target_dir.to_path_buf();
+        return if configured_target_dir.is_absolute() {
+            configured_target_dir.to_path_buf()
+        } else {
+            manifest_dir.join(configured_target_dir)
+        };
     }
 
     if let Some(repo_root) = git_common_dir.and_then(Path::parent) {
@@ -344,6 +348,18 @@ mod tests {
                 Some(Path::new("/shared/target"))
             ),
             Path::new("/shared/target")
+        );
+    }
+
+    #[test]
+    fn default_wasm_target_dir_resolves_relative_configured_target_dir_from_manifest_dir() {
+        assert_eq!(
+            default_wasm_target_dir_from_common_dir(
+                Path::new("/repo/.worktrees/impl"),
+                Some(Path::new("/repo/.git")),
+                Some(Path::new("target"))
+            ),
+            Path::new("/repo/.worktrees/impl/target")
         );
     }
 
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -271,3 +284,86 @@ fn build_browser_bundle(app_dir: &Path, out_dir: &Path, required_npm: &str) -> R
...
+    default_wasm_target_dir_from_common_dir(
+        manifest_dir,
+        git_common_dir(manifest_dir).as_deref(),
+        env::var_os("CARGO_TARGET_DIR")
+            .map(PathBuf::from)
+            .as_deref(),
+    )
+}
+
+fn default_wasm_target_dir_from_common_dir(
```
tohboeh5, line 294: Forwarding `CARGO_TARGET_DIR` unchanged is only safe when it is absolute. If the outer Cargo invocation used a relative target dir such as `target`, Cargo resolves that relative to the original workspace, but the npm/wasm-pack hop now runs from `app/` and `app/wasm/`, so the same string points at a different directory. Please normalize configured target dirs to an absolute path before passing them down, otherwise this fix still regresses shared-target setups that rely on a relative `CARGO_TARGET_DIR`.

tohboeh5:
Overview review:

This heads in the right direction: sharing the Wasm target directory across worktrees is the right root cause to address, and the small cleanup change in `build-wasm.mjs` keeps the generated target directory intact when that shared path is in use.

The remaining issue is path semantics. The new forwarding logic treats any configured `CARGO_TARGET_DIR` as ready to reuse, but relative target dirs change meaning once the build hops into `app/` and then `app/wasm/`. I left one inline comment on that because it can still break exactly the shared-target setups this PR is trying to stabilize.

With that normalization tightened up, this looks like a strong fix for reused worktrees.

PR #482, opened 2026-04-19
feat: include spec coverage in reports

### LINKED ISSUE #352: [Feature]: Surface requirement and feature coverage summaries in syu report
The requirement/feature coverage summary currently exists only as CI-facing Markdown output.

That helps reviewers in Actions, but it does not help users who inspect repository status locally through `syu report` or other first-party outputs.

### What change do you want?
Expose the requirement/feature coverage summary through `syu report` as well, so local and CI reporting surfaces stay aligned.

- a Markdown section in `syu report`
- reuse of the same summary logic CI already uses
- output that stays useful even when the repository has mixed-language traces and partial line instrumentation

### Specification impact
This extends the reporting surface and likely needs updates to report-related docs/spec references.

### Alternatives considered
Leaving the summary in CI-only artifacts works for maintainers, but it makes local repository health inspection less complete than the automated review path.

### Before submitting

tohboeh5:
## Summary
- append a requirement and feature coverage section to syu report when LCOV data is available
- compute the section from the loaded workspace so local reports can summarize traced Rust test and implementation coverage directly
- refresh the checked-in generated report snapshot and add report command coverage tests

Closes #352


src/command/report.rs
```diff
@@ -85,6 +96,254 @@ fn resolve_configured_report_output(workspace_root: &Path, output: &Path) -> Res
     Ok(workspace_root.join(normalized))
 }
 
+fn render_spec_coverage_summary(workspace: &Workspace) -> Result<Option<String>> {
+    let lcov_path = workspace.root.join("target/coverage/lcov.info");
```
tohboeh5, line 112: This makes `syu report` depend on untracked local state under `target/coverage/lcov.info`, so the checked-in generated report is no longer deterministic. CI proves the problem already: `syu validate .` passes, then the quality/precommit jobs rerun the generator in a clean checkout without that LCOV file and immediately delete the new section as stale output. Please decouple checked-in report generation from ephemeral local coverage artifacts (or persist the coverage input somewhere versioned), otherwise this feature will keep fighting the repository freshness gates.

tohboeh5:
Overview review:

The spec-oriented coverage summary is a useful idea in principle: translating raw Rust coverage into requirement and feature terms would make local reports much easier to discuss in review.

Right now the implementation is blocked by determinism. The new section only appears when an ephemeral `target/coverage/lcov.info` file happens to exist, so the checked-in generated report depends on machine-local state and immediately trips the repository freshness gates in CI. I left one inline comment on that because the current behavior is not mergeable as-is.

Once the input source is made reproducible, the rendering approach here looks like a promising addition to `syu report`.

PR #483, opened 2026-04-19
fix: make docs-site validation installs repeatable

### LINKED ISSUE #439: [Bug]: Make repeated docs-site installs idempotent in reused worktrees

tohboeh5:
## Summary
- route validate-website.sh through the cleanup-aware docs-site dependency installer
- keep the repository quality assertion aligned with the new validation path
- validate the docs-site wrapper twice in the same worktree to confirm repeated runs stay deterministic

Closes #439

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -22,6 +22,7 @@ fn repository_declares_precommit_and_quality_gates() {
     let quality_script = read_file("scripts/ci/quality-gates.sh");
     let validate_app_script = read_file("scripts/ci/validate-app.sh");
     let validate_website_script = read_file("scripts/ci/validate-website.sh");
+    let install_docs_site_deps_script = read_file("scripts/ci/install-docs-site-deps.sh");
     let ci_workflow = read_file(".github/workflows/ci.yml");
     let contributing = read_file("CONTRIBUTING.md");
     let repo_config = read_file("syu.yaml");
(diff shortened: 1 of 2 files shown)


tests/repository_quality.rs
```diff
@@ -47,7 +47,7 @@ fn repository_declares_precommit_and_quality_gates() {
     assert!(validate_website_script.contains("FEAT-QUALITY-001"));
     assert!(validate_website_script.contains("validate_website"));
     assert!(validate_website_script.contains("scripts/ci/quality-gates.sh"));
-    assert!(validate_website_script.contains("npm --prefix website ci"));
+    assert!(validate_website_script.contains("install-docs-site-deps.sh"));
```
tohboeh5, line 51: Now that the docs-site install behavior moved into `scripts/ci/install-docs-site-deps.sh`, this assertion only proves the wrapper script mentions the helper name. It no longer guards the actual contract that mattered here (`website/node_modules` gets rebuilt from a clean slate and `npm --prefix website ci` still runs). Please add repository-quality coverage for the helper script contents too, otherwise a future regression inside that helper can slip through while this test still passes.

tohboeh5:
Overview review:

The functional fix looks right: switching `validate-website` to the shared install helper matches the repo guidance better, and the website validation flow completed successfully in local execution.

The main thing still missing is protection against future regressions. The updated repository-quality test now checks only that the wrapper calls `install-docs-site-deps.sh`, not that the helper still performs the clean reinstall behavior this PR is relying on. I left one inline comment on that gap.

With that extra coverage in place, this would be a tidy and focused fix for the repeated-install problem.

PR #484, opened 2026-04-19
fix: stop auto-installing browser app deps

tohboeh5:
## Summary
- stop build.rs from running npm ci automatically when app dependencies are missing or stale
- surface explicit setup commands that point contributors to pinned npm installation from the repo root
- lock the behavior in repository quality coverage and document it in CONTRIBUTING

## Validation
- cargo fmt --all --check
- cargo test --test repository_quality repository_ships_browser_app -- --exact
- cargo run -- validate .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,19 @@ jobs:
         with:
           components: rustfmt,clippy
 
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: "25"
+          cache: npm
+          cache-dependency-path: app/package-lock.json
+
+      - name: Install browser app dependencies
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


tohboeh5:
Overview review:\n\nThis is a well-scoped and worthwhile change. Moving the browser app dependency install path from an implicit build-time side effect to an explicit setup step makes Rust-only and docs-only workflows much more predictable, and the updated contributor guidance matches the new behavior clearly.\n\nWhat I liked most is that the PR does not just remove the auto-install logic: it replaces it with a direct, actionable failure path and keeps the repository-quality coverage aligned with that contract. I also checked the surrounding workflow shape, and the new error message points at the existing pinned npm helper in a way that fits the current CI and contributor setup.\n\nI did not find any blocking issues in the changed code. The behavior, UX of the failure message, and the documentation update all look consistent with the repository's current direction.

Issue #485, opened 2026-04-19
[Bug]: Cover every checked-in example syu.yaml with repository-quality freshness checks


tohboeh5:
### What area is affected?

Documentation or examples

### What happened?

The repository now ships multiple checked-in example workspaces, but repository-quality coverage still only guards some of the freshness contracts around example inputs. That leaves room for new example `syu.yaml` files to drift away from the current CLI version without the same protection.

### What did you expect to happen?

Every checked-in example workspace should get the same freshness protection for its versioned `syu.yaml` inputs so contributors do not accidentally ship stale example metadata.

### syu version

main

### Environment

Repository checkout, example workspaces, and `tests/repository_quality.rs`.

### Steps to reproduce

1. Inspect the checked-in examples under `examples/`.
2. Compare them with the repository-quality assertions that protect example freshness.
3. Notice that coverage is not expressed as a broad contract for every checked-in example `syu.yaml`.

### Logs, screenshots, or failing output

Relevant repository surfaces include `examples/**/syu.yaml` and the example freshness checks in `tests/repository_quality.rs`.

### Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

PR #495, opened 2026-04-19
test: guard all example workspace versions

tohboeh5:
## Summary\n- extend repository-quality example freshness assertions to every checked-in example syu.yaml\n- keep the existing example validation contract explicit for team-scale as well\n\n## Testing\n- bash scripts/ci/quality-gates.sh\n- cargo run -- validate .\n\nCloses #485

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -702,34 +702,42 @@ fn repository_ships_example_workspaces() {
     let rust_example_requirement =
         read_file("examples/rust-only/docs/syu/requirements/core/rust.yaml");
     let rust_example_config = read_file("examples/rust-only/syu.yaml");
+    let python_example_config = read_file("examples/python-only/syu.yaml");
     let python_example_requirement =
         read_file("examples/python-only/docs/syu/requirements/core/python.yaml");
     let csharp_fallback_requirement =
         read_file("examples/csharp-fallback/docs/syu/requirements/core/csharp.yaml");
     let csharp_fallback_config = read_file("examples/csharp-fallback/syu.yaml");
     let csharp_fallback_readme = read_file("examples/csharp-fallback/README.md");
+    let docs_first_config = read_file("examples/docs-first/syu.yaml");
     let docs_first_requirement =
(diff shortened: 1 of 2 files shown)


tests/repository_quality.rs
```diff
@@ -691,30 +691,41 @@ fn repository_ships_example_workspaces() {
...
         read_file("examples/python-only/docs/syu/requirements/core/python.yaml");
+    let docs_first_config = read_file("examples/docs-first/syu.yaml");
     let docs_first_requirement =
         read_file("examples/docs-first/docs/syu/requirements/core/docs.yaml");
+    let go_example_config = read_file("examples/go-only/syu.yaml");
     let go_example_requirement = read_file("examples/go-only/docs/syu/requirements/core/go.yaml");
     let go_example_readme = read_file("examples/go-only/README.md");
+    let polyglot_config = read_file("examples/polyglot/syu.yaml");
     let polyglot_feature = read_file("examples/polyglot/docs/syu/features/languages/polyglot.yaml");
+    let team_scale_config = read_file("examples/team-scale/syu.yaml");
```
tohboeh5, line 723: This list is still incomplete against current `main`: `examples/csharp-fallback` is also a checked-in workspace now, and `tests/example_workspaces.rs` already exposes `csharp_fallback_example_validates`. Rebasing this PR as-is would keep the new team-scale guard but silently drop freshness coverage for the newest example. Please carry the `csharp-fallback` `syu.yaml` assertion and the matching `example_tests.contains("csharp_fallback_example_validates")` check alongside these additions.

tohboeh5:
Overview review:

The intent here is good: tightening the repository-quality guard so example workspaces cannot drift away from the published CLI version is exactly the kind of regression this test should catch, and adding the team-scale validation contract is a useful extension.

The problem is completeness against current `main`. This branch only carries a subset of the now-expected example freshness assertions, so rebasing/merging it in its current shape would still leave `csharp-fallback` outside the version guard even though that example is already checked in and already has an end-to-end validation test. I left one inline comment on the exact spot where that omission shows up.

Once the branch is rebased and keeps the `csharp-fallback` assertions alongside the new team-scale coverage, this should be in much better shape.

PR #498, opened 2026-04-19
test: lock docs site helper contract

### LINKED ISSUE #486: [Bug]: Validate the docs-site install helper contract, not just the wrapper call site
### What area is affected?

Repository automation or CI

### What happened?

The docs-site install flow now routes through `scripts/ci/install-docs-site-deps.sh`, but the repository-quality contract is still easier to satisfy by only asserting that the wrapper mentions the helper. That leaves the actual clean-install behavior less protected than the call site.

Repository-quality coverage should lock in the helper script's real contract: clean reinstall behavior, the expected `npm --prefix website ci` step, and the maintainer-facing reason this helper exists.

### syu version

main

### Environment

Repository checkout, `scripts/ci/validate-website.sh`, `scripts/ci/install-docs-site-deps.sh`, and `tests/repository_quality.rs`.

### Steps to reproduce

1. Inspect `scripts/ci/validate-website.sh`.
2. Inspect `scripts/ci/install-docs-site-deps.sh`.
3. Compare both with the current repository-quality assertions.
4. Notice that the helper behavior itself is not protected as strongly as the wrapper reference.

Relevant files: `scripts/ci/validate-website.sh`, `scripts/ci/install-docs-site-deps.sh`, `tests/repository_quality.rs`.

### Before submitting

tohboeh5:
## Summary
- strengthen repository-quality assertions around scripts/ci/install-docs-site-deps.sh itself
- lock in the helper's clean reinstall loop, repo-root setup, and website npm ci contract

## Testing
- npm --prefix app ci
- cargo run -- validate .

Closes #486

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -50,9 +50,16 @@ fn repository_declares_precommit_and_quality_gates() {
     assert!(validate_website_script.contains("scripts/ci/quality-gates.sh"));
     assert!(validate_website_script.contains("install-docs-site-deps.sh"));
     assert!(validate_website_script.contains("npm --prefix website run build"));
+    assert!(install_docs_site_deps_script.contains("set -euo pipefail"));
+    assert!(install_docs_site_deps_script.contains("repo_root="));
+    assert!(install_docs_site_deps_script.contains("cd \"$repo_root\""));
     assert!(install_docs_site_deps_script.contains("Branch switches can leave behind"));
     assert!(install_docs_site_deps_script.contains("website/node_modules"));
+    assert!(install_docs_site_deps_script.contains("for attempt in range(5):"));
+    assert!(install_docs_site_deps_script.contains("if not node_modules.exists():"));
     assert!(install_docs_site_deps_script.contains("shutil.rmtree"));
+    assert!(install_docs_site_deps_script.contains("if attempt == 4:"));
+    assert!(install_docs_site_deps_script.contains("time.sleep(1)"));
     assert!(install_docs_site_deps_script.contains("npm --prefix website ci"));
     assert!(install_precommit.contains("site --user-base"));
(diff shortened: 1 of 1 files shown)


tests/repository_quality.rs
```diff
@@ -50,9 +50,16 @@ fn repository_declares_precommit_and_quality_gates() {
     assert!(validate_website_script.contains("scripts/ci/quality-gates.sh"));
     assert!(validate_website_script.contains("install-docs-site-deps.sh"));
     assert!(validate_website_script.contains("npm --prefix website run build"));
+    assert!(install_docs_site_deps_script.contains("set -euo pipefail"));
+    assert!(install_docs_site_deps_script.contains("repo_root="));
+    assert!(install_docs_site_deps_script.contains("cd \"$repo_root\""));
     assert!(install_docs_site_deps_script.contains("Branch switches can leave behind"));
     assert!(install_docs_site_deps_script.contains("website/node_modules"));
+    assert!(install_docs_site_deps_script.contains("for attempt in range(5):"));
```
tohboeh5: This assertion block is locking the current implementation shape of `install-docs-site-deps.sh`, not its externally visible contract. Renaming `repo_root`, changing the retry loop from Python to shell, or extracting the backoff constant would all fail repository-wide CI even if the helper still cleaned stale `website/node_modules` and ran `npm --prefix website ci` correctly. Could we keep the test focused on the observable guarantees we actually want to preserve?

tohboeh5:
Overview review:\n\nI like the intent here: the repository-quality suite should protect the docs-site helper because that script is part of the contributor workflow contract.\n\nMy concern is that the new assertions pin several internal implementation details instead of the contract itself. I left one inline comment on that block because it would make harmless refactors fail CI without changing the observable helper behavior.

PR #502, opened 2026-04-19
fix: skip ownership checks for ignored generated paths

### LINKED ISSUE #278: [Bug]: Exclude generated build directories from symbol ownership and trace coverage checks

tohboeh5:
- Closes #278

## Summary
- skip inline and sidecar ownership enforcement for paths already excluded from generated-path coverage
- keep the ignore list configurable so repositories can opt generated paths back in
- add regression tests for ignored and opt-in ownership behavior

## Validation
- scripts/ci/quality-gates.sh
- cargo run -- validate .

### DIFF
--- a/src/command/check.rs
+++ b/src/command/check.rs
@@ -15,7 +15,10 @@ use crate::{
     cli::{CheckArgs, OutputFormat, ValidationGenreFilter, ValidationSeverityFilter},
     command::shell_quote_path,
     config::{SyuConfig, TraceOwnershipMode},
-    coverage::{normalize_relative_path, validate_symbol_trace_coverage},
+    coverage::{
+        normalize_relative_path, normalized_symbol_trace_coverage_ignored_paths,
+        path_matches_ignored_generated_directory, validate_symbol_trace_coverage,
+    },
     inspect::{apply_symbol_doc_fix, inspect_symbol, supports_rich_inspection},
     language::adapter_for_language,
     model::{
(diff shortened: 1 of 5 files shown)


src/command/check.rs
```diff
@@ -2374,74 +2377,77 @@ fn verify_trace_reference(
...
-        OwnershipDeclaration::Satisfied => {}
-        OwnershipDeclaration::Missing { manifest_path } => {
-            let (message, suggestion) = match manifest_path {
-                Some(manifest_path) => {
-                    let manifest_display = manifest_path
-                        .strip_prefix(root)
-                        .unwrap_or(manifest_path.as_path());
-                    (
+    let ignored_paths = normalized_symbol_trace_coverage_ignored_paths(config);
+    if !path_matches_ignored_generated_directory(display_path, &ignored_paths) {
```
tohboeh5, line 2381: This condition changes `validate.symbol_trace_coverage_ignored_paths` from a coverage-discovery filter into an ownership-enforcement bypass as well, but I couldnt find any matching config/help/docs update in this PR. That means users can now suppress `SYU-trace-id-001` for ignored paths without any user-facing explanation of that trade-off. Please document the new ownership interaction anywhere we explain this config knob, otherwise reviewers and adopters will keep reading it as "coverage only" and miss the weaker ownership guarantee.

tohboeh5:
Overview review:

This is a worthwhile fix: the new helper and tests make the generated-path behavior much more consistent, and I confirmed the added ignored-path tests pass on the branch.

My main concern is the scope of the behavior change. `validate.symbol_trace_coverage_ignored_paths` now affects ownership enforcement as well as generated-path coverage, but the PR does not update any user-facing docs/help/config reference to explain that stronger opt-out. I left one inline comment on that point because the implementation is otherwise in good shape, but the changed contract is too implicit right now.

PR #503, opened 2026-04-19
fix: stop Cargo from auto-installing browser app dependencies

### LINKED ISSUE #399: [Bug]: Stop cargo builds from auto-installing browser app dependencies
## What area is affected?
CLI or runtime behavior

## What happened?
`build.rs` currently makes ordinary Rust builds responsible for the embedded browser app: it checks `app/node_modules/.package-lock.json`, runs `npm ci` when the marker is missing or stale, rebuilds the browser Wasm bridge, and then builds the browser bundle.

In practice that means a contributor touching Rust-only or docs-only code can hit Node dependency installation and browser build failures just by running Cargo commands. It also makes worktree-heavy flows and low-disk environments much more fragile than the project's “simple, low-friction workflow” messaging suggests.

Ordinary Cargo builds should not have to install browser dependencies or rebuild the browser bundle unless the contributor is explicitly working on the browser surface or packaging a release artifact that truly needs the embedded app.

## syu version
main

## Environment
Observed on Linux while working across multiple git worktrees and running the normal local validation commands documented in `CONTRIBUTING.md`.

## Steps to reproduce
1. Start from a worktree without a ready `app/node_modules/.package-lock.json` install marker.
2. Run a normal Cargo-driven workflow such as `cargo test` or `cargo run -- validate .`.
3. Observe that `build.rs` attempts `npm ci`, Wasm regeneration, and a browser bundle build before the Rust command can finish.

Relevant code path: `build.rs` -> `ensure_app_dependencies()` -> `rebuild_browser_wasm_bindings()` -> `build_browser_bundle()`.

tohboeh5:
- Closes #399

## Summary
- stop `build.rs` from running `npm ci` implicitly during ordinary Cargo-driven builds
- teach contributor docs and CI jobs to provision app dependencies explicitly before embedded-browser build steps
- lock the new contract into repository-quality assertions

## Validation
- cargo run -- validate .


tohboeh5:
Blocking review note:

GitHub currently shows this PR with **0 changed files** against `main`, and a local `git diff origin/main...fix/issue-399-no-auto-install` is also empty. That means there is no remaining review surface to inspect inline and, as opened today, this branch does not actually introduce a mergeable delta anymore.

I did verify the intended behavior separately in a detached worktree: with `app/node_modules` removed, `cargo run -- validate .` now fails fast with the explicit dependency-preparation message instead of silently running `npm ci`, and after `scripts/ci/pinned-npm.sh install app && npm --prefix app ci` the same command succeeds. So the direction of the fix is good, but this PR needs to be refreshed so GitHub shows the actual delta again, or closed if an equivalent change has already landed another way.

tohboeh5:
Overview review:

The underlying fix is a good one: failing fast when embedded browser dependencies are missing is much closer to the project's low-friction philosophy than letting ordinary Cargo workflows mutate `app/node_modules` behind the contributor's back. I also like that the change was paired with docs/CI/repository-quality updates so the new contract is explicit instead of living only in `build.rs`.

The bad news is that this PR is currently not reviewable on GitHub because it shows **no diff against `main`**. That is a real process problem, not just a UI quirk: reviewers cannot leave code-anchored comments, inspect the exact lines that would land, or confirm that the docs/CI/test updates still match the implementation. As it stands, I would not merge this PR until the branch is refreshed to expose the intended delta again, or the PR is closed as already-equivalent work.

PR #504, opened 2026-04-19
test: cover every example syu.yaml freshness

### LINKED ISSUE #485: [Bug]: Cover every checked-in example syu.yaml with repository-quality freshness checks

tohboeh5:
- Closes #485

## Summary
- replace per-example version assertions with a repository-wide scan of checked-in examples/**/syu.yaml files
- keep the existing example-specific content assertions alongside the broader freshness contract
- ensure future example workspaces automatically inherit current-version coverage

## Validation
- npm --prefix app ci
- cargo test --test repository_quality repository_ships_example_workspaces
- cargo run -- validate .

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashSet, fs, path::PathBuf};
+use std::{
+    collections::HashSet,
+    fs,
+    path::{Path, PathBuf},
+};
 
 use serde_json::Value;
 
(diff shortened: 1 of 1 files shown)


tests/repository_quality.rs
```diff
@@ -14,6 +18,26 @@ fn read_json(path: &str) -> Value {
     serde_json::from_str(&read_file(path)).expect("repository JSON should parse")
 }
 
+fn collect_named_files(root: &Path, file_name: &str, output: &mut Vec<PathBuf>) {
+    for entry in fs::read_dir(root).expect("directory should be readable") {
+        let entry = entry.expect("directory entry should exist");
+        let path = entry.path();
+
+        if path.is_dir() {
+            collect_named_files(&path, file_name, output);
```
tohboeh5: This broadens the contract from "top-level checked-in example workspaces" to "any `syu.yaml` anywhere under `examples/`". If we later add nested fixtures, support workspaces, or intentionally stale negative examples below an existing example root, this test will start failing even though those files are not first-class shipped examples. I think the scan should stay scoped to the example workspace roots the product actually documents.

tohboeh5:
Overview review:

I like the direction here: replacing the per-example version assertions with a broader repository-quality contract is exactly the kind of maintenance guard this repository needs, and the test still keeps the example-specific content checks that explain why each workspace exists.

My main concern is scope. The new recursive walk currently treats every nested `examples/**/syu.yaml` file as a shipped example contract, which is stricter than the issue statement and could make future nested fixtures or support workspaces awkward to add. I left one inline comment on that point. If you tighten the scan to the documented example workspace roots, I think this will be a solid improvement.

PR #505, opened 2026-04-19
test: lock go-only init help guidance

### LINKED ISSUE #347: [Feature]: Add a go-only starter template to syu init

tohboeh5:
- Closes #347

## Summary
- add an integration test that locks `syu init --help` to list `go-only` as a supported starter template
- assert the help output includes both the `possible values` list and a concrete `--template go-only` example
- keep the existing scaffold-and-validate coverage while making the CLI discovery path explicit

## Validation
- npm --prefix app ci
- cargo test --test init_command init_help_lists_go_only_template_as_a_supported_starter
- cargo run -- validate .

### DIFF
--- a/tests/init_command.rs
+++ b/tests/init_command.rs
@@ -94,6 +94,31 @@ fn init_command_interactive_requires_a_terminal() {
 
 #[test]
 // REQ-CORE-009
+fn init_help_lists_go_only_template_as_a_supported_starter() {
+    let output = Command::cargo_bin("syu")
+        .expect("binary should build")
+        .arg("init")
+        .arg("--help")
+        .output()
+        .expect("help should run");
+
+    assert!(
+        output.status.success(),
+        "stdout:\n{}\nstderr:\n{}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    );
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("go-only"));
+    assert!(stdout.contains(
+        "possible values: generic, docs-first, rust-only, python-only, go-only, polyglot"
+    ));
+    assert!(stdout.contains("syu init . --template go-only"));
+}
+
+#[test]
+// REQ-CORE-009
 fn init_command_bootstraps_language_templates_that_validate_accept() {
     for (template, requirement_path, feature_path, requirement_id, feature_id) in [
         (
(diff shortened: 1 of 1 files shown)


tests/init_command.rs
```diff
@@ -92,6 +92,31 @@ fn init_command_interactive_requires_a_terminal() {
...
+        output.status.success(),
+        "stdout:\n{}\nstderr:\n{}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    );
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("go-only"));
+    assert!(stdout.contains(
+        "possible values: generic, docs-first, rust-only, python-only, go-only, polyglot"
```
tohboeh5: This assertion locks the entire Clap-generated `possible values` string, not just the `go-only` discovery contract. That means an unrelated template addition, value reordering, or help wrapping change will fail the test even if `go-only` is still clearly documented. I think the test should stay focused on the new user-facing guarantee: `go-only` is mentioned for `--template` and there is a concrete `syu init . --template go-only` example.

tohboeh5:
Overview review:

The good part here is that it closes a real discoverability gap: people should be able to find the new starter from `syu init --help`, and I verified the current help output does include both the template name and the example command.

The weak point is that the new test currently freezes more of Claps generated formatting than the feature really needs. I left one inline comment where the assertion couples this test to the full rendered `possible values` list instead of the narrower `go-only` help contract. Once that is tightened, the added coverage will be much easier to keep stable as the template matrix grows.

PR #506, opened 2026-04-19
test: lock coverage summary metadata loading

### LINKED ISSUE #351: [Feature]: Batch workspace lookups when generating spec-aware CI coverage reports
The spec-aware coverage summary shells out through `cargo run -- list/show ...` repeatedly to load requirement and feature metadata.

That keeps the first implementation small, but it also means the report cost grows with repository size and can recompile or relaunch the CLI more than necessary in CI.

### What change do you want?
Optimize spec-aware coverage summary generation so it loads workspace metadata once instead of repeatedly spawning the CLI.

- a single-purpose Rust helper or subcommand that returns all needed coverage/spec metadata in one pass
- keeping the Markdown output stable for existing CI consumers
- measuring the improvement on the repository's own coverage job

### Specification impact
This is mostly repository quality/performance work, but it may add a small internal CLI/reporting surface.

### Alternatives considered
The current shell-out approach is acceptable for the small repository today, but it is a weak default if the report is meant to scale to larger self-hosting repositories.

### Before submitting

tohboeh5:
- Closes #351

## Summary
- add a repository-quality contract that locks spec-aware coverage summary generation to the current single metadata-load path
- assert the helper uses `run_syu_json(...)` exactly once beyond its definition and does not shell out through repeated `syu show` calls
- keep the existing coverage-summary script contract while making the performance-sensitive behavior explicit

## Validation
- npm --prefix app ci
- cargo test --test repository_quality repository_declares_coverage_gate_at_one_hundred_percent
- cargo run -- validate .

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -130,10 +130,11 @@ fn repository_declares_coverage_gate_at_one_hundred_percent() {
     assert!(spec_summary_script.contains("list\", \"--with-path\", \"--format\", \"json\""));
     assert!(spec_summary_script.contains("yaml.safe_load"));
     assert!(spec_summary_script.contains("Rust implementation coverage"));
-    assert_eq!(
-        spec_summary_script.matches("run_syu_json(").count(),
-        2,
-        "expected one helper definition and one metadata-loading call"
+    assert!(
+        spec_summary_script.contains(
+            "items = run_syu_json(repo_root, \"list\", \"--with-path\", \"--format\", \"json\")"
+        ),
+        "expected the coverage summary to load workspace metadata with one `syu list --with-path --format json` call"
     );
     assert!(
         !spec_summary_script.contains("\"show\","),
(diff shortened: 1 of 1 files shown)


tests/repository_quality.rs
```diff
@@ -130,6 +130,15 @@ fn repository_declares_coverage_gate_at_one_hundred_percent() {
     assert!(spec_summary_script.contains("list\", \"--with-path\", \"--format\", \"json\""));
     assert!(spec_summary_script.contains("yaml.safe_load"));
     assert!(spec_summary_script.contains("Rust implementation coverage"));
+    assert_eq!(
+        spec_summary_script.matches("run_syu_json(").count(),
```
tohboeh5: I agree with protecting the single-metadata-load behavior, but counting raw `run_syu_json(` occurrences hard-codes the current script shape more tightly than the performance contract itself. A future refactor could add another JSON helper call for unrelated data and still preserve the "load workspace metadata once, do not loop through repeated `syu show` calls" guarantee, yet this test would fail anyway. I think the contract should target the specific metadata-loading path rather than the total helper call count.

tohboeh5:
Overview review:

I like the motivation here. Locking the coverage-summary performance contract in repository-quality tests makes sense, and I confirmed the current script really does take the efficient path today: one `run_syu_json(...)` metadata load and no repeated `syu show` loop.

The concern is that the new assertion is a little too implementation-shaped. Right now it counts the helper name across the whole script, which means harmless internal refactors can fail the test even if the user-visible and performance-sensitive contract is still intact. I left one inline comment on that specific assertion. If you reframe the test around the single metadata-loading path rather than the total helper call count, this looks like a worthwhile guardrail.

PR #509, opened 2026-04-19
feat: add java-only starter template

### LINKED ISSUE #494: [Feature]: Add a Java-first starter template and example workspace

The current starter/template matrix covers generic, Rust, Python, Go, and polyglot repos. That leaves teams in other common ecosystems without a first-class starting point, even when they only need the core four-layer workflow and basic trace examples.

### What change do you want?

Add a Java-first starter template plus a matching checked-in example workspace. A good first scope would include:
- Java-oriented IDs and file names
- a minimal source and test pair
- docs that explain what works today even if deeper language-specific trace inspection remains more limited than Rust or Go

### Specification impact

This expands the language-onboarding surface without changing the core model.

### Alternatives considered

Teams can start from `generic` or `polyglot`, but that does not feel as inviting as a language-shaped entry point for a widely used ecosystem.

### Before submitting

tohboeh5:
## Summary
- add a `java-only` starter template plus matching checked-in example workspace
- document the Java-first onboarding path across README and guide pages
- cover the new template and example in init/help/templates/repository-quality tests

Closes #494


tests/init_command.rs
```diff
@@ -201,6 +208,27 @@ fn init_command_bootstraps_language_templates_that_validate_accept() {
             assert!(requirement.contains("TestGoRequirement"));
             assert!(feature.contains("status: implemented"));
             assert!(feature.contains("GoFeatureImpl"));
+        } else if template == "java-only" {
+            assert!(workspace.join("pom.xml").exists(), "missing pom.xml");
```
tohboeh5, line 235: This only proves that `syu init --template java-only` writes the expected files and strings; it never exercises the generated Maven project itself. Because the starter now promises a real `pom.xml` plus Java source and test files, a broken package name, plugin config, or test wiring would still pass CI. I think the new template needs at least one smoke check that actually runs the scaffolded Java project.

tohboeh5:
Overview review:\n\nThis is a strong addition to the starter matrix overall. The Java-first scaffold fills a real onboarding gap, the docs are wired through the usual entry points, and the init/templates tests cover the basic repository integration points well.\n\nThe weak spot is behavioral validation: the new checks prove that the files exist and that `syu validate` accepts the spec graph, but they never exercise the generated Maven project itself. That leaves room for a broken Java scaffold to ship while the repository tests still pass. I left one inline comment on that gap.

PR #510, opened 2026-04-19
test: cover every checked-in example config

### LINKED ISSUE #485: [Bug]: Cover every checked-in example syu.yaml with repository-quality freshness checks

tohboeh5:
## Summary
- enumerate top-level checked-in examples instead of hard-coding version checks one by one
- require every example `syu.yaml` to carry the current CLI version
- require every example workspace to have a dedicated validation smoke test

Closes #485


tests/repository_quality.rs
```diff
@@ -720,55 +742,60 @@ fn repository_declares_devcontainer_configuration() {
...
+            "go-only",
+            "polyglot",
+            "python-only",
+            "rust-only",
+            "team-scale",
+        ]
+    );
+    for example_name in &example_workspace_names {
+        let config = read_file(&format!("examples/{example_name}/syu.yaml"));
+        let validate_test_name = format!("{}_example_validates", example_name.replace('-', "_"));
```
tohboeh5: This assertion couples the repository-quality contract to the current Rust test function naming convention rather than the user-facing guarantee from the issue. A future refactor to a shared helper, table-driven test, or renamed smoke test would fail here even if every checked-in example still validates. I think the contract should point at the actual example-validation coverage path, not a string match on `*_example_validates` names.

tohboeh5:
Overview review:\n\nI like the direction here: replacing the repeated per-example assertions with a single inventory-style check is cleaner, and it moves the repository-quality contract closer to the actual user concern of every checked-in example staying current and covered.\n\nMy concern is that the new test still reaches for a fairly implementation-shaped signal when it checks for `*_example_validates` names in the test source text. That makes the contract brittle against harmless test refactors even if example coverage still exists. I left one inline comment on that point.

PR #511, opened 2026-04-19
docs: clarify opt-in browser app provisioning

### LINKED ISSUE #400: [Feature]: Add an explicit offline or opt-in path for embedded app builds
The current embedded-app build path assumes Cargo may trigger `npm ci` itself. That is convenient when everything is already provisioned, but it is awkward for offline, hermetic, or security-sensitive environments because an ordinary Rust build can implicitly become a networked JavaScript dependency install.

That makes the supply-chain story less explicit than the rest of the project's install and verification guidance.

## What change do you want?
Add an explicit offline/prebuilt path for the embedded browser app bundle, or otherwise require opt-in before Cargo triggers package-manager installs during normal builds.

Examples:
- use a checked-in/pre-generated browser artifact for contributor flows
- gate the `npm ci` step behind an env var or dedicated build mode
- fail with a clear message that tells contributors how to provision the browser app intentionally instead of doing it automatically

## Specification impact
This would touch contributor workflow, app build behavior, and likely the repository-automation story.

## Alternatives considered
Keeping the current automatic install is simple in the happy path, but it hides a networked side effect inside ordinary Cargo usage.

tohboeh5:
## Summary
- document the explicit browser-app dependency provisioning flow before Cargo builds
- explain that normal Cargo builds no longer run `npm ci` implicitly
- lock the README guidance in repository-quality assertions

Closes #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -712,7 +712,9 @@ binaries without checking `app/dist/` into `main`.
 
 Use the checked-in Node 25 version from `app/package.json` and `app/.nvmrc`
 when working in `app/`; that is the local contributor expectation, not only a
-CI matrix detail.
+CI matrix detail. Before you run the install commands below, switch your shell
+to that Node major with a version manager such as `nvm`, `fnm`, or `Volta`
+using the checked-in `app/.nvmrc`.
 Before you run Cargo commands that embed the browser app, provision the app
 dependencies intentionally from the repository root:
 
--- a/README.md
+++ b/README.md
@@ -713,17 +713,29 @@ binaries without checking `app/dist/` into `main`.
 Use the checked-in Node 25 version from `app/package.json` and `app/.nvmrc`
 when working in `app/`; that is the local contributor expectation, not only a
 CI matrix detail.
+Before you run Cargo commands that embed the browser app, provision the app
+dependencies intentionally from the repository root:
+
+```bash
+scripts/ci/pinned-npm.sh install app
+npm --prefix app ci
+```
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -713,17 +713,29 @@ binaries without checking `app/dist/` into `main`.
 Use the checked-in Node 25 version from `app/package.json` and `app/.nvmrc`
 when working in `app/`; that is the local contributor expectation, not only a
 CI matrix detail.
+Before you run Cargo commands that embed the browser app, provision the app
+dependencies intentionally from the repository root:
+
+```bash
+scripts/ci/pinned-npm.sh install app
```
tohboeh5, line 722: The new command block still is not self-contained for someone following it in a fresh shell: it tells readers to install the pinned npm release, but it never tells them to switch to the required Node 25 runtime before `npm ci`. On Node 20 or 24 this can still succeed with engine warnings and leave contributors on an unsupported app toolchain. I think the block should either include the version-manager step inline or route people through the `validate-app.sh` flow that already carries the expected environment.

tohboeh5:
Overview review:\n\nThe clarification around opt-in browser-app provisioning is useful, especially the explanation that Cargo no longer hides a networked `npm ci` inside ordinary Rust builds. That lines up well with the repos hermetic and offline story.\n\nThe remaining gap is that the new command blocks still are not self-contained for a contributor starting from a fresh shell: they tell readers which npm release to use, but not how to get onto the required Node 25 runtime before running the commands. I left one inline comment on that workflow mismatch.

PR #512, opened 2026-04-19
feat: add VS Code inline trace actions

### LINKED ISSUE #490: [Feature]: Add CodeLens or hover actions for spec IDs and traced files in the VS Code extension

The VS Code extension already exposes diagnostics, a context tree, and command-palette actions, but common navigation still depends on manual commands. When you are reading a traced file or a YAML spec item, the editor does not surface inline actions such as "open linked spec", "show traced files", or "trace this active symbol" where the cursor already is.

### What change do you want?

Add inline VS Code affordances such as CodeLens and/or hover actions for:
- spec IDs inside YAML files
- traced source files and symbols
- jumping into the existing trace/context commands without opening the command palette first

### Specification impact

This is an editor UX improvement for the existing traceability model.

### Alternatives considered

The current command-palette flow works, but it is slower than in-place affordances when someone is reviewing or editing a file already open in the editor.

### Before submitting

tohboeh5:
## Summary
- add YAML CodeLens actions for spec IDs plus traced files and symbols in the VS Code extension
- reuse the existing open/related/trace commands with direct arguments instead of requiring the command palette
- cover the inline target parser in extension tests and lock the new UX in repository-quality checks

Closes #490

### DIFF
--- a/editors/vscode/src/model.js
+++ b/editors/vscode/src/model.js
@@ -510,23 +510,44 @@ function collectInlineNavigationTargets(documentText) {
       continue
     }
 
-    const symbolMatch = /^(\s*)-\s*["']?([A-Za-z_][A-Za-z0-9_]*)["']?\s*$/u.exec(text)
-    if (!symbolMatch) {
+    const symbolValue = parseTraceSymbolEntry(text)
+    if (!symbolValue) {
       continue
     }
 
-    const startCharacter = text.indexOf(symbolMatch[2])
+    const startCharacter = text.indexOf(symbolValue)
     targets.push({
       kind: 'traceSymbol',
       file: activeTraceFile,
-      symbol: symbolMatch[2],
+      symbol: symbolValue,
       line,
       startCharacter,
-      endCharacter: startCharacter + symbolMatch[2].length
+      endCharacter: startCharacter + symbolValue.length
     })
   }
 
   return targets
+}
+
+function parseTraceSymbolEntry(text) {
+  const match = /^\s*-\s*(.+?)\s*$/u.exec(text)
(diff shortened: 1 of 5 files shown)


editors/vscode/src/model.js
```diff
@@ -435,6 +435,100 @@ function specIdFromText(value) {
...
+      continue
+    }
+
+    if (indent <= symbolsIndent) {
+      inSymbols = false
+      symbolsIndent = -1
+      continue
+    }
+
+    const symbolMatch = /^(\s*)-\s*["']?([A-Za-z_][A-Za-z0-9_]*)["']?\s*$/u.exec(text)
```
tohboeh5: The new CodeLens parser only recognizes simple identifier symbols here, but this repository already uses wildcard trace symbols like `*` in checked-in spec YAML. Those entries will silently lose inline symbol actions even though the README now advertises CodeLens coverage for traced symbols and files. I think this needs to handle `*` (and ideally the broader symbol spellings syu accepts) or deliberately fall back to a file-level lens for those cases.

tohboeh5:
Overview review:\n\nThis is a compelling UX improvement. Inline navigation is exactly the kind of affordance that makes the VS Code extension feel more like a first-class reviewer surface instead of just a diagnostics wrapper, and the command reuse in `extension.js` keeps the feature grounded in the existing flows.\n\nThe main issue is parser coverage. The new CodeLens target extraction only recognizes simple identifier symbols, while checked-in spec YAML in this repo already uses wildcard trace entries such as `*`. Those traces will silently miss inline symbol actions, so the advertised coverage is wider than the implementation today. I left one inline comment there.

PR #513, opened 2026-04-19
test: lock report coverage batching contract

### LINKED ISSUE #351: [Feature]: Batch workspace lookups when generating spec-aware CI coverage reports

tohboeh5:
## Summary
- add a repository-quality contract that keeps report coverage generation on the in-process workspace path
- trace the new repository-quality test back to `REQ-CORE-004`
- commit the generated docs refreshed by the updated self-spec

Closes #351

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -1060,10 +1060,10 @@ fn repository_generates_spec_coverage_without_relaunching_the_cli() {
     assert!(report_command.contains("load_workspace(&args.workspace)"));
     assert!(report_command.contains("collect_check_result_from_workspace(&workspace)"));
     assert!(report_command.contains("render_spec_coverage_summary(&workspace"));
+    assert!(report_command.contains("if let Some(spec_coverage_summary) = spec_coverage_summary"));
     assert!(!report_command.contains("cargo run -- list"));
     assert!(!report_command.contains("cargo run -- show"));
     assert!(!report_command.contains("run_syu_json"));
-    assert!(!report_command.contains("Command::new"));
 }
 
 #[test]
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -1052,6 +1052,21 @@ fn repository_ships_browser_app() {
 }
 
 #[test]
+// REQ-CORE-004
+fn repository_generates_spec_coverage_without_relaunching_the_cli() {
+    let report_command = read_file("src/command/report.rs");
+
+    assert!(report_command.contains("FEAT-REPORT-001"));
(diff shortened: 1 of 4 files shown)


tests/repository_quality.rs
```diff
@@ -1051,6 +1051,21 @@ fn repository_ships_browser_app() {
...
+    let report_command = read_file("src/command/report.rs");
+
+    assert!(report_command.contains("FEAT-REPORT-001"));
+    assert!(report_command.contains("load_workspace(&args.workspace)"));
+    assert!(report_command.contains("collect_check_result_from_workspace(&workspace)"));
+    assert!(report_command.contains("render_spec_coverage_summary(&workspace"));
+    assert!(!report_command.contains("cargo run -- list"));
+    assert!(!report_command.contains("cargo run -- show"));
+    assert!(!report_command.contains("run_syu_json"));
+    assert!(!report_command.contains("Command::new"));
```
tohboeh5: This assertion makes the contract much broader than the batching behavior the issue is trying to protect. Any future `Command::new(...)` added elsewhere in `src/command/report.rs` for unrelated output or diagnostics work would fail the test even if spec coverage still loads the workspace once and avoids repeated CLI relaunches. I think the test should stay focused on the spec-coverage path instead of banning every subprocess in the entire file.

tohboeh5:
Overview review:\n\nI like the motivation here. Protecting the batched spec-coverage path with a repository-quality test makes sense, and adding the trace link to `REQ-CORE-004` keeps the self-hosted spec honest about where that guarantee lives.\n\nThe current assertion set drifts a bit too far into file-shape coupling, though. In particular, banning every `Command::new` occurrence in `src/command/report.rs` is broader than the single-batch contract the issue is about, so harmless future work elsewhere in the file could fail this test for the wrong reason. I left one inline comment on that.

PR #523, opened 2026-04-19
docs: align docs home chooser with README (FEAT-DOCS-002, REQ-CORE-010)

### LINKED ISSUE #514: [Feature]: Keep the docs-site home page on the same task chooser path as the checked-in README

The checked-in README has a stronger newcomer chooser with task-based entry points like getting started, migration, reviewer workflow, app guide, and troubleshooting. The published docs-site home page is helpful, but it still sends users back to GitHub for the quick-start chooser instead of keeping the primary newcomer journey inside the docs site itself.

That split makes the published site feel like a secondary surface even though it is supposed to be the hosted documentation entry point.

### What change do you want?

Bring the docs-site home page closer to the README chooser so newcomers can stay inside the published docs experience. A good first scope would include:
- mirroring the README task-chooser structure or equivalent wording on the site home
- replacing GitHub-only quick-start links with site-local destinations where possible
- keeping README and docs-site onboarding copy aligned so they do not drift into two different newcomer stories

### Specification impact

This is documentation and docs-site UX work. It likely needs README/docs-site contract updates so both entry points stay aligned.

### Alternatives considered

The current site home already links to useful guides, but it still relies on GitHub README anchors for part of the primary onboarding path.

### Before submitting

### Evidence

- `README.md` has a dedicated chooser and task-based onboarding sections near `## Choose your path`.
- `website/src/pages/index.js:117-138` sends users to GitHub `#quick-start` anchors for both quick-start buttons instead of keeping them on site-local pages.
- `website/src/pages/index.js:167-186` has a journeys grid already, so the site has most of the structure needed but not the same primary chooser flow.

tohboeh5:
## Summary\n- add a task-oriented chooser section on the docs-site home page that mirrors the README entry points\n- replace GitHub quick-start hero links with site-local guide destinations\n- lock the chooser copy and local-link contract in repository-quality assertions\n\n## Linked issue or specification\n- Closes #514\n- Requirement / feature IDs: FEAT-DOCS-002, REQ-CORE-010\n\n## Validation\n- [x] scripts/ci/quality-gates.sh\n- [ ] scripts/ci/coverage.sh summary (required when Rust logic changes)\n- [x] cargo run -- validate .\n- [x] Docs, examples, or self-spec updated when behavior changed\n\n## Release notes\n- [ ] This change should appear in the next release notes\n- [x] No release note needed


website/src/pages/index.js
```diff
@@ -26,6 +26,63 @@ const layers = [
...
+    title: 'Getting started',
+    description:
+      'Choose this if you are new to syu, want the guided first-run path, and do not already know the four-layer model.',
+    to: '/docs/guide/getting-started'
+  },
+  {
+    title: 'Quick start',
+    description:
+      'Stay on the shortest site-local install-to-validate path when you want the fastest route into `syu validate .`.',
+    to: '/docs/guide/getting-started#before-you-begin'
```
tohboeh5: This `Quick start` card says it is the fastest route into `syu validate .`, but `#before-you-begin` lands on the installer preamble rather than the compact command path. That means the shortest entry point from the docs home now takes people to the long-form setup page before they see any of the actual quick-start commands. I think this should jump to the quick-start-oriented section instead of the preflight section.

tohboeh5:
Overview review:

I like the overall direction here. Moving the docs-site home onto the same task-oriented chooser language as the README makes the published site feel much less like a second-class entry point, and the repository-quality assertions are the right way to keep that copy from drifting.

The main gap is the new **Quick start** card target: it currently lands on the long-form `Before you begin` preamble instead of the shortest command path the card text promises. That mismatch weakens the newcomer UX right at the point where this PR is trying to make the docs chooser more trustworthy.

PR #524, opened 2026-04-20
feat: add TypeScript-only starter template

### LINKED ISSUE #515: [Feature]: Add a TypeScript-first starter template and example workspace

`syu` already understands TypeScript and JavaScript source discovery well enough to validate trace coverage, and the docs mention TypeScript/JavaScript as a supported rich-inspection path. But the starter/example matrix still has no TypeScript-first onboarding path.

That leaves one of the most common repository shapes on the generic/polyglot path even though the implementation language already has deeper support than many fallback cases.

### What change do you want?

Add a `typescript-only` (or equivalent JavaScript/TypeScript-first) starter template plus a matching checked-in example workspace. A good first scope would include:
- TypeScript-oriented IDs and starter copy
- a minimal `package.json`, source file, and test file
- docs that point TypeScript-first users to the new template/example instead of only to generic or polyglot starters

### Specification impact

This expands the starter/example onboarding surface for a language that already has rich trace support, so it likely touches init/template docs plus the self-hosted example/test contracts.

### Alternatives considered

Today users can start from `generic` or `polyglot`, but that keeps a well-supported ecosystem on the same path as repos that do not yet have a language-shaped starter at all.

### Before submitting

### Evidence

- `src/coverage.rs:15,47,531-583` already includes TypeScript target discovery and symbol inspection.
- `docs/guide/getting-started.md:248-266` calls out TypeScript/JavaScript as a current code-level tracing path.
- `docs/guide/examples-and-templates.md:29-34` still lists only docs-first, generic, Rust, Python, Go, and polyglot starters.

tohboeh5:
## Summary
- add a typescript-only starter template that scaffolds a minimal Node + TypeScript workspace
- add a matching checked-in examples/typescript-only workspace plus docs links for TypeScript-first onboarding
- extend init/template/example/repository-quality coverage so the new starter stays exercised end to end

Closes #515


examples/typescript-only/package.json
```diff
@@ -0,0 +1,13 @@
+{
```
tohboeh5, line 1: The new TypeScript starter introduces a Node-based workflow, but this package metadata still does not tell generated users which Node runtime the starter expects. The repo has been moving its Node surfaces toward explicit, reproducible runtime guidance (`.nvmrc` / pinned npm / engine metadata), and this starter currently drops people into `node --import tsx --test ...` with no version contract at all. I think the starter needs an explicit runtime hint here so the first run does not become a trial-and-error Node mismatch.

tohboeh5:
Overview review:

I like this addition a lot. The new starter fills a real onboarding gap for JavaScript/TypeScript-heavy repositories, and the end-to-end coverage through init/example/docs/tests is exactly the right shape for keeping a starter honest over time.

My main concern is runtime determinism. The generated TypeScript workspace now depends on a fairly specific Node execution path, but the starter metadata does not actually tell the user which Node version to use, so the supposedly smooth first-run path can still fail as soon as it lands in a shell with the wrong runtime.

examples/typescript-only/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "typescript-only",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "node --import tsx --test src/app.test.ts"
+  },
+  "devDependencies": {
```
tohboeh5, line 12: This starter imports `node:test` and `node:assert/strict`, but the package does not add `@types/node`. `tsc` and editor IntelliSense will report missing Node built-in type declarations in a fresh workspace unless the contributor installs them separately. Please add `@types/node` to the devDependencies so the TypeScript-first example works cleanly out of the box.

src/cli.rs
```diff
@@ -609,6 +609,8 @@ pub enum StarterTemplate {
     PythonOnly,
     GoOnly,
     JavaOnly,
+    #[value(name = "typescript-only", alias = "type-script-only")]
```
tohboeh5: `type-script-only` looks like an accidental alias typo rather than an intentional compatibility spelling. If we keep it, the CLI will silently bless a misspelled flag spelling forever; if we do not need backwards compatibility yet, I would remove this alias and keep the canonical `typescript-only` / `ts-only` forms only.

tests/help_command.rs
```diff
@@ -207,6 +207,7 @@ fn init_help_lists_starter_templates() {
     assert!(stdout.contains("python-only"));
     assert!(stdout.contains("go-only"));
     assert!(stdout.contains("java-only"));
+    assert!(stdout.contains("typescript-only"));
```
tohboeh5, line 210: The new assertion covers the template list, but `INIT_AFTER_HELP` in `src/cli.rs` still does not show a `syu init . --template typescript-only` example. That means the TypeScript starter is technically supported yet still missing from the most discoverable `init --help` examples. Please add one there too so the help text stays aligned with the template catalog.

PR #525, opened 2026-04-20
feat: add manual syu app refresh controls

### LINKED ISSUE #516: [Feature]: Add a manual refresh action and last-refresh timestamp to syu app

The browser app already keeps a validation snapshot and attempts to refresh it while the tab stays visible or when the user returns to the tab. But when that refresh path falls behind or fails, the current UX leaves users with a stale snapshot banner and no explicit way to force a reload or confirm how old the visible data is.

That makes review and debugging slower in the exact cases where the user most needs confidence about freshness.

### What change do you want?

Add explicit refresh affordances to `syu app`, such as:
- a visible "Refresh now" action in the header or stale-data banner
- a last-successful-refresh timestamp in the UI
- clearer state around whether the visible issues/trace counts are current or stale

### Specification impact

This is a browser-app UX improvement for the existing validation snapshot workflow. It may need updates to the app guide and self-hosted spec references.

### Alternatives considered

Today users can keep the tab visible, switch away and back, or restart `syu app`, but those are indirect recovery steps when the UI already knows refresh failed.

### Before submitting

### Evidence

- `docs/guide/app.md:134-140` documents that the app keeps the last successfully loaded snapshot visible when refresh fails.
- `docs/guide/app.md:180-183` tells users to keep the tab visible or switch back to it so refresh happens.
- `app/src/App.tsx:254-259` wires refresh to `visibilitychange`.
- `app/src/App.tsx:661-671` shows a stale refresh banner but no manual retry control or displayed timestamp.

tohboeh5:
## Summary
- add a visible Refresh now action plus current/stale/refreshing status in the syu app header and workspace card
- show the last successful refresh timestamp in the stale banner and core workspace chrome
- cover the manual refresh recovery flow in Playwright and document the new recovery path in the app guide

Closes #516

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -80,6 +80,7 @@ function App() {
   const [snapshotVersion, setSnapshotVersion] = useState<string | null>(null);
   const [isRefreshing, setIsRefreshing] = useState(false);
   const [refreshError, setRefreshError] = useState<string | null>(null);
+  const [lastSuccessfulRefreshAt, setLastSuccessfulRefreshAt] = useState<string | null>(null);
 
   const applyWorkspace = useCallback((browserWorkspace: BrowserWorkspace) => {
     setWorkspace(browserWorkspace);
@@ -156,6 +157,7 @@ function App() {
         setError(null);
         setRefreshError(null);
         setSnapshotVersion(snapshot);
+        setLastSuccessfulRefreshAt(new Date().toISOString());
         applyWorkspace(browserWorkspace);
       } catch (loadError) {
         if (refreshing) {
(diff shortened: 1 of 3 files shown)


app/src/App.tsx
```diff
@@ -607,16 +628,33 @@ function App() {
...
+              <button
+                type="button"
+                onClick={resetNavigation}
+                className="transition hover:text-sky-300"
+                aria-label="syu — go to first item"
+              >
+                syu
+              </button>
+            </h1>
+            <div className="flex items-center gap-3 md:hidden">
```
tohboeh5, line 642: The desktop header now shows the last successful refresh timestamp, but the mobile-only header block drops that information entirely and only keeps the state pill plus button. That means the stale-snapshot recovery story is noticeably weaker on small screens: a mobile user can retry, but still cannot tell how old the currently displayed snapshot is without scrolling down into the workspace card. I think the mobile header needs the same timestamp context if this is supposed to be a top-level recovery affordance.

tohboeh5:
Overview review:

This improves the app in the right direction. A visible retry button plus a success timestamp makes the stale-banner state much less passive, and the Playwright coverage is aimed at the exact recovery flow that matters.

The remaining UX gap is mobile. The desktop header gets the full timestamp context, but the mobile header only gets the pill and button, so small-screen users lose the key piece of information that tells them whether they are looking at a fresh or very old snapshot.

PR #526, opened 2026-04-20
feat: add explain command for spec chains

### LINKED ISSUE #489: [Feature]: Add a spec-explanation CLI that evaluates one symbol or ID against the full linked chain

The current CLI can trace a file/symbol and can relate a selector to connected spec items, but it still takes several commands to answer a higher-level question: "does this source symbol still make sense against the connected requirement, feature, policy, and philosophy chain?" That makes spec audits more manual than they need to be.

### What change do you want?

Add a command that accepts a source symbol, file path, or spec ID and prints a focused explanation of:
- the connected requirement/feature/policy/philosophy chain
- the traced files and symbols involved
- obvious gaps or mismatches in the current chain
- machine-readable output for automation or CI experiments

### Specification impact

This would extend the traceability and graph-inspection workflow rather than adding a new spec concept.

### Alternatives considered

Users can piece this together today with `trace`, `relate`, `show`, and `validate`, but that is slower and harder to teach than a single "explain this against the spec" command.

### Before submitting

tohboeh5:
## Summary
- add a new syu explain command that turns one ID, file, or symbol into a focused assessment with the connected chain, traces, and obvious gaps
- wire the command into CLI dispatch, help, spec coverage, and generated docs
- document the new reviewer-oriented path in the README and workflow guides

Closes #489


src/command/explain.rs
```diff
@@ -0,0 +1,335 @@
...
+            serde_json::to_string_pretty(&output)
+                .expect("serializing explain output to JSON should succeed")
+        ),
+    }
+
+    Ok(0)
+}
+
+fn build_explain_output(relation: JsonRelateOutput) -> ExplainOutput {
+    let assessment = if relation.gaps.is_empty() {
```
tohboeh5: This assessment only looks at `relation.gaps`, so an ambiguous selector that fans out to multiple direct matches can still be labeled `aligned` as long as the graph itself is internally complete. For a command whose job is to explain how **one** ID/path/symbol fits the chain, that is a misleading success signal: multiple equally plausible direct matches are exactly the case where the reviewer still needs attention. I think ambiguity needs to downgrade the assessment too, not just graph gaps.

tohboeh5:
Overview review:

The command shape is good. Reusing the relation graph instead of inventing a second traversal keeps the feature conceptually small, and the text/JSON outputs are both easy to scan.

The part I am not comfortable with yet is the top-level assessment. Right now the output can still say `aligned` in cases where the selector is ambiguous rather than cleanly explained, which makes the headline feel more confident than the underlying evidence really is.

PR #527, opened 2026-04-20
feat: expand reviewer log history

### LINKED ISSUE #488: [Feature]: Add a reviewer-focused git history mode that expands related spec items and traces

`syu log` is already useful for one traced requirement or feature, but reviewer workflows often start from a broader question: "show me the recent commit history for this requirement plus the related feature and traced implementation/test paths, preferably scoped to the current review range." The current command surface requires multiple manual follow-up lookups.

### What change do you want?

Add a reviewer-oriented history mode that can:
- expand one selected requirement or feature into its related traced files
- optionally include connected items from `syu relate`
- scope the output to a merge-base or explicit git range
- keep the output readable in both text and JSON modes

### Specification impact

This extends the git-backed reviewer tooling around traceable requirements and features.

### Alternatives considered

Today reviewers can script together `syu relate`, `syu trace`, and `syu log`, but that is much slower than a built-in review-focused path.

### Before submitting

tohboeh5:
## Summary
- add reviewer-oriented syu log options for related trace expansion and git-range scoping
- annotate matched history with the owning selected or related surface in text and JSON output
- document the new reviewer flow and extend log command coverage

Closes #488

### DIFF
--- a/src/command/log.rs
+++ b/src/command/log.rs
@@ -1194,6 +1194,57 @@ mod tests {
     }
 
     #[test]
+    fn collect_related_tracked_paths_excludes_selected_definition_from_related_set() {
+        let workspace_root = tempdir().expect("tempdir should exist");
+        write_related_workspace_fixture(workspace_root.path());
+        let workspace =
+            crate::workspace::load_workspace(workspace_root.path()).expect("workspace should load");
+
+        let tracked = super::collect_related_tracked_paths(
+            &workspace,
+            "REQ-HIST-001",
+            HistoryKind::Definition,
+            None,
+        )
+        .expect("related tracked paths should resolve");
+        assert!(tracked.iter().any(|entry| {
+            entry.owner_id == "FEAT-HIST-001"
+                && entry.owner_kind == "feature"
+                && entry.kind == "definition"
+                && entry.source == "related"
+        }));
+        assert!(
+            !tracked
+                .iter()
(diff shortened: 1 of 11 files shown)


src/command/log.rs
```diff
@@ -331,6 +421,67 @@ fn collect_trace_paths(
...
+                owner_kind: trace.owner_kind,
+                owner_id: trace.owner_id.clone(),
+                source: "related",
+                language: Some(trace.language.clone()),
+                symbols: trace.symbols.clone(),
+            });
+        }
+    }
+
+    if let Some(path_filter) = path_filter {
```
tohboeh5, line 468: Applying `--path` to the entire related tracked-path set here strips out the related definition documents as soon as the filter points at a source subtree like `src/command`. That means `--include-related --path ...` loses the very requirement/feature documents that give the reviewer the broader spec context, even though the new mode is explicitly about expanding review surface. I think the path filter should constrain related traces, but not silently discard the related definition nodes.

tohboeh5:
Overview review:

I like the direction here. `--include-related`, merge-base scoping, and the richer ownership metadata all make `syu log` much more reviewer-friendly without turning it into a second tracing command.

The main issue I found is the interaction between `--include-related` and `--path`: once you scope the log to a code subtree, the implementation currently drops the related definition documents too, which undercuts the broader review context this mode is supposed to add.

PR #529, opened 2026-04-20
feat: add reviewer workflow completions and spec-only validation

### LINKED ISSUE #381: [Feature]: Add shell completion and a spec-only mode for reviewer workflows

Reviewers working primarily on specification files still have to memorize IDs, type long commands, and run full validation flows that include source/trace checks unrelated to their change. That creates friction for spec-review workflows even though `syu` already models the layered spec clearly.

### What change do you want?

Add reviewer-oriented ergonomics such as:
- shell completion for spec IDs and common commands
- a `--spec-only` validation mode that limits checks to the spec graph and spec-document consistency when reviewers only need to verify YAML-side changes
- documentation that explains the intended reviewer workflow around these shortcuts

### Specification impact

This likely needs a new CLI feature plus requirement/feature docs for reviewer and contributor workflows.

### Alternatives considered

Reviewers can manually filter output today, but that still makes the common "review the spec change first" workflow more cumbersome than necessary.

### Before submitting

tohboeh5:
## Summary
- add a `syu completion <shell>` command for shell setup
- add `syu validate --spec-only` for reviewer-focused spec validation without trace checks
- document the reviewer workflow and update generated spec/report artifacts

Closes #381


src/command/check.rs
```diff
@@ -346,7 +359,8 @@ pub fn run_check_command(args: &CheckArgs) -> Result<i32> {
             } else {
                 with_validate_overrides(workspace, args)
             };
-            let mut result = collect_check_result_from_workspace(&workspace);
+            let mut result =
+                collect_check_result_from_workspace_with_mode(&workspace, args.spec_only);
```
tohboeh5, line 363: Running `syu validate . --spec-only` on this branch prints `traceability: requirements=0/0 ... features=0/0` on a workspace that definitely does have traces. Because this path just skips trace collection, the text report currently reads like “there are no trace declarations” instead of “trace checks were intentionally skipped.” For a reviewer-facing mode, that is pretty misleading. Can we plumb the mode into the rendered summary so it explicitly says trace validation was skipped rather than zeroing the counters?

tohboeh5:
Overview review:

There is a lot to like here: shell completions are a small but meaningful usability win, and `--spec-only` is a sensible reviewer entry point that fits the low-ceremony philosophy.

The part I would tighten before merging is the reviewer output contract. On this branch, `syu validate . --spec-only` prints `0/0` traceability counts, which reads like the workspace has no traced items instead of making it clear that trace enforcement was intentionally skipped. That undercuts the UX of the new mode because the very first summary line is easy to misread.

PR #531, opened 2026-04-20
feat: add C# trace validation support (FEAT-CHECK-001, REQ-CORE-002)

### LINKED ISSUE #314: [Feature]: Add C# language support to trace validation and symbol ownership
syu already validates traces for several source and document languages, and the current backlog is extending that support further with Go and Java. A common baseline language that is still missing is C#, which leaves .NET repositories without the same traceability workflow that Rust, Python, and TypeScript users already get.

That makes syu harder to adopt for teams whose main production code and tests live in C#.

### What change do you want?
Add C# language support for trace validation and symbol ownership, including both implementation and test discovery.

A good first cut would cover:
- class, method, property, field, and record symbol detection
- xUnit / NUnit / MSTest test discovery
- validation guidance that matches the existing language-adapter UX
- coverage mode support once the core adapter exists

### Specification impact
This likely extends the language adapter layer, trace validation tests, docs, and the supported-language references.

### Alternatives considered
Teams can keep using generic file-level ownership today, but that loses the symbol-level precision that makes syu valuable in multi-language repositories.

### Before submitting

tohboeh5:
## Summary
- add a built-in C# trace adapter plus strict coverage discovery for public symbols and test methods
- extend the passing fixture, integration tests, docs, and generated site-spec pages for the new C# support
- reframe the csharp-fallback example as a staged-adoption reference instead of an unsupported-language workaround

## Testing
- cargo run -- validate .
- npm --prefix website ci

Closes #314

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -49,6 +49,9 @@ struct GoAdapter;
 struct JavaAdapter;
 
 #[derive(Debug)]
+struct CSharpAdapter;
+
+#[derive(Debug)]
 struct ShellAdapter;
 
 #[derive(Debug)]
@@ -68,6 +71,7 @@ static PYTHON_ADAPTER: PythonAdapter = PythonAdapter;
 static TYPESCRIPT_ADAPTER: TypeScriptAdapter = TypeScriptAdapter;
 static GO_ADAPTER: GoAdapter = GoAdapter;
 static JAVA_ADAPTER: JavaAdapter = JavaAdapter;
+static CSHARP_ADAPTER: CSharpAdapter = CSharpAdapter;
 static SHELL_ADAPTER: ShellAdapter = ShellAdapter;
 static YAML_ADAPTER: YamlAdapter = YamlAdapter;
 static JSON_ADAPTER: JsonAdapter = JsonAdapter;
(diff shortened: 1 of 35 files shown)


src/language.rs
```diff
@@ -204,6 +208,37 @@ impl LanguageAdapter for JavaAdapter {
...
+    }
+
+    fn patterns(&self, symbol: &str) -> Vec<String> {
+        let escaped = regex::escape(symbol);
+        vec![
+            format!(
+                r"(?m)\b(?:public|protected|internal|private)?\s*(?:sealed\s+|abstract\s+|static\s+|partial\s+)*(?:class|interface|enum|record|struct)\s+{escaped}\b"
+            ),
+            format!(
+                r"(?m)\bpublic\s+(?:static\s+)?(?:sealed\s+|override\s+|virtual\s+|abstract\s+|async\s+|partial\s+|new\s+)*(?:<[^>{{}}]+>\s*)?(?:[\w<>\[\],?.]+\s+)+{escaped}\s*\("
```
tohboeh5: The new C# adapter only treats methods, properties, and fields as matches when they are `public`, while the other language adapters accept any explicitly traced symbol that exists in the file. That means a checked-in trace to an `internal` or private C# symbol will fail validation even though the symbol is present, which is a surprising regression in trace parity across languages. Can we keep coverage discovery public-only if that is the intended policy, but make explicit symbol existence checks language-agnostic the way the other adapters are?

tohboeh5:
Overview review:

This is a valuable expansion of the built-in trace model, and I like that it updates fixtures, docs, and examples instead of treating C# as a one-off parser tweak. The staged-adoption framing for the example also fits the repository philosophy well.

The main thing I would fix before merging is the explicit symbol matcher: right now the C# adapter only recognizes `public` members for several symbol forms, so valid traces to `internal` or private symbols can fail even though the symbol exists in the file. That makes C# trace validation stricter than the existing adapters in a way that feels accidental rather than intentional.

PR #533, opened 2026-04-20
feat: add four-layer audit command (FEAT-AUDIT-001, REQ-CORE-025)

### LINKED ISSUE #493: [Feature]: Add a contradiction and overlap audit for the four-layer spec model

The four-layer model is one of the repository's strongest ideas, but auditing for overlap or contradiction across philosophy, policy, requirements, and features is still mostly a manual reading exercise. As the workspace grows, that makes deeper consistency review slower than normal validation.

### What change do you want?

Add a generated audit or CLI command that helps maintainers spot:
- overlapping requirements that appear to say the same thing
- features that satisfy a requirement but pull against a different policy or philosophy
- orphaned policy language that no longer has concrete downstream obligations
- summary output that can be reviewed in PRs

### Specification impact

This extends workspace analysis and philosophy/policy consistency tooling.

### Alternatives considered

`syu validate` is great for structural rules, but it does not yet help much with semantic overlap or contradiction review.

### Before submitting

tohboeh5:
## Summary
- add `syu audit` with text and JSON output for overlap, tension, and orphaned-policy candidates
- trace the new command through the checked-in spec, guides, README, and generated docs
- cover workspace discovery and heuristic edge cases so the coverage gate stays green

## Testing
- cargo run -- validate .
- npm --prefix website ci

Closes #493


src/command/audit.rs
```diff
@@ -0,0 +1,439 @@
...
+    "a", "an", "and", "as", "be", "by", "for", "from", "in", "into", "is", "it", "must", "of",
+    "on", "or", "should", "that", "the", "to", "when", "with",
+];
+const OPPOSITE_TERMS: &[(&str, &str)] = &[
+    ("manual", "automatic"),
+    ("manual", "automated"),
+    ("always", "never"),
+    ("required", "optional"),
+    ("strict", "lightweight"),
+    ("browser", "terminal"),
```
tohboeh5: Running `cargo run -- audit . --format json` on this branch already produces several tension findings on the self-hosted repo that are not actual drift (for example the browser-vs-terminal pairs). Treating `browser` and `terminal` as opposite terms makes the default output noisy enough that reviewers will start ignoring the command. Can we either drop that pair or require stronger evidence than a single token match before emitting a tension finding?

tohboeh5:
Overview review:

I like the command shape and the fact that it stays read-only with both text and JSON output. The overlap/orphaned-policy heuristics are easy to understand, and the docs wiring is thorough.

My main concern is trust in the default findings. On this branch, `syu audit .` already reports several tension candidates on the self-hosted repo that look like false positives rather than real drift, especially around the browser/terminal pair. I would tighten that heuristic before merging so the new command starts from a high-signal baseline.

PR #545, opened 2026-04-20
docs: add shared contributor Node workflow guide (FEAT-DOCS-001, REQ-CORE-010)

### LINKED ISSUE #536: [Documentation]: Explain the repository's multi-Node local workflow in one place

The repository currently expects contributors to understand several different Node expectations at once: the browser app wants its own Node major, the docs site has a different engines.node range, the devcontainer installs node:lts, and editor-related workflows add their own version notes. That information is spread across multiple files, which makes the first browser/docs/editor task feel more fragile than it should.

### What change do you want?

Add one checked-in guide that explains which Node major each local surface expects today, when contributors need to switch versions, which tasks are safe with the default devcontainer/runtime setup, and the recommended commands for app, docs-site, and editor-extension work.

### Specification impact

This would extend contributor documentation and reduce onboarding ambiguity around local runtime expectations.

### Alternatives considered

Contributors can piece this together from .nvmrc, package.json, CI files, and docs, but that is a poor first-run experience.

### Before submitting

- [x] I can help validate the wording against the current repo setup.

tohboeh5:
## Summary
- add one guide that explains which Node major contributors should use for app, website, and VS Code work
- link the new guide from README, CONTRIBUTING, and VS Code extension docs
- cover the new guide and cross-links in repository quality checks

Closes #536


docs/guide/node-workflow.md
```diff
@@ -0,0 +1,112 @@
...
+## Do not trust the current shell by default
+
+The devcontainer installs `node:lts`, and a reused local shell may already be
+pointing at some other major. Treat the checked-in `.nvmrc` files as the source
+of truth for contributor tasks instead of assuming the current shell is already
+correct.
+
+When you switch tasks, switch Node first:
+
+```bash
```
tohboeh5: This block reads like a copy-paste setup recipe, but running all three `nvm use ...` commands in sequence leaves the shell on the last runtime (Node 20 from `editors/vscode`). That is the opposite of the one-place-for-fast-answers goal for someone who only wants app work, because the first snippet quietly steers them onto the wrong major. I would either replace this with per-surface examples or rewrite it so the first code block cannot be executed verbatim into the wrong runtime.

tohboeh5:
Overview review:

Centralizing the runtime guidance is a good move, and the quick matrix makes the app/docs/editor split easier to reason about. The main thing I would tighten before merging is the first switching example: the current three-command block looks like a copy-paste recipe but leaves the shell on the last runtime, which can steer contributors onto the wrong Node major for their next task.

PR #547, opened 2026-04-20
feat: add contributor tooling bootstrap script (REQ-CORE-013, FEAT-CONTRIB-004)

### LINKED ISSUE #543: [Feature]: Add a one-shot contributor bootstrap for optional app and docs-site tooling

The repository already has good opt-in tooling for browser-app and docs-site work, but contributors still need to know which commands to run, in which order, and when each surface is actually required. That makes fresh worktrees and intermittent docs/app tasks slower than they should be.

### What change do you want?

Add one checked-in bootstrap path, whether a script or a documented task entrypoint, that can prepare optional local surfaces together. A useful first scope would cover browser-app npm dependencies, docs-site npm dependencies, optional Playwright/browser install, and clear next-step messaging for contributors who only need a subset.

### Specification impact

This would improve contributor tooling and reduce repeated local setup friction.

### Alternatives considered

The current repo already documents the pieces, but contributors still have to compose the bootstrap manually.

### Before submitting

- [x] I can help validate the first cut against the current devcontainer and local docs flow.

tohboeh5:
## Summary
- add a checked-in bootstrap script for optional app, website, vscode, and Playwright contributor tooling
- document the new entrypoint in README and CONTRIBUTING
- extend the self-hosted contributor spec and generated docs for the new workflow

Closes #543


scripts/ci/bootstrap-contributor-tooling.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
+# FEAT-CONTRIB-004
+
+set -euo pipefail
+
+usage() {
+  cat >&2 <<'EOF'
+Usage: scripts/ci/bootstrap-contributor-tooling.sh [--app] [--website] [--vscode] [--playwright] [--all]
+
+Without flags, installs the default optional contributor surfaces:
```
tohboeh5: The default no-flag path installs both the browser app and the docs site together, but this repository intentionally expects different Node majors for those surfaces (`app/` uses Node 25 while `website/` uses Node 20). In a normal shell only one major can be active at once, so the advertised one-shot happy path is not actually deterministic. Please either narrow the default to one surface or make the workflow explicitly handle the cross-major switch before claiming that no flags is the default bootstrap.

tohboeh5:
Overview review:

The idea of a single bootstrap entrypoint is strong and the script/help/docs wiring is thorough. The main blocker is the default no-flag path: because this repo intentionally uses Node 25 for app work and Node 20 for website/editor work, bootstrapping app + website together from one shell does not actually give contributors a deterministic happy path yet. Once the default flow matches the multi-Node strategy, this will be a meaningful UX improvement.

PR #548, opened 2026-04-20
fix: require explicit opt-in for remote app binds (REQ-CORE-017, FEAT-APP-001)

### LINKED ISSUE #540: [Bug]: Require explicit opt-in before syu app binds to non-loopback addresses
### What area is affected?

Local browser app security and operator safety

### What happened?

syu app defaults to loopback, but the CLI and config still allow non-loopback binds such as 0.0.0.0. The runtime prints a warning, yet it will still serve workspace data and source-linked content on a network-visible address without an explicit confirmation or stronger safety gate.

Binding the local app to a non-loopback address should require a clearer, explicit opt-in such as a dedicated flag or a stronger safety acknowledgement path.

### syu version

main

### Environment

Any environment where contributors can set app.bind or --bind to a non-loopback address

### Steps to reproduce

1. Configure app.bind: 0.0.0.0 or pass syu app --bind 0.0.0.0.
2. Start the app.
3. Observe that the server still starts after only printing a warning.

The code already warns that workspace data and source documents may be reachable from other machines on the network.

### Before submitting

- [x] I can help validate a safer UX against the current app behavior.

tohboeh5:
## Summary
- add a dedicated --allow-remote opt-in before syu app accepts non-loopback binds
- keep the existing stdout warning path after the explicit opt-in is present
- update checked-in app/config docs and self-spec coverage for the safer startup flow

Closes #540

### DIFF
--- a/docs/generated/site-spec/config/app.md
+++ b/docs/generated/site-spec/config/app.md
@@ -40,7 +40,9 @@ description: "Generated reference for docs/syu/config/app.yaml"
     - |
       `syu app` uses this address when `--bind` is not passed. Repositories can
       keep a checked-in default when contributors usually inspect the app through
-      a shared local workflow or demo setup.
+      a shared local workflow or demo setup. Non-loopback addresses such as
+      `0.0.0.0` still require the explicit `--allow-remote` CLI opt-in before
+      the server starts.
 - **key**: app.port
   - **type**: integer
   - **default**: 3000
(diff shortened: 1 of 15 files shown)


tohboeh5:
Overview review:

I like the direction here. Requiring an explicit opt-in before `syu app` serves workspace data on a non-loopback address is a good default for an alpha tool, and the docs/tests updates keep the behavior discoverable. I did not find a blocking issue in the diff.

PR #549, opened 2026-04-20
feat: add JSON output for non-interactive browse snapshots

### LINKED ISSUE #539: [Feature]: Add JSON output for syu browse --non-interactive

syu browse --non-interactive is a nice review/debug snapshot, but it only prints plain text. That makes it awkward to reuse in CI experiments, editor integrations, or PR automation where teams want the same workspace snapshot in a machine-readable shape.

### What change do you want?

Add a JSON mode for the non-interactive browse snapshot that includes workspace metadata, per-layer counts, grouped items, and current validation issues.

### Specification impact

This would extend the browse CLI rather than introducing a new spec concept.

### Alternatives considered

Users can combine list, show, and validate --format json, but that loses the single high-level workspace snapshot that browse --non-interactive already provides.

### Before submitting

- [x] I can help validate the output shape against the current browse text snapshot.

tohboeh5:
## Summary
- add JSON output for syu browse --non-interactive with workspace metadata, grouped items, and validation issues
- document the browse snapshot as text-or-JSON in the CLI help, README, and getting-started guide
- extend browse/list/repository-quality coverage for the new non-interactive JSON contract

Closes #539

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -73,7 +73,7 @@ const WORKSPACE_HELP: &str = "Workspace root or any child directory; syu walks u
 
 const LIST_AFTER_HELP: &str = "\
 Choose `syu list` when you want list-shaped output that can be narrowed to one layer or emitted as JSON for automation.
-Choose `syu browse --non-interactive` when you want the browse snapshot instead: workspace metadata, per-layer counts, and the current validation errors in plain text.
+Choose `syu browse --non-interactive` when you want the browse snapshot instead: workspace metadata, per-layer counts, grouped items, and the current validation errors in text or JSON.
 
 Examples:
   syu list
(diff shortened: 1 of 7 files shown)


src/cli.rs
```diff
@@ -89,12 +89,13 @@ Note:
   Pass the workspace root, the configured spec.root directory, or any child directory.
   syu walks upward until it finds syu.yaml, then resolves the configured spec.root from that workspace.";
 const BROWSE_AFTER_HELP: &str = "\
-Choose `syu browse --non-interactive` when you want the browse snapshot in plain text: workspace metadata, per-layer counts, grouped items, and the current validation errors.
-Choose `syu list` when you want list-shaped output that can be narrowed to one layer or emitted as JSON for automation.
+Choose `syu browse --non-interactive` when you want the browse snapshot in text or JSON: workspace metadata, per-layer counts, grouped items, and the current validation errors.
+Choose `syu list` when you want list-shaped output that can be narrowed to one layer.
 
 Examples:
   syu browse .
```
tohboeh5, line 96: BROWSE_AFTER_HELP now says `syu list` is only for narrowing to one layer, but `syu list --format json` is still the automation-oriented list output described elsewhere in this patch. Dropping that hint makes the new help text regress discoverability for scripting users coming from browse help. Please keep the JSON/automation guidance here too so the two help blocks stay consistent.

tohboeh5:
Overview: the JSON browse snapshot itself is a strong addition, and I like that the tests exercise the new machine-readable contract directly. The main gap I found is in the updated help copy: `browse` help no longer tells users that `syu list` remains the JSON-friendly automation path, so the two help blocks drift out of sync and make scripting discovery worse. Once that wording is tightened back up, this looks close to ready.

PR #550, opened 2026-04-20
feat: align embedded app builds with pinned npm

### LINKED ISSUE #538: [Bug]: Align browser-app build.rs tooling with the repository pinned npm workflow
### What area is affected?

Local browser-app embedding during Rust builds

### What happened?

The repository uses checked-in pinned npm workflows in CI and contributor docs, but build.rs still shells out to raw node, npm, and npx from PATH. That makes local browser-app embedding less deterministic than the rest of the repo tooling and leaves contributors with different behavior depending on whichever Node/npm tooling happens to be first on their machine.

The browser-app build path used by Cargo should follow the same pinned or explicitly resolved toolchain story as the rest of the repository automation.

### syu version

main

### Environment

Contributor local builds with multiple Node/npm installations available

### Steps to reproduce

1. Compare the repository pinned npm scripts and CI guidance with build.rs.
2. Observe that build.rs directly invokes node, npm, and npx from PATH.
3. Note that this bypasses the repository more explicit toolchain story.

This is currently visible in the checked-in build logic rather than one specific runtime failure.

### Before submitting

- [x] I can help validate a first implementation against local and CI behavior.

tohboeh5:
## Summary
- parse app/package.json directly in build.rs instead of shelling out to node
- verify the app toolchain with scripts/ci/pinned-npm.sh before running embedded browser-app npm commands
- remove the npx fallback path so Cargo follows the repository's pinned npm workflow consistently

Closes #538


tohboeh5:
Overview: I like the direction here—moving build.rs off a Node-based packageManager probe and aligning the browser-app build with the repository helper makes the bootstrap path easier to reason about. The main issue is that the patch only verifies the pinned npm setup before continuing to run plain `npm` from PATH, so the actual build can still drift away from the pinned toolchain on machines with multiple npm installs. I left one inline comment on that gap; once the invoked commands are pinned as well as the preflight check, this should be in much better shape.

build.rs
```diff
@@ -169,30 +161,13 @@ fn needs_npm_ci(app_dir: &Path) -> bool {
 
 fn run_npm(
     app_dir: &Path,
-    required_npm: &str,
     args: &[String],
     action: &str,
     extra_env: &[(&str, String)],
 ) -> Result<(), String> {
```
tohboeh5, line 189: This change removes the `npx npm@<required>` fallback, but `run_npm` still invokes plain `npm` from PATH after the pinned helper check. That means a machine can pass `pinned-npm.sh check` and still build with a different globally-installed npm, which breaks the goal of following the pinned workflow consistently. Please route the actual build commands through the pinned helper too, not just the preflight check.

PR #551, opened 2026-04-20
feat: add Ruby starter and trace adapter support

### LINKED ISSUE #544: [Feature]: Add Ruby language support to trace validation and starter flows

syu already supports several mainstream implementation languages, but Ruby repositories still do not have a native trace adapter or starter story. That leaves teams to fall back to unsupported-language patterns even when they only need a fairly standard Ruby source and test workflow.

### What change do you want?

Add first-class Ruby support covering a Ruby trace adapter for file and symbol validation, starter or example coverage for a Ruby-first repository, and docs that place Ruby in the same capability matrix as the existing supported languages.

### Specification impact

This would extend language support, trace validation, examples, and the documentation matrix.

### Alternatives considered

Teams can use a fallback pattern today, but that is a weaker first-run story than native support.

### Before submitting

- [x] I can help validate the language shape against a simple Ruby project story.

tohboeh5:
## Summary
- add a built-in Ruby trace adapter so explicit ruby mappings validate against .rb files
- add a ruby-only starter template plus a checked-in example workspace
- document Ruby in the starter and trace-adapter guides

Closes #544


src/language.rs
```diff
@@ -122,6 +126,30 @@ impl LanguageAdapter for PythonAdapter {
...
+
+    fn patterns(&self, symbol: &str) -> Vec<String> {
+        let escaped = regex::escape(symbol);
+        vec![
+            format!(r"(?m)\b(?:class|module)\s+{escaped}\b"),
+            format!(r"(?m)^\s*def\s+(?:self\.)?{escaped}\b"),
+            format!(r"(?m)^\s*{escaped}\s*="),
+            format!(r"(?m)\b{escaped}\b"),
+        ]
+    }
```
tohboeh5, line 152: The final fallback pattern here is too permissive for validation use: `\\b{escaped}\\b` will match comments, strings, and ordinary prose anywhere in the file, so a Ruby trace can pass even when no class/module/method/assignment with that symbol actually exists. That makes Ruby support noticeably less trustworthy than the other adapters. Please drop the generic word-boundary fallback and keep the matcher constrained to real Ruby declaration shapes.

tohboeh5:
Overview: adding Ruby starter/example support is a valuable expansion, and the docs plus scaffold coverage make the feature feel well integrated instead of bolted on. The main problem I found is in the Ruby symbol matcher: the final fallback regex is broad enough to accept arbitrary word mentions in comments or strings, so validation can succeed without a real declaration being present. I left that inline because it weakens trust in the new adapter; tightening the matcher to declaration-shaped patterns would make this much safer.

PR #553, opened 2026-04-20
feat: add reviewer git range workflows

### LINKED ISSUE #542: [Feature]: Add git-range-aware reviewer entry points for syu trace and syu relate

Reviewer workflows still require a manual jump from a git diff to syu trace or syu relate. Even with the current command set, a reviewer usually has to collect changed files first and then run several lookups one by one.

### What change do you want?

Add a reviewer mode that starts from a git range or merge-base and expands the touched files or symbols through trace and relate. The output should resolve changed traced files automatically, group the connected requirements, features, policies, and philosophies, and stay readable in both text and JSON modes.

### Specification impact

This would extend the existing traceability and reviewer CLI workflow rather than adding a new model concept.

### Alternatives considered

Today reviewers can script together git diff, syu trace, and syu relate, but that is more manual than it needs to be.

### Before submitting

- [x] I can help validate the first scope against real PR review flows.

tohboeh5:
## Summary
- add trace --range and relate --range reviewer workflows backed by git diff file discovery
- cover range-specific success, empty, and failure cases across command and unit tests
- keep text and JSON outputs aligned for changed-file summaries and traced gaps

Closes #542


src/command/relate.rs
```diff
@@ -123,6 +156,109 @@ pub fn run_relate_command(args: &RelateArgs) -> Result<i32> {
...
+    let report = JsonRelateRangeOutput {
+        range: range.to_string(),
+        philosophies: catalog.nodes_for(LookupKind::Philosophy, &expanded_ids.philosophies),
+        policies: catalog.nodes_for(LookupKind::Policy, &expanded_ids.policies),
+        requirements: catalog.nodes_for(LookupKind::Requirement, &expanded_ids.requirements),
+        features: catalog.nodes_for(LookupKind::Feature, &expanded_ids.features),
+        traces: collect_related_traces(
+            workspace,
+            &expanded_ids,
+            &SelectionSource::Path {
```
tohboeh5: Range mode currently passes a synthetic path (`"multiple files"`) into `collect_related_traces`, but `selection_matches_reference` only marks traces as direct when the source path exactly equals `reference.file`. In practice that means every trace in `syu relate --range ...` is reported as indirect, so the reviewer loses the exact changed-file signal this feature is supposed to add.

tohboeh5:
Overview: the git-range entry points are a good reviewer-facing addition, especially because they reuse the existing trace/relate model instead of inventing a second workflow. The main issue I found is in the relate range implementation: once the command aggregates multiple files, it loses the ability to mark which traces were direct matches for the changed paths, so the richer output becomes less trustworthy at exactly the point where a reviewer needs that signal most. Tightening that direct-match bookkeeping would make the new workflow much easier to rely on in real PR review.

PR #554, opened 2026-04-20
feat: add editor-facing lsp server

### LINKED ISSUE #313: [Feature]: Extract an LSP server for syu editor integrations
Editor integrations will be much easier to maintain if syu exposes a language-server-style interface instead of forcing each editor plugin to shell out to ad-hoc CLI commands.

Right now the repository has the validation, lookup, and trace logic inside the Rust CLI, but no standalone server process that other editors or tools can reuse.

### What change do you want?
Add an LSP or LSP-like server that exposes syu diagnostics and navigation primitives for editor clients.

A useful first scope would cover:
- validation diagnostics for YAML and traced source files
- go-to-definition / go-to-related-item for spec IDs and traced symbols
- hover details for linked requirements/features/policies/philosophies
- a protocol stable enough for both VS Code and non-VS Code clients

### Specification impact
This likely needs a new feature entry, protocol documentation, and tests around request/response behavior.

### Alternatives considered
Editor plugins can shell out to syu commands today, but that leads to duplicated glue code and makes richer editor experiences harder to build consistently.

### Before submitting

tohboeh5:
## Summary
- add the syu lsp subcommand and shared LSP protocol/server handling for editor integrations
- implement initialize and hover flows backed by the existing workspace and specification model
- trace the new editor-facing feature and refresh generated specification outputs

Closes #313

### DIFF
--- a/src/lsp/handlers.rs
+++ b/src/lsp/handlers.rs
@@ -1,16 +1,16 @@
 // FEAT-LSP-001
 // REQ-CORE-001
 
-use anyhow::{Result, bail};
 use regex::Regex;
 use serde_json::Value;
-use std::path::Path;
+use std::path::PathBuf;
 use std::sync::LazyLock;
+use url::Url;
 
 use crate::workspace::{Workspace, load_workspace};
 
 use super::protocol::{
-    Hover, InitializeParams, InitializeResult, MarkupContent, ServerCapabilities,
+    Hover, InitializeParams, InitializeResult, LspError, MarkupContent, ServerCapabilities,
     TextDocumentPositionParams,
 };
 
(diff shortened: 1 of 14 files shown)


src/lsp/handlers.rs
```diff
@@ -0,0 +1,361 @@
...
+            && let Some(hover_content) = create_hover_for_spec_id(workspace, &spec_id)
+        {
+            return Ok(Some(hover_content));
+        }
+
+        Ok(None)
+    }
+}
+
+fn uri_to_path(uri: &str) -> Result<std::path::PathBuf> {
```
tohboeh5: `uri_to_path` treats every `file://` URI as a raw path substring, so percent-encoded paths (`%20`, non-ASCII bytes) and Windows-style file URIs will resolve to the wrong file. Real editor clients send proper URI encoding here, so hover will quietly fail on perfectly valid workspaces unless this goes through URI decoding instead of string stripping.

src/lsp/server.rs
```diff
@@ -0,0 +1,353 @@
...
+                id: request.id,
+                result: Some(value),
+                error: None,
+            }),
+            Err(e) => Ok(Response {
+                jsonrpc: "2.0".to_string(),
+                id: request.id,
+                result: None,
+                error: Some(ResponseError {
+                    code: super::protocol::error_codes::INTERNAL_ERROR,
```
tohboeh5: The protocol module already defines `METHOD_NOT_FOUND` and `INVALID_PARAMS`, but this branch collapses all request failures into `INTERNAL_ERROR`. That makes unsupported requests and bad client input look like server bugs to editor clients, which is exactly the distinction LSP error codes are meant to preserve.

tohboeh5:
Overview: the LSP extraction is a strong direction, and I like that the PR keeps the first scope intentionally small instead of pretending to ship a full editor platform on day one. The current implementation is close to a useful foundation, but I found two protocol-level issues that feel important before other clients build on it: URI handling is still too naive for real `file://` traffic, and request failures are flattened into `INTERNAL_ERROR` instead of returning the more precise LSP error codes clients expect. Once those are tightened up, this will be much safer to treat as the base for editor integrations.

PR #555, opened 2026-04-20
feat: add local readiness doctor command

### LINKED ISSUE #537: [Feature]: Add a syu doctor command for local prerequisite checks

Local setup checks are currently scattered across README snippets, contributor docs, build errors, and CI scripts. That makes it hard to answer simple questions like whether the right runtimes are installed, whether the app or docs-site npm dependencies are missing, and whether the current workspace/toolchain supports the command a contributor is about to run.

### What change do you want?

Add a syu doctor command that reports local readiness for common workflows, including Rust/Cargo prerequisites, browser-app dependency state, docs-site dependency state, optional Playwright/browser readiness, and key runtime versions with obvious mismatch warnings.

### Specification impact

This would extend the contributor-facing CLI surface and onboarding workflow.

### Alternatives considered

The current approach relies on failed commands teaching the user what to install next, but that is slower and harder to automate.

### Before submitting

- [x] I can help validate the first scope against current contributor workflows.

tohboeh5:
## Summary
- add `syu doctor` so contributors can verify workspace config, Rust/Cargo, Node/npm, dependency freshness, and Playwright readiness before deeper checks
- trace the command through the spec/docs so the workflow is discoverable from the README and contributing guide
- cover the new command with help, integration, and full-coverage unit tests without mutating global process environment

Closes #537


src/command/doctor.rs
```diff
@@ -0,0 +1,1221 @@
...
+                id: surface_check_id(relative_dir, "deps"),
+                label: surface_label(label_prefix, "dependencies"),
+                status: DoctorStatus::Skipped,
+                message: format!("`{}` is not present", package_json.display()),
+                fix: None,
+            },
+        ];
+    }
+
+    let metadata = PackageMetadata::from_path(&package_json).unwrap_or_default();
```
tohboeh5: `PackageMetadata::from_path` can fail for exactly the workspace states this command is supposed to diagnose (broken or half-edited `package.json`). Falling back to `Default` here quietly turns that into “no npm requirement declared”, so `syu doctor` can report the npm check as ok/skipped instead of surfacing the malformed package metadata that will block contributors. I would convert the read/parse failure into an explicit warning/error check instead of swallowing it.

src/command/doctor.rs
```diff
@@ -0,0 +1,1221 @@
...
+    let workspace_root = resolve_workspace_root(workspace)?;
+    let mut checks = Vec::new();
+
+    checks.push(workspace_config_check(&workspace_root));
+    checks.extend(rust_toolchain_checks(&workspace_root));
+    checks.extend(surface_checks(
+        &workspace_root,
+        "app",
+        "Browser app",
+        "scripts/ci/pinned-npm.sh install app && npm --prefix app ci",
```
tohboeh5: The fix strings here still recommend plain `npm`/`npx` from PATH. That undercuts the repositorys pinned npm workflow and can send contributors through the wrong npm version on machines that juggle multiple Node installs—the exact situation `doctor` is meant to clarify. I would route these suggestions through `scripts/ci/pinned-npm.sh exec ...` as well, including the Playwright install hint below, so the remediation path matches the documented toolchain contract.

tohboeh5:
Overview: the doctor command is a strong addition for first-run contributor UX, and I like that the PR wires it through specs, docs, help text, and integration coverage instead of treating it as a one-off utility. The output is easy to scan and the local checks line up with the repo's real setup friction points. The two issues I left inline are both about trust in the remediation path: package metadata parse failures are currently swallowed instead of surfaced, and several suggested fixes still point people at unpinned npm/npx commands even though this repository depends on pinned toolchain flows. Once those are tightened up, this should be a very solid readiness entry point.

PR #566, opened 2026-04-21
fix(REQ-CORE-010): include typescript starter in init help

### LINKED ISSUE #556: [Bug]: `syu init --help` omits `typescript-only` from the starter-template description
### What area is affected?

CLI or runtime behavior

### What happened?

`syu init --help` advertises a stale starter-template list. The `--template` help string still says `generic, docs-first, rust-only, python-only, go-only, java-only, or polyglot`, but the same help output also shows `typescript-only` in the generated possible-values list.

That makes the first-line explanation less trustworthy for newcomers because one part of the help says TypeScript is supported while the descriptive help text omits it.

The human-written `--template` help text should stay in sync with the actual `StarterTemplate` enum values and include every supported starter.

### syu version

main

### Environment

Linux, Rust 1.88, `cargo run -- init --help`

### Steps to reproduce

1. Run `cargo run -- init --help` from the repository root.
2. Read the `--template <TEMPLATE>` help text.
3. Compare it with the generated `[possible values: ...]` list on the same line.
4. Observe that `typescript-only` is present in possible values but missing from the prose help text.

`src/cli.rs` currently hard-codes the shorter list in the `InitArgs.template` help string, while the enum-derived possible-values list includes `typescript-only`.

### Before submitting

tohboeh5:
## Summary
- replace the hand-maintained `syu init --help` template prose list with a stable pointer to `syu templates`
- keep the existing init-help coverage focused on the rendered catalog and possible-values output instead of duplicating variant names in prose
- leave the existing REQ-CORE-010 help coverage intact without adding another drift-prone spec test

- Closing keyword issue: Closes #556
- Requirement / feature IDs: REQ-CORE-010

## Validation

## Release notes
- [x] No release note needed


src/cli.rs
```diff
@@ -650,7 +650,7 @@ pub struct InitArgs {
     pub spec_root: Option<PathBuf>,
 
     #[arg(
-        help = "Starter layout to scaffold (generic, docs-first, rust-only, python-only, ruby-only, go-only, java-only, or polyglot)"
+        help = "Starter layout to scaffold (generic, docs-first, rust-only, python-only, ruby-only, go-only, java-only, typescript-only, or polyglot)"
```
tohboeh5: This still hard-codes the starter list in prose, so the next template addition can drift again in exactly the same place. Please derive this sentence from the shared `StarterTemplate` catalog (or stop enumerating the variants in prose) so the help text and value enum stay coupled.

tohboeh5:
Overview:

Good catch on the newcomer-facing mismatch, and I like that you added a regression test for the rendered help output. The blocking issue for me is that the fix repeats the starter-template list as hand-written prose in `src/cli.rs`, so the same drift can happen again the next time a template is added. CI is also red right now because this PR changes `docs/syu/` but the PR body is missing the required linked issue/spec section from the template. Once the help text is derived from a single source of truth and the PR metadata is updated, this should be much easier to trust long-term.

PR #567, opened 2026-04-21
docs(FEAT-VSCODE-002): add dedicated syu lsp guide

### LINKED ISSUE #557: [Feature]: Add a dedicated `syu lsp` guide with transport, capability, and troubleshooting details

The repository now ships a `syu lsp` command, but the checked-in docs do not explain how to use it directly. `cargo run -- lsp --help` only prints a one-line description, `README.md` has no `syu lsp` entry, and `docs/guide/` has no dedicated LSP guide.

That leaves editor authors and contributors without a discoverable place to learn the transport (`stdio`), supported requests, and debugging workflow.

### What change do you want?

Add a dedicated guide for `syu lsp` that documents:
- how to launch it from an installed binary or `cargo run`
- the transport model (JSON-RPC 2.0 over stdio)
- the currently supported requests/capabilities
- a minimal client configuration example
- troubleshooting steps for initialization failures

### Specification impact

This would extend editor-integration documentation around the existing LSP feature in `docs/syu/features/editor/vscode.yaml`.

### Alternatives considered

Relying on `--help` output alone is too thin, and folding LSP details into the VS Code guide makes non-VS Code client support harder to discover.

### Before submitting

tohboeh5:
## Summary
- document that `syu lsp` currently reads `rootUri` or falls back to the server cwd
- call out that `workspaceFolders` and `rootPath` are not consumed yet so client authors know when they need an explicit override
- keep the repository-quality guard aligned with that initialization guidance

- Closing keyword issue: Closes #557
- Requirement / feature IDs: FEAT-VSCODE-002

## Validation

## Release notes
- [x] No release note needed


docs/guide/lsp.md
```diff
@@ -0,0 +1,148 @@
...
+- `Content-Length: <bytes>`
+- blank line
+- JSON payload
+
+The expected startup flow is:
+
+1. send `initialize`
+2. send `initialized`
+3. send requests such as `textDocument/hover`
+4. send `shutdown`
```
tohboeh5, line 50: Please call out explicitly here that the current server only reads `rootUri` (or falls back to the process cwd) and does **not** consume `workspaceFolders` / `rootPath` yet. Many clients send workspace folders instead of `rootUri`, so the current wording is more permissive than the implementation and will send editor authors down the wrong debugging path.

tohboeh5:
Overview:

This is a strong documentation addition overall: the guide is discoverable from the right entry points, the scope stays intentionally narrow, and the troubleshooting section makes the current hover-only surface easy to understand. My main concern is precision around initialization semantics. The guide currently reads as if any standard workspace-root shape is fine, but the implementation only looks at `rootUri` or the server process cwd today. Tightening that language would make the doc much more trustworthy for editor authors who are wiring a real client.

PR #568, opened 2026-04-21
docs(FEAT-VSCODE-001): clarify VS Code extension and LSP roles

### LINKED ISSUE #558: [Feature]: Clarify how the VS Code extension should relate to the new `syu lsp` server

The VS Code extension guide still describes the extension as a CLI-and-YAML-based integration only:
- diagnostics come from `syu validate . --format json`
- navigation reads checked-in YAML directly

That was accurate before `syu lsp` landed, but now the guide does not explain whether the extension will stay CLI-backed, how the new LSP server relates to the extension, or what migration path extension contributors should expect.

### What change do you want?

Update `docs/guide/vscode-extension.md` so it explicitly covers:
- the current extension architecture versus the new `syu lsp` server
- whether the extension should remain CLI-backed for now or start adopting LSP capabilities
- which editor features still require CLI/YAML access versus which can move to LSP
- the short-term roadmap for extension contributors

### Specification impact

This is a documentation and contributor-workflow update around the existing VS Code and editor-integration features.

### Alternatives considered

Keeping the guide silent forces contributors to infer architecture from source code and merged PR history, which is slower and more error-prone than a checked-in statement of intent.

### Before submitting

tohboeh5:
## Summary
- replace the time-relative VS Code/LSP wording with guidance that stays correct after adjacent docs land
- keep the extension/LSP role split explicit without telling contributors to reason about PR timing
- leave the existing roadmap and repository-quality coverage intact

- Closing keyword issue: Closes #558
- Requirement / feature IDs: FEAT-VSCODE-001

## Validation

## Release notes
- [x] No release note needed


docs/guide/vscode-extension.md
```diff
@@ -21,6 +21,46 @@ This first cut stays deliberately small:
...
+**CLI-backed first** today rather than LSP-backed first.
+
+Current split:
+
+- **Still uses the CLI directly**: validation diagnostics come from
+  `syu validate . --format json`
+- **Still reads checked-in YAML directly**: the tree view, spec-ID lookups, and
+  related-file navigation read the workspace files without a language server in
+  the middle
+- **LSP server exists as a shared editor foundation**: `syu lsp` already gives
```
tohboeh5, line 39: This sentence is going to go stale almost immediately: it tells contributors to inspect `src/lsp/` "until the dedicated guide lands", but the repository already has a separate issue/PR for that guide. Please point to a stable source of truth here (either link the guide once available or describe the current limitation without time-relative wording) so this section does not age into confusing historical context.

tohboeh5:
Overview:

The architecture split is much clearer after this change, especially the explicit "CLI-backed first" framing and the short-term roadmap for contributors. The part I would still tighten is the time-relative guidance around the LSP docs: telling readers to inspect `src/lsp/` "until the dedicated guide lands" will age poorly and become confusing once that guide exists. Replacing that with a stable pointer would make this feel like long-lived contributor guidance instead of a snapshot of PR timing.

PR #569, opened 2026-04-21
fix(FEAT-APP-001): add baseline security headers to app assets

### LINKED ISSUE #559: [Bug]: `syu app` serves embedded assets without baseline security headers
### What area is affected?

CLI or runtime behavior

### What happened?

`syu app` serves embedded HTML/CSS/JS/Wasm assets without baseline security headers. In `src/command/app.rs`, `asset_response()` currently sets only `Content-Type` and returns the response.

That means the local UI is missing common defense-in-depth headers such as:
- `Content-Security-Policy`
- `X-Content-Type-Options: nosniff`
- `X-Frame-Options`
- `Referrer-Policy`

Even though the app is local-first, the command also supports remote binding behind `--allow-remote`, so adding safe defaults would reduce avoidable exposure.

Asset responses from `syu app` should include a minimal baseline set of security headers that matches the app's static asset model.

### syu version

main

### Environment

Linux, local `syu app` server, browser or `curl -I`

### Steps to reproduce

1. Run `syu app .` in a workspace.
2. Fetch the root document headers with `curl -I http://127.0.0.1:3000/`.
3. Observe that the response has `Content-Type` but no CSP, `nosniff`, frame, or referrer-policy headers.

`src/command/app.rs:606-612` builds the asset response and inserts only the `Content-Type` header.

### Before submitting

tohboeh5:
## Summary
- extend the shared asset-response regression test so it covers a non-HTML asset as well as the root document
- keep the new baseline security headers pinned across SVG responses, not just the index shell
- rerun the app-facing validation flow including Playwright coverage

- Closing keyword issue: Closes #559

## Validation

## Release notes
- [x] No release note needed


src/command/app.rs
```diff
@@ -2203,6 +2219,24 @@ mod tests {
             root.headers().get(header::CONTENT_TYPE),
```
tohboeh5, line 2219: Because `asset_response()` is shared by every embedded asset, I would extend this test to assert the headers on at least one non-HTML asset too (for example `/assets/...js` or the fallback-free nested asset case). Right now the regression coverage only proves the root document path, so a later special-case on JS/CSS/Wasm responses could silently drop the hardening headers without failing this test.

tohboeh5:
Overview:

This moves the app in a good direction: the chosen headers are sensible baseline hardening for a local-first UI, and wiring them through the shared asset response helper keeps the behavior centralized. The gap I still see is regression coverage. The new assertions only exercise the root HTML response even though the helper now mutates every embedded asset response, so I would like one non-HTML asset assertion before I trust this against future refactors.

PR #570, opened 2026-04-21
feat(FEAT-APP-001): add app dev-server mode

### LINKED ISSUE #560: [Feature]: Add an opt-in dev-server mode for faster `syu app` frontend iteration

The repository already ships a frontend dev server in `app/package.json` (`npm run dev` on port 4173), but `syu app` only serves the compile-time embedded dist.

That makes browser-app iteration slower than it needs to be because contributors have to rebuild frontend assets and the Rust binary before checking each UI change.

### What change do you want?

Add an explicit development-mode path for `syu app`, for example a `--dev-server` or `--proxy-app-dev-server` flag, so contributors can point the CLI server at the existing Vite dev server during local UI work.

A good first scope would:
- keep the current embedded-dist behavior as the default
- proxy or redirect to the existing dev server only when opted in
- document the workflow in `docs/guide/app.md` and `CONTRIBUTING.md`

### Specification impact

This would extend the developer-experience story around `FEAT-APP-001` without changing release/distribution behavior.

### Alternatives considered

Rebuilding embedded assets after every UI edit works, but it creates a much slower loop than the dev server the repository already maintains for frontend work.

### Before submitting

tohboeh5:
## Summary
- fail `syu app --dev-server` early when the frontend dev server is not actually reachable
- probe `http://127.0.0.1:4173/@vite/client` before printing the normal ready messages
- cover both the happy path and the missing-dev-server failure path in app command tests

- Closing keyword issue: Closes #560

## Validation

## Release notes
- [x] No release note needed


tohboeh5:
Overview:

I like the shape of this feature: keeping `/api/*` on the Rust side while letting the browser load Vite modules is the right low-friction development loop, and the docs/tests do a good job explaining the intended workflow. The blocker for me is startup trust. The command still prints the usual ready output after probing only the Rust server, so `--dev-server` can look healthy even when the Vite server is missing and the browser shell is unusable. I would want a dev-server reachability check before this becomes the recommended frontend iteration path.

src/command/app.rs
tohboeh5, line 1: One blocker in the current implementation: `--dev-server` never verifies that the Vite server is reachable before the command prints the normal ready messages. That means `/health` can succeed, startup looks healthy, and the browser still renders a broken shell because `http://127.0.0.1:4173` is down. Please add a dev-server reachability check (or an explicit startup failure/warning path) before treating this as a successful fast-path.

PR #571, opened 2026-04-21
fix(FEAT-TRACE-001): report skipped files in trace range output

### LINKED ISSUE #561: [Bug]: `syu trace --range` silently skips changed files that fail workspace normalization
### What area is affected?

CLI or runtime behavior

### What happened?

`syu trace --range ...` silently drops changed files when they normalize outside the workspace root or fail path validation. In `src/command/trace.rs`, the range loop catches `normalize_lookup_file()` errors and immediately `continue`s with a comment saying `Skip files outside workspace or with invalid paths`.

That means users can run a git-range trace audit and get a clean-looking summary without any indication that some changed files were never evaluated.

If `syu trace --range` skips changed files, the command should report which files were skipped and why instead of silently omitting them from the final summary.

### syu version

main

### Environment

Linux, `syu trace --range origin/main...HEAD`

### Steps to reproduce

1. Create a workspace and a git range that includes at least one changed file outside the normalized workspace root (or a path that resolves with `..`).
2. Run `syu trace --range origin/main...HEAD`.
3. Observe that the file is absent from the summary and owner grouping output.
4. Inspect `src/command/trace.rs:201-210` and note the silent `continue` path.

`src/command/trace.rs:209-210`:

```rust
// Skip files outside workspace or with invalid paths
continue;
```

### Before submitting

tohboeh5:
## Summary
- report files skipped during `syu trace --range` normalization instead of silently dropping them in text output
- show the full changed-file count in text mode even when some files are skipped
- cover skipped-file collection and rendering with trace unit tests

- Closing keyword issue: Closes #561
- Requirement / feature IDs: FEAT-TRACE-001

## Validation

## Release notes
- [x] No release note needed


src/command/trace.rs
```diff
@@ -273,14 +270,39 @@ fn compute_range_summary(results: &[TraceLookupOutput]) -> TraceRangeSummary {
...
+            Err(error) => {
+                skipped.push(TraceSkippedFile {
+                    file: file.display().to_string(),
+                    reason: error.to_string(),
+                });
+            }
+        }
+    }
+    (results, skipped)
+}
```
tohboeh5, line 294: This patch introduces two different meanings for "total files": `changed_files_total` now means files in the git range, while `TraceRangeSummary.total_files` still means only inspected results. That is easy to misuse later (especially when JSON work lands right after this). Please rename the summary field or split the model so the code makes the distinction explicit instead of carrying two incompatible totals under similar names.

tohboeh5:
Overview:

I agree with the direction here: surfacing skipped files in text mode is a big UX improvement over silently dropping them, and the new tests cover the basic regression path well. The thing I would still clean up is the data model. After this patch, the code has both `changed_files_total` and `TraceRangeSummary.total_files`, but they mean different things (all changed files vs. inspected files). That ambiguity is likely to trip the next change, especially with the JSON-side follow-up already in flight.

PR #572, opened 2026-04-21
feat(FEAT-TRACE-001): add skipped-file accounting to trace ranges

### LINKED ISSUE #562: [Feature]: Add explicit skipped-file accounting to `syu trace --range` text and JSON output

Automation that consumes `syu trace --range --format json` cannot tell whether the command inspected every changed file. The JSON output only returns `files` plus the final summary counts; it has no `skipped_files` or `errors` section for range-level omissions.

That makes CI or review tooling treat "0 skipped, all scanned" and "some files silently omitted" as the same shape.

### What change do you want?

Extend `syu trace --range` output with explicit skipped-file accounting, for example:
- `skipped_files` array in JSON with `file` and `reason`
- a text-mode section such as `Skipped files:` below the coverage summary
- summary counters that distinguish `changed_files_total`, `inspected_files`, and `skipped_files`

### Specification impact

This would extend `FEAT-TRACE-001` output rather than adding a new spec concept.

### Alternatives considered

Relying on stderr warnings would be harder for JSON consumers than a first-class structured field in the command output.

### Before submitting

tohboeh5:
## Summary
- add `skipped_files` to `syu trace --range` JSON output and expose changed / inspected / skipped counters in the summary
- surface the same counters in text output and keep per-file skipped details visible
- remove the redundant `summary.total_files` field now that the explicit counters are present

- Closing keyword issue: Closes #562
- Requirement / feature IDs: FEAT-TRACE-001

## Validation

## Release notes
- [x] No release note needed


src/command/trace.rs
```diff
@@ -105,11 +105,15 @@ struct TraceLookupOutput {
 struct TraceRangeOutput {
     range: String,
     files: Vec<TraceLookupOutput>,
+    skipped_files: Vec<TraceSkippedFile>,
     summary: TraceRangeSummary,
 }
 
 #[derive(Debug, Clone, Serialize)]
 struct TraceRangeSummary {
```
tohboeh5, line 113: Now that the summary exposes `changed_files_total`, `inspected_files`, and `skipped_files`, keeping `total_files` as a fourth counter is ambiguous because it is just another name for `inspected_files`. JSON consumers have no way to know whether `total_files` means "all changed" or "inspected only" without reading the implementation. I would either remove it or rename it so the schema has exactly one term for each concept.

tohboeh5:
Overview:

This is the right shape for the feature: the new `skipped_files` array makes automation safer, and adding explicit changed/inspected/skipped counters closes the biggest ambiguity from the previous text-only fix. My remaining concern is schema clarity. After this patch, the summary still carries `total_files` alongside `changed_files_total` and `inspected_files`, even though `total_files` now means the same thing as `inspected_files`. I would prefer to collapse that before clients start depending on an ambiguous field set.

PR #573, opened 2026-04-21
fix(FEAT-CHECK-001): guard escaped coverage paths

### LINKED ISSUE #563: [Bug]: Coverage discovery should not panic on scanned files that escape the workspace root
### What area is affected?

CLI or runtime behavior

### What happened?

Coverage discovery still has multiple `strip_prefix(root).expect("scanned file should remain under the workspace root")` calls in `src/coverage.rs`.

That is safe only if every scanned path always stays under the workspace root. If a repository contains symlinked source/test entries or a future scanner starts yielding canonicalized paths from outside the root, `syu validate --require-symbol-trace-coverage` can panic instead of surfacing a regular validation issue.

Coverage discovery should treat out-of-root scanned files as a normal validation/runtime error and continue reporting cleanly, not panic with an `expect(...)` failure.

### syu version

main

### Environment

Linux, workspaces that include symlinked source trees or unusual scanner path handling

### Steps to reproduce

1. Create a workspace with `validate.require_symbol_trace_coverage: true`.
2. Introduce a scanned source/test path that resolves outside the workspace root (for example through a symlinked tree).
3. Run `syu validate .`.
4. Observe that the code path in `src/coverage.rs` uses `expect(...)` rather than a recoverable error.

Representative locations in `src/coverage.rs` include lines with:

```rust
.strip_prefix(root)
.expect("scanned file should remain under the workspace root")
```

This pattern currently appears in multiple language discovery passes.

### Before submitting

tohboeh5:
## Summary
- replace coverage discovery path panics with recoverable inventory issues
- keep each language scanner moving when one scanned file escapes the workspace root
- cover both the escaped-path guard and the continued scan behavior with focused coverage tests

- Closing keyword issue: Closes #563
- Requirement / feature IDs: FEAT-CHECK-001

## Validation

## Release notes
- [x] No release note needed


src/coverage.rs
```diff
@@ -1352,6 +1328,24 @@ fn should_skip_generated_directory(
         .is_some_and(|relative| path_matches_ignored_generated_directory(relative, ignored_paths))
```
tohboeh5, line 1408: This is better than panicking, but the callers still use `?`, so the first escaped path now aborts the entire coverage discovery pass for that language. That means one bad path can suppress unrelated coverage findings from the rest of the workspace. Can we surface the issue and keep scanning the remaining files instead of turning it into an early-return for the whole pass?

tohboeh5:
Overview:

Centralizing the workspace-root guard is the right fix, and replacing the `expect(...)` panics with a real `Issue` is a clear reliability improvement. My remaining concern is recovery scope. The new helper still bubbles an error out of each language discovery pass immediately, so one escaped file can stop coverage inventory for the rest of that language instead of reporting the bad path and continuing. I would prefer to keep the new issue reporting while making the scanner as non-fatal as possible for unrelated files.

PR #574, opened 2026-04-21
feat(FEAT-APP-001): show remote access warning in syu app

### LINKED ISSUE #564: [Feature]: Show an in-app remote-access warning when `syu app` runs with `--allow-remote`

`syu app` already requires an explicit `--allow-remote` opt-in for non-loopback binds, but once the browser UI is running there is no in-app warning that the workspace snapshot is being served to remote clients.

That leaves a gap between the one-time terminal warning and the long-lived browser session: people who open the UI later do not see that the app is in a remotely exposed mode.

### What change do you want?

Add an explicit in-app remote-access warning state when `syu app` is running with a non-loopback bind, for example:
- a banner in the header or workspace summary panel
- the effective bind/port rendered in the UI
- a short explanation that the current session is remotely reachable because `--allow-remote` was used intentionally

### Specification impact

This would extend the UX and safety story around `FEAT-APP-001` and the existing `--allow-remote` behavior.

### Alternatives considered

Keeping the warning in terminal output only is easy to miss once contributors switch to the browser tab or share the running server with another device.

### Before submitting

tohboeh5:
## Summary

Add app-server session metadata to the browser payload and show a persistent in-app warning when `syu app` is intentionally exposed beyond loopback. The UI now renders the effective bind/port and explains that the session is remotely reachable because `--allow-remote` was used.

- Closing keyword issue: `Closes #564`
- Requirement / feature IDs: `FEAT-APP-001`

## Validation

## Release notes

- [x] No release note needed


app/src/App.tsx
```diff
@@ -771,6 +771,27 @@ function App() {
...
+        {workspace.app_server.remotely_reachable && (
+          <div
+            role="alert"
+            className="md:col-span-2 rounded-3xl border border-amber-400/40 bg-amber-400/10 px-5 py-4 text-sm text-amber-50 shadow-2xl shadow-amber-950/10"
+          >
+            <p className="font-semibold">Remote access is enabled for this session.</p>
+            <p className="mt-2 leading-7 text-amber-100">
+              This browser UI is reachable at{" "}
+              <span className="font-mono text-amber-50">
+                http://{workspace.app_server.bind}:{workspace.app_server.port}
```
tohboeh5: When the bind address is IPv6, this renders a malformed URL like `http://::1:3000` because RFC 3986 requires IPv6 host literals to be wrapped in brackets. Since the whole point of this banner is to tell people exactly how the session is reachable, can we format IPv6 as `http://[::1]:3000` here (ideally via a small helper) before shipping this warning?

tohboeh5:
Overview:
This is a useful security/UX improvement overall: surfacing the fact that `--allow-remote` made the browser UI reachable beyond loopback is exactly the kind of explicitness this flow needs, and I like that the banner explains both the effective bind/port and the safer local-only alternative. The main problem I found is URL correctness for IPv6 binds. Right now the banner interpolates the raw host string, which produces an invalid URL when the server is bound to an IPv6 address. Once that display path brackets IPv6 hosts correctly, this will be much easier to trust.

PR #575, opened 2026-04-21
feat(FEAT-INIT-006, FEAT-INIT-007): share starter template help catalog

### LINKED ISSUE #565: [Feature]: Derive starter-template help text from the shared template catalog to prevent drift

Starter-template names are currently duplicated across multiple CLI and docs surfaces. The drift is already visible today: `syu init --help` hard-codes a shorter template list in `src/cli.rs`, while the enum-derived possible-values list shows additional supported templates.

As more starters land, this duplication makes it easy for help text, examples, and newcomer docs to fall out of sync.

### What change do you want?

Centralize starter-template display strings so CLI help and newcomer docs can derive from the same shared catalog that powers `syu templates`.

A useful first scope would:
- remove the hand-maintained template list from `InitArgs.template` help text
- expose one shared rendering helper for template names/examples
- add regression coverage so future starter additions cannot update the enum without updating newcomer-facing copy

### Specification impact

This would tighten the consistency story around `FEAT-INIT-006` and `FEAT-INIT-007`.

### Alternatives considered

Manually updating every help string and guide each time a template is added keeps working until it doesn’t; the current mismatch shows that it is already easy to miss one surface.

### Before submitting

tohboeh5:
## Summary

Derive `syu init` starter-template help text from the shared starter catalog so the CLI examples and newcomer-facing template list cannot drift from `syu templates`. This also updates the getting-started guide and repository-quality checks so Python and polyglot starters stay covered.

- Closing keyword issue: `Closes #565`
- Requirement / feature IDs: `FEAT-INIT-006`, `FEAT-INIT-007`

## Validation

## Release notes

- [x] No release note needed


tests/repository_quality.rs
```diff
@@ -14,6 +14,24 @@ fn read_json(path: &str) -> Value {
     serde_json::from_str(&read_file(path)).expect("repository JSON should parse")
 }
 
+fn assert_newcomer_template_examples(document: &str) {
+    for template in [
```
tohboeh5: This helper still hard-codes the starter names in a local array, so the repository-quality check can drift from the shared starter catalog even after this PR removes that drift from CLI help text. Could we derive the expected templates from the same shared catalog/helper instead of maintaining another manual list here?

tohboeh5:
Overview:
I like the direction here a lot: moving newcomer-facing help toward a shared starter catalog is the right way to keep `syu init --help`, `syu templates`, and the docs from drifting apart, and the added repository-quality coverage clearly tries to protect that contract. The remaining gap is that the new repository-quality helper reintroduces a hard-coded template list in test code, which undercuts the maintainability win of the production change. If that expectation is also derived from the shared catalog, this PR will feel much more complete.

Issue #578, opened 2026-04-21
[Bug]: Keep keyboard search navigation stable after pointer interactions in the browser app


tohboeh5:
### What area is affected?

CLI or runtime behavior

### What happened?

The browser app supports keyboard navigation through search results, but pointer interactions reset the focused-result bookkeeping in a way that makes ArrowUp/ArrowDown navigation lose continuity after hovering or clicking results.

That makes the search UI feel inconsistent for keyboard-first users who mix pointer and keyboard input while browsing spec items.

### What did you expect to happen?

Pointer interactions should not leave the search widget in a state where the next keyboard navigation step feels detached from the currently highlighted result.

### syu version

main

### Environment

Linux, local `syu app` session, browser search UI

### Steps to reproduce

1. Run `syu app .`.
2. Type in the search box and move through results with ArrowDown.
3. Hover or click one of the results with the mouse.
4. Continue with ArrowUp/ArrowDown and observe that the focused-result state was reset instead of preserved coherently.

### Logs, screenshots, or failing output

`app/src/App.tsx` maintains `focusedResultIndex`, but pointer handlers reset that state during result interaction rather than keeping keyboard navigation anchored on the current item.

### Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
Closed because current main preserves keyboard search focus after pointer interactions in app/src/App.tsx, and app/tests/browser-app.spec.ts covers the pointer/keyboard handoff.

PR #588, opened 2026-04-21
fix(FEAT-APP-001): keep keyboard search navigation stable after pointer input

tohboeh5:
## Summary\n- keep browser-app search focus anchored to item ids instead of transient list indexes\n- preserve keyboard navigation continuity after hover/click interactions in the result list\n- add probe coverage that keeps the shared app command gate at 100% after merging the latest app changes\n\nCloses #578\n


src/command/app.rs

scripts/ci/__pycache__/write-spec-coverage-summary.cpython-312.pyc

tohboeh5:
I reviewed the search-navigation change and the user-facing direction is good: anchoring keyboard movement to stable result IDs is the right UX improvement, and the added browser coverage exercises the pointer+keyboard path that motivated the bug. The main problems are around PR hygiene, though: this branch also includes an unrelated Rust test edit and a committed `__pycache__` artifact, and CI currently shows a failing `precommit` job. Please trim the unrelated files, regenerate a clean diff, and rerun the branch checks before this is ready.

PR #589, opened 2026-04-21
fix(FEAT-APP-001): abort stale browser refresh requests

### LINKED ISSUE #576: [Bug]: Abort in-flight browser refresh requests when the app UI unmounts
### What area is affected?

CLI or runtime behavior

### What happened?

The browser app polls `/api/app-data.json` and `/api/version`, but the fetches are launched without an `AbortController`. The cleanup path only flips a local `cancelled` flag, so in-flight requests keep running until the server responds.

That makes refresh behavior harder to reason about in long-lived tabs and can leave avoidable network work behind when contributors close or navigate away from the page mid-refresh.

When the app component unmounts or a newer refresh cycle replaces an older one, the browser should cancel the stale requests instead of only ignoring their result.

### syu version

main

### Environment

Linux, local `syu app` session, browser tab with the app open

### Steps to reproduce

1. Run `syu app .`.
2. Open the browser app and trigger a refresh while watching the Network tab.
3. Navigate away or close the tab before the requests complete.
4. Notice that `fetch('/api/app-data.json')` and `fetch('/api/version')` were not wired to an abort signal.

`app/src/App.tsx` starts both fetches in `Promise.all(...)`, but the effect cleanup only sets `cancelled = true` and removes the `hashchange` listener. There is no `AbortController` passed into the requests.

### Before submitting

tohboeh5:
## Summary
- abort stale app-data and version requests whenever a newer load supersedes them or the app unmounts
- ignore abort-shaped failures so refresh state only surfaces real errors to the user
- extend browser coverage around refresh recovery while keeping the shared coverage gate green

Closes #576


src/command/app.rs

scripts/ci/__pycache__/write-spec-coverage-summary.cpython-312.pyc

tohboeh5:
I reviewed the refresh-abort change and the core approach looks sensible: using `AbortController` for superseded loads is the right way to avoid stale in-flight work and the polling cleanup path is much easier to reason about afterward. What blocks approval is the extra unrelated diff that came along for the ride: the branch also changes `src/command/app.rs` and commits a generated `__pycache__` file, neither of which belongs to this focused browser-refresh fix. Please remove those unrelated hunks so the PR matches its stated scope.

tohboeh5 (commented):
I reviewed the refresh-abort flow and it reads as a solid fix: aborting superseded fetches and ignoring stale completions addresses the race directly, and the cleanup path now avoids leaving the app in a half-refreshed state. I didn't find blocking issues in the browser changes or the new test coverage.

PR #590, opened 2026-04-21
fix(FEAT-APP-001): announce browser refresh state changes

### LINKED ISSUE #577: [Bug]: Announce browser refresh-state changes with an aria-live region
### What area is affected?

CLI or runtime behavior

### What happened?

The browser app visibly updates its refresh badge between states like `current`, `refreshing`, and `stale`, but those changes are not exposed through an `aria-live` region.

That means screen-reader users do not get notified when the workspace snapshot becomes stale or when a refresh succeeds unless a harder failure triggers the separate alert banner.

The dynamic refresh status should be announced politely to assistive technology users so the browser UI stays understandable without relying on visual badge changes alone.

### syu version

main

### Environment

Linux, local `syu app` session, browser + screen reader or accessibility inspection tools

### Steps to reproduce

1. Run `syu app .`.
2. Open the browser UI and inspect the refresh status badge.
3. Trigger refresh transitions by waiting for polling or pressing "Refresh now".
4. Notice that only error banners use alert semantics; the normal status badge is not exposed through a live region.

`app/src/App.tsx` computes refresh-state labels and renders them in badges, but the rendered status elements do not use `aria-live="polite"` or a similar announcement strategy.

### Before submitting

tohboeh5:
## Summary
- add a polite live region that mirrors the app refresh state for assistive technology users
- announce stale, refreshing, and recovered snapshot states without relying on the visible badge alone
- cover the new accessibility behavior in Playwright while keeping the shared coverage gate green

Closes #577


src/command/app.rs

scripts/ci/__pycache__/write-spec-coverage-summary.cpython-312.pyc

tohboeh5:
I reviewed the live-region change and the accessibility intent is strong: adding a polite status surface for refresh transitions is the right UX direction, and the Playwright coverage gives the behavior a concrete regression test. The blocking issue is that the branch is not actually scoped to that behavior alone: it also carries an unrelated Rust test change plus a generated `__pycache__` artifact. Please clean those out so reviewers can evaluate the accessibility change without unrelated noise.

tohboeh5 (commented):
I reviewed the live-region refresh announcements and the accessibility direction is good: the new status surface is polite, stable, and covered by a focused Playwright test. I didn't find blocking issues in the implementation.

PR #593, opened 2026-04-21
fix(FEAT-APP-001): explain safe refresh failures safely

tohboeh5:
## Summary
- return structured refresh failure payloads from the app API with safe summaries and guidance
- surface that guidance in the browser refresh banner instead of exposing raw backend errors
- document the refreshed failure behavior and keep coverage green with targeted app command tests

## Testing
- bash scripts/ci/validate-app.sh --e2e
- cargo run -- validate .
- bash scripts/ci/coverage.sh summary


src/command/app.rs

scripts/ci/__pycache__/write-spec-coverage-summary.cpython-312.pyc

tohboeh5:
I reviewed the safe refresh-failure change and the product direction looks solid: returning bounded refresh categories and surfacing actionable guidance in the browser is a good UX/security balance, and the new tests cover the intended flow. The remaining issue is review scope: this branch also includes an unrelated Rust test change and a committed `__pycache__` artifact. Please strip those out so the PR only contains the refresh-error work it claims to implement.

Issue #600, opened 2026-04-30
[Bug]: some CIs not cached


tohboeh5:
### What area is affected?

Repository automation or CI

### What happened?

Some CIs like pre commit (not only this)  is not cached. 

### What did you expect to happen?

CIs use cache to make it run smoothly.

### syu version

0.0.1-alpha.9

### Environment

Github actions

### Steps to reproduce

Any time using CI.

### Logs, screenshots, or failing output

```shell

```

### Before submitting

- [x] I searched for existing issues first.
- [x] I can share a minimal reproduction if more detail is needed.

tohboeh5:
Closed because current CI already caches pre-commit hooks and other setup paths in .github/workflows/ci.yml and related workflows.

PR #606, opened 2026-04-30
fix(FEAT-QUALITY-001): cache pre-commit hooks

tohboeh5:
## Summary\n- cache the pre-commit hook environment in CI so repeated runs reuse downloaded hooks\n- document the cache step in the repository quality spec and generated docs\n- extend the repository quality test to cover the new cache step\n\nCloses #600

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,14 @@ jobs:
           cache: pip
           cache-dependency-path: .pre-commit-config.yaml
 
+      - name: Cache pre-commit hooks
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-${{ runner.os }}-${{ hashFiles('.pre-commit-config.yaml') }}
+          restore-keys: |
+            pre-commit-${{ runner.os }}-
+
       - name: Install pre-commit
         run: python -m pip install --upgrade pip pre-commit
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
           tool: cargo-llvm-cov
 
       - name: Set up Python with pip cache
+        id: setup-python
         uses: actions/setup-python@v6
         with:
           python-version: "3.x"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,14 @@ jobs:
           cache: pip
           cache-dependency-path: .pre-commit-config.yaml
 
+      - name: Cache pre-commit hooks
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-${{ runner.os }}-${{ hashFiles('.pre-commit-config.yaml') }}
```
tohboeh5: Consider including the resolved Python version in this cache key. `actions/setup-python` is pinned to `3.x`, so this broad key can restore hook envs built under a different interpreter and slowly erode the cache hit rate as the runner image advances.

tohboeh5 (commented):
Good change overall: caching pre-commit hooks is the right optimization here, and the CI/spec/test updates keep the behavior documented and enforced. The main concern is that the cache key is still a bit too coarse; I would tighten it to the resolved Python version so the cache does not silently accumulate mismatched hook envs as the runner image moves forward.

.github/workflows/ci.yml
```diff
@@ -52,12 +52,21 @@ jobs:
...
         with:
           python-version: "3.x"
           cache: pip
           cache-dependency-path: .pre-commit-config.yaml
 
+      - name: Cache pre-commit hooks
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('.pre-commit-config.yaml') }}
```
tohboeh5, line 66: The cache key is missing runner.arch, so arm64 and x64 runners can share the same pre-commit cache namespace. Those hook envs can include compiled wheels/native deps, which makes cross-arch restores risky. Please include the architecture in both the key and restore prefix.

tohboeh5 (commented):
Good optimization direction, but the cache namespace is too coarse as written: arm64 and x64 runners will share the same pre-commit cache, and those hook envs can contain platform-specific wheels/native deps. Please include runner.arch in the key and restore prefix so architectures do not restore each other's caches.

Issue #605, opened 2026-04-30
[Bug]: ignore Python bytecode caches under scripts/ci


tohboeh5:
What happened?\n\nPR 591 removed scripts/ci/__pycache__/write-spec-coverage-summary.cpython-312.pyc, but the repository does not ignore Python bytecode caches. Because .gitignore has no __pycache__/ or *.pyc entry, running the script will recreate the compiled file and make it easy to recommit by accident.\n\nWhat did you expect to happen?\n\nPython cache files under scripts/ci/ should stay untracked so cleanup of generated artifacts remains durable.\n\nEnvironment\n\nRepository automation / CI, Python 3.12 runtime used by scripts/ci/write-spec-coverage-summary.py.\n\nSteps to reproduce\n\n1. Run scripts/ci/write-spec-coverage-summary.py.\n2. Observe scripts/ci/__pycache__/write-spec-coverage-summary.cpython-312.pyc reappears.\n3. Run git status and see the bytecode file as untracked.

tohboeh5:
Closed because .gitignore already ignores __pycache__/ and *.py[cod] entries, so scripts/ci bytecode caches stay untracked.

PR #607, opened 2026-04-30
fix(ci): ignore Python bytecode caches

tohboeh5:
## Summary\n- ignore Python bytecode caches so scripts under scripts/ci do not leave untracked artifacts behind\n\nCloses #605

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .tooling
 .DS_Store
 __pycache__/
-*.pyc
+*.py[cod]
 # FEAT-CONTRIB-002: keep contributor-local helper worktrees out of routine status output
 /.worktrees/
 website/.docusaurus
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,6 +1,8 @@
 /target
 .tooling
 .DS_Store
+__pycache__/
+*.pyc
```
tohboeh5: This catches the current `.pyc` files, but if the goal is to ignore all Python bytecode artifacts, `*.py[cod]` is the more conventional pattern and also covers the optimized variants that other tooling can emit.

tohboeh5 (commented):
This is a clean, narrowly scoped fix for the bytecode noise, and it fits the repository well. I only have one small concern: the ignore rule is a little narrower than the usual Python bytecode glob, so it may leave room for the same problem to reappear under a different suffix.

Issue #602, opened 2026-04-30
[Feature]: Make starter templates and examples visible in the app and README chooser


tohboeh5:
### What problem are you trying to solve?

Templates and examples exist, but they are still too hidden for newcomers.

That makes the project feel heavier than it really is.

### What change do you want?

Add a starter gallery that surfaces `templates` and the checked-in `examples/` directly from the app and the top-level docs.

Each starter should say when to use it, what language shape it fits, and what the first validation win looks like.

### Specification impact

Docs and browse/app navigation only; no core validation semantics need to change.

### Alternatives considered

Keeping the current command card and example directories as-is.

### Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

tohboeh5:
Closed because the current app onboarding and the starter docs already surface templates and checked-in examples with usage guidance and summaries (see app/src/App.tsx, docs/guide/examples-and-templates.md, and related tests).

PR #608, opened 2026-04-30
feat(FEAT-APP-001): surface starter gallery in the browser app

tohboeh5:
Adds a starter gallery to the onboarding banner so the app surfaces templates and checked-in examples alongside the existing README chooser.\n\nCloses #602

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -34,6 +34,11 @@ type SearchResult = {
   kind: SectionKind;
 };
 
+type StarterGalleryEntry = {
+  name: string;
+  summary: string;
+};
+
 type SearchFilter = "all" | SectionKind;
 type RefreshState = "current" | "refreshing" | "stale";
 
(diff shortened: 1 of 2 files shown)


app/src/App.tsx
```diff
@@ -42,6 +47,24 @@ const SEARCH_RESULTS_LIST_ID = "spec-search-results-list";
 const SEARCH_FILTER_GROUP_ID = "spec-search-filter-group";
 const REFRESH_POLL_MIN_MS = 2_000;
 const REFRESH_POLL_MAX_MS = 10_000;
+const STARTER_TEMPLATES: StarterGalleryEntry[] = [
```
tohboeh5, line 50: This gallery hard-codes only part of the starter catalog and already omits generic, python-only, ruby-only, and java-only. That makes the browser onboarding disagree with syu templates and guarantees drift as the CLI catalog changes. Please source this from the shared starter catalog instead.

tohboeh5 (commented):
Nice visual addition, but this list is already diverging from the canonical starter catalog in src/command/init.rs. It omits generic, python-only, ruby-only, and java-only, so the browser onboarding will disagree with syu templates and drift as new starters are added. Could this render from the shared catalog instead of hard-coding a partial subset?

Issue #603, opened 2026-04-30
[Feature]: Promote audit and report as reviewer tools


tohboeh5:
### What problem are you trying to solve?

`audit` and `report` are valuable, but they do not read as first-class reviewer tools yet.

That makes the project look more CLI-heavy than workflow-heavy.

### What change do you want?

Add a reviewer-focused entry point that explains when to use `audit` versus `validate`.

Show how to turn findings into a shareable report, and how to move from report → fix → re-validate.

### Specification impact

Reviewer workflow docs, command card placement, and possibly one extra example report in the docs.

### Alternatives considered

Leaving reviewer guidance inside the existing validation docs

### Before submitting

- [x] I searched for existing issues and discussions first.
- [x] I can help validate the change against a real project story.

PR #609, opened 2026-04-30
docs(FEAT-DOCS-001): promote audit and report as reviewer tools

tohboeh5:
Adds a reviewer-focused entry point that explains when to use audit versus validate, and how to move from audit to report to re-validation.\n\nCloses #603

### DIFF
--- a/docs/guide/command-card.md
+++ b/docs/guide/command-card.md
@@ -50,6 +50,8 @@ syu relate FEAT-CHECK-001
 syu trace src/command/check.rs --symbol run_check_command
 syu log FEAT-CHECK-001 --kind implementation --path src/command
 syu validate . --id FEAT-CHECK-001
+syu audit .
+syu report .
 ```
 
 ### Share the current state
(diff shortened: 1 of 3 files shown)


docs/guide/command-card.md
```diff
@@ -20,6 +20,7 @@ If a pull request already exists, pair this page with the
 | Check the workspace | `syu validate .` | run the full graph, trace, and coverage validation pass |
 | Focus one validation view | `syu validate . --id FEAT-CHECK-001` | keep the visible output anchored on one requirement or feature after the normal validation run |
 | Focus trace failures first | `syu validate . --genre trace` | inspect trace-specific problems before reading the full validation output |
+| Audit a review target | `syu audit .` | scan for likely overlap, drift, and other review notes before you decide whether `validate` should fail |
```
tohboeh5, line 23: Adding `syu audit .` here is useful, but the quick **Reviewer loop** bundle later on this page still omits both `syu audit .` and `syu report .`. That leaves the fastest copy-paste path inconsistent with the new guidance, so readers who skim for the canned sequence will still miss the tools this PR is trying to promote. Could we update that bundle too?

tohboeh5:
Overall this is a strong documentation improvement: the new audit/report framing is clearer, the reviewer workflow reads more like an actual decision process, and the supporting repository-quality assertions keep the docs from silently drifting. The main gap I found is consistency inside the command card itself: the table now promotes `audit`, but the later reviewer-loop bundle still reflects the old flow. Tightening that last shortcut would make the guidance feel fully coherent.

Issue #599, opened 2026-04-30
[Bug]: macOS installer defaults to latest package track instead of the downloaded release tag


tohboeh5:
## Summary
When I install `syu` on macOS using a pinned release installer, the binary that gets installed is not necessarily the same release that I downloaded. The installer script defaults to `SYU_VERSION=latest`, so it resolves the newest package/release track available for the target rather than the tag that served the installer.

## What I observed
I followed the release install flow with a pinned release tag:

```bash
RELEASE=v0.0.1-alpha.9
curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/install-syu.sh" -o install-syu.sh
curl -fsSL "https://github.com/ugoite/syu/releases/download/${RELEASE}/checksums.sha256" -o checksums.sha256
sha256sum --ignore-missing -c checksums.sha256
bash install-syu.sh
```

On macOS, the resulting `syu` binary looked older than expected. The installed CLI only showed:

```text
Usage: syu <COMMAND>

Commands:
  validate  [aliases: check]
  report
  init
  help
```

## Expected behavior
If I download `install-syu.sh` and `checksums.sha256` from a specific release tag, running the installer should install that same release by default, or the docs should make it explicit that the installer always switches to a separate package track unless `SYU_VERSION` is set.

## Why this seems to happen
The installer defaults the version selector to `latest` in `scripts/install-syu.sh`:

- `scripts/install-syu.sh` defaults `SYU_VERSION` to `latest`
- the resolver then chooses the highest matching package/release tag for the current target
- for macOS, that can differ from the release tag used to download the installer script itself

The README also documents the pinned-release download flow and separately recommends `SYU_VERSION=alpha` for getting the latest alpha, which makes the current default behavior easy to misread.

Relevant references:
- `scripts/install-syu.sh`
- `README.md`

## Impact
This makes the install flow confusing on macOS because the verified installer script and the installed binary can come from different release tags. That is especially surprising during the alpha phase, where people expect the release they downloaded to be the one they get.

## Suggested fix
Either:

1. make the installer default to the release tag that shipped the script, or
2. update the docs to clearly state that the script downloads one release and then resolves a separate package track unless `SYU_VERSION` is set explicitly.

PR #610, opened 2026-04-30
fix(FEAT-INSTALL-001): pin installer default to the shipped release tag

tohboeh5:
Keeps the release installer aligned with the tag that shipped it by default, while still allowing SYU_VERSION to select another package track.\n\nCloses #599


scripts/ci/installer-smoke.sh
```diff
@@ -112,16 +128,20 @@ run_install_case() {
 
 main() {
   local target binary_name
+  local repo_version default_version
 
   trap cleanup EXIT
 
   target="$(resolve_target_triple)"
   binary_name="$(resolve_binary_name "$target")"
+  repo_version="$(awk -F'"' '/^version = "/ {print $2; exit}' "$repo_root/Cargo.toml")"
```
tohboeh5: This couples the smoke test to whatever version happens to be in Cargo.toml, but the mock registry tag set is still hard-coded. After the next release-please bump, the expected default tag will no longer exist in the fixture and this test will fail until both files are updated together. Please derive the expectation from the same source as the registry fixture.

tohboeh5 (commented):
The installer pinning itself is fine, but this smoke test now derives the expected default tag from Cargo.toml while the mock registry stays hard-coded. On the next release bump, that will point at a version the fixture does not serve, so CI will fail unless both places are updated together. Consider deriving the expected tag from the same fixture source or keeping it as an explicit constant.

PR #611, opened 2026-05-01
Fix template catalog coverage gap

tohboeh5:
Adds a small coverage-focused helper and tests in src/command/templates.rs so the template catalog branch stays covered and pushable.\n\nValidation: scripts/ci/quality-gates.sh, cargo run -- validate ., git push pre-push checks.


examples/generic/docs/syu/features/core/core.yaml
```diff
@@ -0,0 +1,11 @@
...
+version: 1
+
+features:
+  - id: FEAT-001
+    title: Bootstrap the generic spec workspace
+    summary: Provide a starter structure that contributors can extend.
+    status: planned
+    linked_requirements:
+      - REQ-001
+    implementations: {}
```
tohboeh5: Promoting `generic` to a first-class example feels premature while the example still has no implementation or test traces. The surrounding guide copy now points people at `examples/generic` to inspect a complete repository story with naming, links, **and traces**, but `syu validate examples/generic` currently reports `requirements=0/0` and `features=0/0` traces validated. Could we either add one minimal traced file/test here, or keep `generic` template-only until the checked-in example demonstrates an end-to-end trace?

tohboeh5:
I like the direction here: fixing the catalog coverage gap, adding the missing checked-in workspace, and wiring tests around `templates` and example validation all make the starter story easier to discover. The part that still feels weak is the substance of the new generic example itself. Right now it mirrors the bare scaffold but stops short of demonstrating traced code or tests, so it does not yet deliver the same inspection value as the other checked-in examples the docs compare against.

PR #612, opened 2026-05-01
Guard Python bytecode ignores

tohboeh5:
Adds a repo-quality assertion so scripts/ci Python bytecode caches stay ignored under .gitignore.\n\nValidation: cargo test repository_declares_contribution_workflow_assets -- --nocapture, bash scripts/ci/quality-gates.sh, git push pre-push checks.

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -1059,6 +1059,8 @@ fn repository_declares_contribution_workflow_assets() {
 
     assert!(gitignore.contains("FEAT-CONTRIB-002"));
     assert!(gitignore.contains("/.worktrees/"));
+    assert!(gitignore.contains("__pycache__/"));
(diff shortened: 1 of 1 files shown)


tohboeh5 (commented):
This change is small but useful: it closes a real gap in the repository-quality guardrail by making the Python bytecode ignore rules explicit in the test suite. I like that the assertions follow the existing style in `repository_declares_contribution_workflow_assets()`, so the new coverage is easy to understand and maintain alongside the other contributor-workflow expectations.

I did not find any blocking issues in the patch itself. The added checks match the current `.gitignore` entries, they validate the intended behavior directly, and the scope stays appropriately narrow without introducing extra test complexity. Overall this looks like a solid regression guard for a contributor-facing hygiene rule.

PR #613, opened 2026-05-01
Cache quality CI setup

tohboeh5:
Uses the shared Rust setup in the quality job so the Rust-heavy CI path reuses the same caches as the other workflow jobs.\n\nValidation: cargo test repository_declares_dependency_hygiene_and_ci_caching -- --nocapture, bash scripts/ci/quality-gates.sh, git push pre-push checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,23 +82,10 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Set up Rust
+      - name: Set up Rust for quality gates
         uses: ./.github/actions/setup-rust
         with:
           components: rustfmt,clippy
-
-      - name: Set up Node.js
-        uses: actions/setup-node@v6
-        with:
-          node-version: "25"
-          cache: npm
-          cache-dependency-path: app/package-lock.json
-
-      - name: Install browser app dependencies
-        shell: bash
-        run: |
-          scripts/ci/pinned-npm.sh install app
-          npm --prefix app ci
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -82,24 +82,11 @@ jobs:
...
-          cache: npm
-          cache-dependency-path: app/package-lock.json
-
-      - name: Install browser app dependencies
-        shell: bash
-        run: |
-          scripts/ci/pinned-npm.sh install app
-          npm --prefix app ci
-
       - name: Run quality gates
```
tohboeh5, line 96: Removing the Node setup/install steps breaks the very next `scripts/ci/quality-gates.sh` run in a fresh checkout. I reproduced this in a clean worktree: `cargo run -- validate .` now fails in `build.rs` with `browser app dependencies are not ready` because the embedded app build still expects `app/node_modules` to exist before any Cargo command runs. Could we keep the pinned npm + `npm --prefix app ci` setup here, or move that dependency preparation into the quality-gates script itself?

tohboeh5:
This change heads in the right direction by trying to simplify the quality job, but in its current form it removes a prerequisite that the Cargo-based quality gates still rely on. In a clean worktree, the first embedded browser-app build now fails because `app/node_modules` is never prepared before `scripts/ci/quality-gates.sh` runs. I like the intent to centralize setup, but this needs to preserve the app dependency bootstrap path before the quality job can be considered safe.

PR #614, opened 2026-05-01
docs(onboarding): make the first 10 minutes smaller

### LINKED ISSUE #604: [Feature]: Make the first 10 minutes feel smaller

The project is strong once you are inside it, but the entry tax is still too high.

A small team needs one obvious first win.

### What change do you want?

Bundle the bootstrap path into a single flow: doctor → init → first slice → validate → browse.

The flow should stop at the first useful success state instead of asking the user to learn every command first.

### Specification impact

Mostly onboarding docs and task ordering.

### Alternatives considered

Expecting users to discover the commands organically from README and guides.

### Before submitting

tohboeh5:
Closes #604\n\nTighten the first-run onboarding path around doctor, init, validate, and browse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ Pick the newcomer path that matches what you need next:
 
 - **Getting started**: choose this if you are new to `syu`, want a guided first
   run, and do **not** already know the four-layer model. Expect the most
-  hand-holding and roughly 10-15 minutes for the first workspace setup.
+  hand-holding and roughly 10-15 minutes for the first workspace setup, with a
+  guided `doctor → init → validate → browse` path.
 - **Quick start**: stay in this README when you want a compact, self-contained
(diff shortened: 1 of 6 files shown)


tohboeh5:
This is a focused onboarding improvement that makes the first-pass path much easier to follow. Adding `syu doctor .` to the top-level help, the command card, and the getting-started guide creates a more coherent newcomer story, and I like that the repository-quality assertions were updated alongside the docs so the guidance stays synchronized over time.

I also checked the behavior behind the new guidance and the command works as a sensible first step even in a fresh directory: it reports tool readiness and clearly explains that workspace-specific checks remain limited until `syu init` creates `syu.yaml`. Given that alignment between docs, CLI help, and actual command behavior, I did not find any blocking issues in this PR.

PR #615, opened 2026-05-01
fix(FEAT-APP-001): surface safe refresh failures in the browser UI

### LINKED ISSUE #579: [Feature]: Expose a safe in-app explanation when `/api/app-data.json` refreshes fail

When `syu app` refreshes fail, the browser app only sees a generic 500 response and the backend logs the real cause to stderr. That keeps internals hidden, which is good, but it also leaves browser-only users with almost no actionable explanation about what kind of workspace problem just happened.

This becomes more noticeable now that `syu app` can intentionally run in remotely reachable mode: people viewing the browser tab later may not have the original terminal nearby.

### What change do you want?

Add a safe, user-facing refresh-failure summary for the browser app, for example:
- a bounded error category or reason code returned by the API
- wording that distinguishes workspace validation/config errors from server lifecycle failures
- browser UI copy that points users toward the terminal only when the richer explanation cannot be surfaced safely

### Specification impact

This would extend the UX story around `FEAT-APP-001` without exposing raw internal stack traces.

### Alternatives considered

Keeping the API response fully generic protects internals, but it makes the browser UI feel opaque during recoverable refresh failures.

### Before submitting

tohboeh5:
Closes #579\n\nReturn bounded refresh failure details from /api/app-data.json and show them safely in the app UI.


tohboeh5:
This patch makes the browser refresh path much safer and easier to understand. The main improvement is the new end-to-end handling of refresh failures: the backend now returns bounded structured error payloads, the frontend validates that shape before using it, and the browser banner surfaces actionable guidance without leaking internal details. I also like that the change distinguishes workspace-specific rebuild failures from broader app-server failures instead of collapsing everything into one opaque 500 message.

The coverage is in a good place for the behavior change. The Rust-side response tests and the Playwright refresh-banner assertion line up with the intended UX, so the change feels well anchored both at the API boundary and in the browser flow. I did not find any blocking issues in the updated implementation.

PR #618, opened 2026-05-01
Add Kotlin trace adapter support

### LINKED ISSUE #583: [Feature]: Add first-class Kotlin trace adapter support for JVM-heavy repositories

The repository already documents Kotlin as an unsupported adapter, which means JVM-heavy repositories that are Kotlin-first still have to stop at layered spec ownership or fall back to examples that are not language-native to their stack.

That leaves a noticeable gap now that Java, Go, Ruby, TypeScript, Python, Rust, and C# workflows all have stronger stories.

### What change do you want?

Add first-class Kotlin trace adapter support, or at least open the path toward it, including:
- file discovery for Kotlin source and test files
- symbol extraction aligned with the current trace model
- documentation and examples showing how Kotlin traces should look
- regression coverage similar to the existing language-adapter tests

### Specification impact

This would extend `FEAT-TRACE-001` and the language-support story rather than changing the core CLI model.

### Alternatives considered

Relying on unsupported-language fallback guidance helps staged adoption, but it still leaves Kotlin-first repositories without native trace validation.

### Before submitting

tohboeh5:
Adds Kotlin trace adapter coverage and related validation updates.

- Closes #583

### DIFF
--- a/src/coverage.rs
+++ b/src/coverage.rs
@@ -1515,8 +1515,10 @@ fn is_java_test_file(path: &Path, contents: &str) -> bool {
 }
 
 fn is_kotlin_test_file(path: &Path, contents: &str) -> bool {
-    let junit_annotation_regex = Regex::new(r"@(?:[\w.]+\.)?(?:Test|RepeatedTest|ParameterizedTest|TestFactory|TestTemplate)\b")
-        .expect("Kotlin test annotation regex should compile");
+    let junit_annotation_regex = Regex::new(
+        r"@(?:[\w.]+\.)?(?:Test|RepeatedTest|ParameterizedTest|TestFactory|TestTemplate)\b",
+    )
+    .expect("Kotlin test annotation regex should compile");
     path.file_name()
         .and_then(|name| name.to_str())
         .is_some_and(|name| name.ends_with("Test.kt") || name.ends_with("Tests.kt"))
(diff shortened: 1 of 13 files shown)


src/coverage.rs
```diff
@@ -1055,6 +1145,47 @@ fn collect_java_public_symbols(contents: &str) -> Vec<String> {
...
+fn collect_kotlin_public_symbols(contents: &str) -> Vec<String> {
+    let type_regex = Regex::new(
+        r"(?m)^\s*(?:public\s+)?(?:abstract\s+|final\s+|sealed\s+|data\s+|enum\s+|annotation\s+)?(?:class|interface|object)\s+(?P<name>[A-Z][A-Za-z0-9_]*)\b",
+    )
+    .expect("Kotlin type regex should compile");
+    let function_regex = Regex::new(
+        r"(?m)^\s*(?:public\s+)?(?:suspend\s+)?fun\s+(?P<name>[A-Za-z_][A-Za-z0-9_]*)\b",
+    )
+    .expect("Kotlin function regex should compile");
+    let property_regex = Regex::new(
```
tohboeh5, line 1154: This regex will classify any unqualified `val`/`var` declaration as a public symbol, including locals inside function bodies. For example, `fun feature() { val temp = 1 }` matches here and will later trigger `SYU-coverage-public-001` for a symbol that is not part of the public trace surface at all. The Rust/Java/C# collectors stay much closer to actual externally visible declarations; Kotlin needs the same guard here or it will over-report ordinary implementation locals.

src/coverage.rs
```diff
@@ -1336,6 +1517,12 @@ fn is_java_test_file(path: &Path, contents: &str) -> bool {
         || contents.contains("extends TestCase")
 }
 
+fn is_kotlin_test_file(path: &Path) -> bool {
```
tohboeh5: Kotlin test/public classification is filename-based here, unlike the Java path which also inspects `@Test` annotations. That means a file such as `tests/TraceCoverage.kt` or `src/test/kotlin/TraceCoverage.kt` that contains JUnit 5 tests but does not end with `Test.kt`/`Tests.kt` will be routed through `collect_kotlin_public_symbols` and its test functions will be reported as untraced public implementation symbols. This will miss legitimate tests and produce the wrong coverage errors.

tohboeh5:
The overall direction here is good: adding Kotlin to both the trace adapter docs and the symbol-trace coverage inventory is the right slice of work, and the test coverage around the new paths is substantial. I did find two blocking correctness issues in the coverage discovery, though: the public-symbol regex currently over-reports ordinary local vals/vars, and the test/public split is too dependent on Test.kt/Tests.kt file names. I left inline comments on both spots.

PR #619, opened 2026-05-01
Fix installer default release tag

### LINKED ISSUE #599: [Bug]: macOS installer defaults to latest package track instead of the downloaded release tag

tohboeh5:
Pins the installer default to the release being shipped and updates the smoke test coverage.

Closes #599

### DIFF
--- a/scripts/install-syu.sh
+++ b/scripts/install-syu.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 DEFAULT_REPOSITORY="ugoite/syu"
 DEFAULT_PACKAGE_HOST="ghcr.io"
 DEFAULT_PACKAGE_SCHEME="https"
-DEFAULT_VERSION_SELECTOR="alpha"
+DEFAULT_RELEASE_TAG="__SYU_RELEASE_TAG__"
 tmp_dir=""
 
 cleanup_tmp_dir() {
@@ -72,7 +72,7 @@ resolve_package_repository() {
 }
 
 normalize_version_selector() {
-  local version="${SYU_VERSION:-$DEFAULT_VERSION_SELECTOR}"
+  local version="${SYU_VERSION:-$(resolve_default_release_tag)}"
 
   case "$version" in
     latest | alpha | beta | stable)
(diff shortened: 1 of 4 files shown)


scripts/install-syu.sh
```diff
@@ -71,7 +71,7 @@ resolve_package_repository() {
 }
 
 normalize_version_selector() {
-  local version="${SYU_VERSION:-latest}"
+  local version="${SYU_VERSION:-v0.0.1-alpha.8}"
```
tohboeh5: This hard-codes one specific alpha tag into the installer default, so the fix only works for the release that shipped this exact script. As soon as a newer pinned release ships, downloading that release's installer and running it without SYU_VERSION will still install v0.0.1-alpha.8, which is the same mismatch reported in #599. Can we derive the default selector from the release tag that shipped the script instead of baking a single version into the source?

tohboeh5:
The intent matches issue #599 and the installer smoke flow passed for me once I moved TMPDIR off the full /tmp mount. The remaining blocker is that the fix hard-codes one specific alpha tag into the installer default, so the next pinned-release installer will still drift away from the release that shipped it. I left an inline comment on the default selector for that.

PR #622, opened 2026-05-02
Prefer git tag for syu --version output

tohboeh5:
Use a git tag for  when the current commit is tagged, falling back to  version otherwise. Also add an optional build script flag  for local check-only builds where browser app dependencies are not installed.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -37,15 +37,15 @@ fn main() {
     fs::create_dir_all(&out_dir).expect("browser app dist directory should be creatable");
 
     let skip_browser_app_build = env::var_os("SYU_SKIP_BROWSER_APP_BUILD").is_some();
-    if !skip_browser_app_build {
-        if let Err(error) = required_npm_version(&app_dir).and_then(|required_npm| {
+    if !skip_browser_app_build
+        && let Err(error) = required_npm_version(&app_dir).and_then(|required_npm| {
             ensure_pinned_npm_ready(&manifest_dir, &app_dir)
                 .and_then(|_| ensure_app_dependencies(&app_dir, &required_npm))
                 .and_then(|_| rebuild_browser_wasm_bindings(&manifest_dir, &app_dir))
                 .and_then(|_| build_browser_bundle(&app_dir, &out_dir))
-        }) {
-            panic!("{error}");
-        }
+        })
+    {
+        panic!("{error}");
     }
 }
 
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -16,6 +16,12 @@ fn main() {
     let out_dir = PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR")).join("syu-app-dist");
 
     println!("cargo:rerun-if-changed=build.rs");
+    if manifest_dir.join(".git").exists() {
+        println!(
+            "cargo:rerun-if-changed={}",
```
tohboeh5: Watching `.git` itself is not sufficient in worktrees. In a worktree checkout, `.git` is just a tiny text file that points at the real gitdir, and its contents do not change when `HEAD` moves or when a tag is added/updated. That means Cargo can keep reusing a stale `SYU_GIT_VERSION` after branch/tag changes in exactly the workflow this PR is trying to fix. We need to watch the resolved gitdir/HEAD and refs, not just `manifest_dir/.git`.

build.rs
```diff
@@ -31,13 +37,18 @@ fn main() {
...
-    if let Err(error) = required_npm_version(&app_dir).and_then(|required_npm| {
-        ensure_pinned_npm_ready(&manifest_dir, &app_dir)
-            .and_then(|_| ensure_app_dependencies(&app_dir, &required_npm))
-            .and_then(|_| rebuild_browser_wasm_bindings(&manifest_dir, &app_dir))
-            .and_then(|_| build_browser_bundle(&app_dir, &out_dir))
-    }) {
-        panic!("{error}");
+    emit_build_version();
+
+    let skip_browser_app_build = env::var_os("SYU_SKIP_BROWSER_APP_BUILD").is_some();
```
tohboeh5, line 39: `SYU_SKIP_BROWSER_APP_BUILD` currently skips the bundle step but it does not create the `$OUT_DIR/syu-app-dist` directory that `include_dir!("$OUT_DIR/syu-app-dist")` requires later in the crate. I reproduced this with `SYU_SKIP_BROWSER_APP_BUILD=1 cargo run -- --version` in this branch: the build fails with `".../out/syu-app-dist" is not a directory`. So the new escape hatch is not usable yet for the lightweight CLI/version check that motivated it.

tohboeh5:
Preferring the Git tag for --version makes sense, and separating the version stamp from the browser-app build is the right direction. I did hit two blocking problems in the current implementation: watching manifest_dir/.git does not invalidate correctly in worktree checkouts, and SYU_SKIP_BROWSER_APP_BUILD currently leaves include_dir! without the directory it needs, so the lightweight version check still fails to compile. I left inline comments with the reproduction details.

PR #633, opened 2026-05-06
Migrate short-running CI jobs to ubuntu-slim

tohboeh5:
Closes #632\n\nMoved the short-running GitHub Actions jobs without runner-specific requirements from ubuntu-latest to ubuntu-slim.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,7 @@ jobs:
 
   quality:
     if: github.event_name != 'schedule'
+    # Keep this on the larger runner; it exceeds ubuntu-slim's time budget.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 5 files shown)


tohboeh5 (commented):
Thanks for narrowing this down after the first pass. Keeping `actionlint` and `docs-site` on `ubuntu-latest` was a good correction, and the remaining short jobs that already passed on `ubuntu-slim` look like sensible candidates.

The blocker is that `quality` and `coverage` are still moved even though this PR's own run shows both jobs being cancelled after roughly 15 minutes on `ubuntu-slim`, which matches GitHub's hard timeout for the single-CPU runner. That means this change would make two core CI gates unreliable instead of cheaper. I don't see other issues in the remaining migrations, but those two jobs need to stay on `ubuntu-latest` (or another larger runner) before this is safe to merge.

PR #660, opened 2026-05-06
[Feature]: Add a dry-run planning mode for syu validate --fix

tohboeh5:
Closes #656\n\nAdds dry-run planning for validate --fix, including machine-readable and text plan output, docs, and tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,8 +708,16 @@ The `syu` repository itself enables `validate.require_non_orphaned_items`,
 
 ## Safe autofix
 
-`syu validate --fix` is intentionally conservative. Today it only repairs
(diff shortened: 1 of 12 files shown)


tohboeh5 (commented):
Nice addition overall: the text/JSON plan output makes `validate --fix` much safer to inspect before writing, and the new integration coverage exercises the happy paths well. The main problem I hit is that `--dry-run` currently degrades into a plain `validate` run unless fixes are already enabled through `--fix` or config, which is a confusing CLI contract for a newly introduced flag. I left a concrete finding with the relevant file/line so that behavior can be tightened up.

PR #661, opened 2026-05-06
[Feature]: Add transactional autofix application and rollback safety

### LINKED ISSUE #657: [Feature]: Add transactional autofix application and rollback safety

Broader autofix support raises the cost of partial writes or half-applied multi-file edits. If `syu validate --fix` starts touching more files and rule types, it needs stronger safety guarantees than best-effort sequential edits.

## What change do you want?

Add transactional autofix application and rollback safety for multi-file fixes.

Acceptance criteria:
- `syu validate --fix` can stage a multi-file fix set and avoid leaving the workspace half-updated on failure
- error reporting makes it clear whether changes were applied, rolled back, or never started
- the safety model works for both mapping-based and ownership-manifest-based workflows
- docs explain how rollback behavior affects normal review and recovery steps

## Specification impact

This likely needs validation docs, troubleshooting updates, and tests that cover partial-failure cases for broader autofix support.

## Alternatives considered

Continue applying fixes one file at a time without rollback. That is simpler, but it becomes riskier as the fix surface expands.

## Before submitting

tohboeh5:
Closes #657\n\nAdds rollback-safe autofix application, explicit failure status reporting, and coverage for partial-write recovery.

### DIFF
--- a/README.md
+++ b/README.md
@@ -715,7 +715,9 @@ documentation-style trace gaps that can be updated mechanically:
 - missing `doc_contains` snippets for Rust, Python, Go, and TypeScript symbols
 
 It does **not** attempt speculative edits like renaming symbols or inventing
-missing files.
+missing files. If a later write fails, `syu` rolls back earlier edits and tells
+you whether the run never started or was rolled back after partial changes, so a
+failed autofix should leave the workspace ready to review again.
 
 ## Example workspaces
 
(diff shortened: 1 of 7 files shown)


tohboeh5 (commented):
The transactional direction here is good: wrapping autofix writes behind a rollback-aware helper is exactly the right shape, and the new failure-path coverage makes the behavior much easier to reason about. The gap I found is in rollback itself: cleanup of newly created files currently ignores delete failures, so the command can claim success while still leaving partial changes behind. I left that as a concrete finding because it cuts directly against the safety guarantee this PR is adding.

PR #662, opened 2026-05-06
Add graph registry autofix

tohboeh5:
## Summary
- dedupe graph links and restore missing reciprocals during syu validate --fix
- resync the feature registry from discovered feature documents
- add regression coverage for reciprocal repair and registry-kind fallback

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

FEAT-CHECK-001

### DIFF
--- a/README.md
+++ b/README.md
@@ -713,6 +713,9 @@ documentation-style trace gaps that can be updated mechanically:
 
 - missing requirement / feature IDs in symbol documentation
 - missing `doc_contains` snippets for Rust, Python, Go, and TypeScript symbols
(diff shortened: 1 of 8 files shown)


tohboeh5 (commented):
I like the overall direction here: deduping reciprocal graph links and resyncing the feature registry should remove a lot of repetitive manual cleanup, and the regression test around graph drift is a solid start. The main hole I found is that the new registry autofix only handles drift once `features.yaml` already exists; a missing registry still aborts the run before autofix gets a chance to repair anything. I left that as a concrete finding because it feels like an important part of the "registry autofix" promise.

PR #663, opened 2026-05-06
[Feature]: Expand syu validate --fix to repair safe trace and ownership hygiene

tohboeh5:
Implements the trace and ownership hygiene autofix expansion for syu validate --fix.\n\nValidation: bash scripts/ci/quality-gates.sh && cargo run -- validate .

### DIFF
--- a/src/command/check.rs
+++ b/src/command/check.rs
@@ -1166,19 +1166,29 @@ fn sync_feature_registry(
     documents: &[MutableLoadedDocument<FeatureDocument>],
 ) -> Result<Option<FeatureRegistryDocument>> {
     let registry_path = feature_root.join("features.yaml");
-    let raw = fs::read_to_string(&registry_path)?;
-    let registry: FeatureRegistryDocument = serde_yaml::from_str(&raw)?;
     let files = feature_registry_entries_for_documents(feature_root, documents);
+    match fs::read_to_string(&registry_path) {
+        Ok(raw) => {
+            let registry: FeatureRegistryDocument = serde_yaml::from_str(&raw)?;
+            if feature_registry_entries_match(&registry.files, &files) {
+                return Ok(None);
+            }
 
-    if feature_registry_entries_match(&registry.files, &files) {
-        return Ok(None);
+            Ok(Some(FeatureRegistryDocument {
+                version: registry.version,
+                updated: registry.updated,
+                files,
+            }))
+        }
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => Ok(Some(
+            FeatureRegistryDocument {
+                version: env!("CARGO_PKG_VERSION").to_string(),
+                updated: None,
+                files,
+            },
+        )),
+        Err(error) => Err(error.into()),
     }
-
-    Ok(Some(FeatureRegistryDocument {
-        version: registry.version,
-        updated: registry.updated,
-        files,
-    }))
 }
 
 fn write_modified_documents<T: serde::Serialize>(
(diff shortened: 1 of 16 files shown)


src/command/check.rs
```diff
@@ -706,9 +731,103 @@ fn apply_autofix(workspace: &Workspace) -> Result<AutofixSummary> {
         }
     }
 
+    apply_graph_autofix(workspace, &mut summary)?;
```
tohboeh5: `apply_autofix_for_trace_map()` writes trace files eagerly before `apply_graph_autofix()` runs, so a later graph write failure leaves the workspace partially modified even though the troubleshooting docs now promise rollback. I reproduced this by making `docs/syu/features/core.yaml` read-only: `validate --fix` returned an error, but `src/lib.rs` had already gained the inserted `REQ-1` doc line. We need to stage these edits and commit them only after every autofix step succeeds, or explicitly roll back earlier writes on error.

src/command/check.rs
```diff
@@ -706,9 +731,103 @@ fn apply_autofix(workspace: &Workspace) -> Result<AutofixSummary> {
...
+        &requirement_index,
+    );
+    apply_requirement_reciprocals(&requirement_docs, &mut feature_docs, &feature_index);
+    apply_feature_reciprocals(&feature_docs, &mut requirement_docs, &requirement_index);
+
+    if feature_root.is_dir()
+        && let Some(updated_registry) = sync_feature_registry(&feature_root, &feature_docs)?
+    {
+        let registry_path = feature_root.join("features.yaml");
+        fs::write(&registry_path, serde_yaml::to_string(&updated_registry)?)?;
```
tohboeh5: This rewrites `features.yaml` through `serde_yaml::to_string(...)`, which drops existing comments. That is a real regression for this repo because `docs/syu/features/features.yaml` currently carries the release-control markers (`# x-release-please-start-version` / `# x-release-please-end`) at the top of the file. Any registry sync would delete those lines, so this needs a preservation strategy instead of a full YAML reserialize.

src/command/check.rs
```diff
@@ -730,6 +849,378 @@ fn apply_autofix_for_trace_map(
...
+fn write_modified_documents<T: serde::Serialize>(
+    documents: &[MutableLoadedDocument<T>],
+    summary: &mut AutofixSummary,
+    root: &Path,
+) -> Result<()> {
+    for document in documents {
+        if !document.changed {
+            continue;
+        }
+        fs::write(&document.path, serde_yaml::to_string(&document.document)?)?;
```
tohboeh5: The same full-document YAML round-trip is destructive for ordinary spec files too: it strips user comments / formatting, and it also injects unrelated noise like `language: null` because the serialized structs do not skip missing optional fields. In a minimal repro, a policy file containing `# keep this comment` lost that comment and gained `language: null` after `validate --fix` only added a reciprocal link. `--fix` should preserve the original document shape here instead of rewriting the entire file.

src/command/check.rs
```diff
@@ -730,6 +849,378 @@ fn apply_autofix_for_trace_map(
...
+        .iter()
+        .map(|document| {
+            let relative = normalize_relative_path(
+                document
+                    .path
+                    .strip_prefix(feature_root)
+                    .unwrap_or(&document.path),
+            );
+            FeatureRegistryEntry {
+                kind: feature_registry_kind(feature_root, &document.path),
```
tohboeh5, line 1298: This recomputes `kind` from the path, which mutates registry metadata that users may have chosen intentionally. The docs for `syu add feature --kind ...` treat `kind` as explicit input, but with this code an entry like `kind: adoption, file: auth/login.yaml` is silently rewritten to `kind: auth` on the next `validate --fix`. Registry sync should preserve the existing `kind` for already-listed files and only fill it in for brand-new entries.

tohboeh5 (commented):
Thanks for expanding `validate --fix` into the graph and registry hygiene cases — the added coverage around duplicate links, reciprocal links, and missing registries is directionally good, and the docs make the intended scope much clearer. I did find a few blocking issues in the current implementation, though: the new autofix path can leave partial writes behind on failure, both registry and document updates currently use lossy YAML round-trips, and registry sync mutates explicit `kind` metadata. Those cases make `--fix` less conservative than the docs promise, so I don't think this is safe to merge yet.

PR #665, opened 2026-05-06
[Feature]: Bring C# trace support up to doc_contains and autofix parity

tohboeh5:
Closes #653

### DIFF
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -274,8 +274,15 @@ fn fix_language_symbol_docs(
     let indent = line_indentation(&lines[existing_line - 1]);
     let line_refs = lines.iter().map(String::as_str).collect::<Vec<_>>();
     if let Some(block) = find_doc_block(&line_refs, existing_line, style) {
-        let inserts = render_missing_doc_lines(&indent, &missing, style);
-        lines.splice(block.end_line - 1..block.end_line - 1, inserts);
+        let start = block.start_line - 1;
+        let end = block.end_line - 1;
+        if matches!(style, DocCommentStyle::Block { .. }) && block.start_line == block.end_line {
+            let replacement = render_rebuilt_block(&indent, &block.text, &missing, style);
+            lines.splice(start..=end, replacement);
+        } else {
+            let inserts = render_missing_doc_lines(&indent, &missing, style);
+            lines.splice(end..end, inserts);
+        }
     } else {
         let inserts = render_new_doc_block(&indent, &missing, style);
         let insertion_line = find_doc_insertion_line(&lines, existing_line);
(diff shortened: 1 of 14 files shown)


tohboeh5 (commented):
Thanks for pushing this forward. Generalizing the doc-inspection/autofix path across Java, C#, and Kotlin is a good direction, and I like that the PR updates the user-facing docs and adds targeted tests instead of only changing the matcher logic.

I found two blocking correctness issues in `src/inspect.rs` during local repros:
1. Java declaration lookup still matches call sites before declarations, so `syu validate --fix` can insert the generated Javadoc into a method body instead of onto the target symbol.
2. Updating an existing one-line Javadoc/KDoc inserts the new lines outside the block comment, which leaves the file broken and the validation still failing.

Because both issues affect the advertised `doc_contains`/autofix support directly, I don't think this is ready to merge yet.

src/inspect.rs
```diff
@@ -185,10 +213,265 @@ pub fn apply_symbol_doc_fix(
...
+    symbol: &str,
+) -> Option<usize> {
+    let patterns = declaration_patterns(adapter, symbol);
+    find_line_by_patterns(contents, &patterns)
+}
+
+fn declaration_patterns(adapter: &dyn LanguageAdapter, symbol: &str) -> Vec<String> {
+    let mut patterns = adapter.patterns(symbol);
+    if patterns.len() > 1 {
+        patterns.pop();
```
tohboeh5, line 311: Blocking: `declaration_patterns()` drops the generic fallback, but the remaining Java method regex is still just `name\s*\(`. In a file where `trace_symbol()` is called before the declaration, `find_line_by_patterns()` resolves the call site first, and `syu validate --fix` inserts the generated Javadoc into the method body instead of onto the declaration. I reproduced this locally with `demo() { trace_symbol(); }` before `public void trace_symbol() {}`. Please make the Java matcher declaration-aware before advertising this support.

src/inspect.rs
```diff
@@ -185,10 +213,265 @@ pub fn apply_symbol_doc_fix(
...
+    if missing.is_empty() {
+        return Ok(None);
+    }
+
+    let mut lines = to_lines(contents);
+    let indent = line_indentation(&lines[existing_line - 1]);
+    let line_refs = lines.iter().map(String::as_str).collect::<Vec<_>>();
+    if let Some(block) = find_doc_block(&line_refs, existing_line, style) {
+        let inserts = render_missing_doc_lines(&indent, &missing, style);
+        lines.splice(block.end_line - 1..block.end_line - 1, inserts);
```
tohboeh5: Blocking: this splice point breaks one-line block docs such as `/** requirement doc line */`. `find_block_doc_block()` only tracks `end_line`, so `lines.splice(block.end_line - 1..block.end_line - 1, inserts)` inserts the new snippet *before* the whole comment line, yielding invalid output like `* feature doc line` above the `/** ... */`. I reproduced this with `syu validate --fix` on a one-line Javadoc fixture. Please insert inside the existing block (or normalize one-line blocks first) before merging.

PR #667, opened 2026-05-07
Issue 658: Graph registry autofix

tohboeh5:
Adds graph/registry autofix behavior and the associated regression coverage.

### DIFF
--- a/src/command/check.rs
+++ b/src/command/check.rs
@@ -82,6 +82,26 @@ struct MutableLoadedDocument<T> {
     changed: bool,
 }
 
+#[derive(Debug, Default, Clone, Serialize)]
+struct AutofixPlan {
+    planned_updates: usize,
+    updated_files: BTreeSet<PathBuf>,
+    changes: Vec<AutofixPlanChange>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+struct AutofixPlanChange {
+    path: PathBuf,
+    rules: Vec<&'static str>,
+    summary: String,
+}
+
+#[derive(Debug, Default)]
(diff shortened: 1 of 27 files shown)


src/command/check.rs
```diff
@@ -814,6 +839,10 @@ fn run_autofix(workspace: &Workspace, mode: AutofixMode) -> Result<AutofixRun> {
             )?;
         }
 
+        if mode == AutofixMode::Apply {
```
tohboeh5: Blocking: this new graph autofix path never runs in dry-run mode. `run_autofix()` only calls `apply_graph_autofix()` behind `if mode == AutofixMode::Apply`, so `syu validate --fix --dry-run` will not show planned removals of duplicate links, reciprocal-link repairs, or feature-registry syncs. That makes the new fixes invisible until after they mutate the workspace, which breaks the contract documented for dry-run planning. Please thread the autofix mode/plan through this path and add a dry-run regression test for the graph cases.

tohboeh5 (commented):
Thanks for pushing this forward. The transactional rollback work is a solid improvement, and broadening `--fix` to cover duplicate graph links, reciprocal repairs, and registry drift is the right direction for the UX.

The blocker is that the new graph autofix path is apply-only right now, so `syu validate --fix --dry-run` still hides all of these planned graph/registry edits. That means users cannot inspect the full set of changes before mutating their workspace, which is a pretty important regression for a command that is explicitly positioned as conservative and reviewable. I left one inline comment on the exact branch point that needs to carry plan-mode handling through this new path. Aside from that, the overall structure looks reasonable.

PR #669, opened 2026-05-07
Issue 655: Ruby trace parity (REQ-CORE-01)

tohboeh5:
Adds Ruby trace inspection and coverage parity, plus regression tests and docs sync.

### DIFF
--- a/src/coverage.rs
+++ b/src/coverage.rs
@@ -1874,6 +1874,8 @@ mod tests {
     };
 
     #[cfg(unix)]
+    use std::os::unix::fs::PermissionsExt;
+    #[cfg(unix)]
     use std::os::unix::fs::symlink;
 
     use tempfile::tempdir;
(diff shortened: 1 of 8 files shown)


src/coverage.rs
```diff
@@ -836,6 +849,74 @@ fn collect_go_test_symbols(contents: &str) -> Vec<String> {
         .collect()
 }
 
+fn collect_ruby_public_symbols(contents: &str) -> Vec<String> {
+    let class_module_regex =
+        Regex::new(r"(?m)^\s*(?:class|module)\s+(?P<name>[A-Z][A-Za-z0-9_:]*)\b")
+            .expect("Ruby class regex should compile");
+    let method_regex =
+        Regex::new(r"(?m)^\s*def\s+(?:self\.)?(?P<name>[A-Za-z_][A-Za-z0-9_!?=]*)\b")
```
tohboeh5, line 857: This regex inventories every `def` as a public symbol, so methods that appear after `private` / `protected` will become mandatory ownership targets too. In a normal Ruby codebase that turns internal helpers into false-positive `require_symbol_trace_coverage` failures. Could we track visibility state while scanning, or otherwise skip methods declared in non-public sections?

src/coverage.rs
```diff
@@ -836,6 +849,74 @@ fn collect_go_test_symbols(contents: &str) -> Vec<String> {
...
+            .expect("Ruby method regex should compile");
+    let constant_regex = Regex::new(r"(?m)^\s*(?P<name>[A-Z][A-Za-z0-9_:]*)\s*=")
```
tohboeh5: The new Ruby inventory now pulls in every constant assignment, but the updated docs and example guidance still describe strict Ruby coverage in terms of classes/modules/public methods. That means ordinary constants like `VERSION` will start failing validation in `lib/` / `src/` with no corresponding trace model. I think we should either drop constants from the inventory or make constants an explicit supported target in the spec/docs/tests before enforcing them.

tohboeh5 (commented):
Thanks for pushing Ruby strict-coverage parity forward. The overall direction makes sense and the new docs/tests make the feature easier to discover. I do think there are two correctness gaps in the symbol inventory, though: it currently treats non-public helper methods as public API, and it starts enforcing constants even though the Ruby coverage contract in this repo still describes classes/modules/public methods only.

PR #670, opened 2026-05-08
feat: track remaining version-managed manifests

tohboeh5:
Closes #620\n\nAdds release-please tracking for the remaining version-pinned manifests and example/fixture YAML files, and extends repository quality coverage to keep them listed.

### DIFF
--- a/tests/repository_quality.rs
+++ b/tests/repository_quality.rs
@@ -310,6 +310,19 @@ fn repository_declares_release_automation() {
     assert!(!release_config.contains("\"skip-changelog\": true"));
     assert!(release_config.contains("\"changelog-type\": \"github\""));
     assert!(!release_config.contains("\"initial-version\""));
+    assert!(release_config.contains("\"file\": \"app/package.json\""));
+    assert!(release_config.contains("\"file\": \"website/package.json\""));
+    assert!(release_config.contains("\"file\": \"editors/vscode/package.json\""));
+    assert!(release_config.contains("\"file\": \"crates/syu-core/Cargo.toml\""));
+    assert!(release_config.contains("\"file\": \"app/wasm/Cargo.toml\""));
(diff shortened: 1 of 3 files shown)


tests/repository_quality.rs
```diff
@@ -310,6 +310,19 @@ fn repository_declares_release_automation() {
     assert!(!release_config.contains("\"skip-changelog\": true"));
     assert!(release_config.contains("\"changelog-type\": \"github\""));
     assert!(!release_config.contains("\"initial-version\""));
+    assert!(release_config.contains("\"file\": \"app/package.json\""));
+    assert!(release_config.contains("\"file\": \"website/package.json\""));
+    assert!(release_config.contains("\"file\": \"editors/vscode/package.json\""));
+    assert!(release_config.contains("\"file\": \"crates/syu-core/Cargo.toml\""));
+    assert!(release_config.contains("\"file\": \"app/wasm/Cargo.toml\""));
+    assert!(release_config.contains("\"file\": \"examples/rust-only/syu.yaml\""));
```
tohboeh5, line 318: These assertions only cover a subset of the new `extra-files` entries. The added example manifests from `browser-ui`, `csharp-fallback`, `generic`, `team-scale`, `go-only`, `python-only`, `ruby-only`, and `docs-first` are still unasserted here, so this guard will pass even if one of those release-please entries disappears later. Can we assert the full list (or derive the expected set from the config) so the test protects the whole change?

tohboeh5 (commented):
Nice expansion of the release automation surface: this closes a real gap by bringing more version-stamped manifests under release-please. The remaining concern is test coverage. `tests/repository_quality.rs` now checks some of the new entries, but it still leaves part of the expanded manifest list unguarded, so a future regression in those paths would slip through.

PR #672, opened 2026-05-08
feat: define request artifact format

### LINKED ISSUE #646: [Feature]: Define a request artifact format for syu task workflows

If `syu` is going to classify or scope incoming requests, it needs a stable input shape. Free-form text is useful, but without a small request artifact format it is hard to build repeatable task commands or app flows around the same request.

## What change do you want?

Define a lightweight request artifact format for `syu task` workflows.

Acceptance criteria:
- `syu` documents a small request input format such as Markdown frontmatter, YAML, or another explicit artifact shape
- the format can capture the request text plus optional context like affected area or repository constraints
- the format is small enough for CLI use and predictable enough for later app support
- docs explain when to use the request artifact instead of starting directly from raw spec edits

## Specification impact

This likely needs a new task-planning feature or requirement plus documentation about the request artifact contract.

## Alternatives considered

Accept only raw terminal text. That may be fine for experiments, but it makes repeatable workflows and app support much harder.

## Before submitting

tohboeh5:
Closes #646\n\nAdds a compact request-artifact guide for syu task workflows, plus docs links from the existing getting-started and reviewer guides.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -94,6 +94,196 @@
               "replace": "version: \"{{version}}\""
             }
           ]
+        },
+        {
+          "type": "generic",
+          "file": "app/package.json",
+          "replacers": [
+            {
+              "search": "\"version\": \"(?<version>.*)\"",
+              "replace": "\"version\": \"{{version}}\""
+            }
+          ]
+        },
+        {
+          "type": "generic",
+          "file": "website/package.json",
+          "replacers": [
+            {
+              "search": "\"version\": \"(?<version>.*)\"",
+              "replace": "\"version\": \"{{version}}\""
+            }
+          ]
+        },
+        {
+          "type": "generic",
+          "file": "editors/vscode/package.json",
+          "replacers": [
+            {
+              "search": "\"version\": \"(?<version>.*)\"",
+              "replace": "\"version\": \"{{version}}\""
+            }
+          ]
+        },
+        {
+          "type": "generic",
+          "file": "crates/syu-core/Cargo.toml",
+          "replacers": [
+            {
+              "search": "name = \"syu-core\"\nversion = \"(?<version>.*)\"",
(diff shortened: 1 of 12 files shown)


src/command/check.rs
```diff
@@ -839,7 +839,9 @@ fn run_autofix(workspace: &Workspace, mode: AutofixMode) -> Result<AutofixRun> {
             )?;
         }
 
-        apply_graph_autofix(workspace, &mut transaction, &mut run, mode)?;
+        if mode == AutofixMode::Apply {
```
tohboeh5: `apply_graph_autofix` now runs only in `AutofixMode::Apply`, so `syu validate --fix --dry-run` no longer plans or reports graph/registry fixes at all. That is a user-visible regression from the previous behavior and breaks the dry-run contract for graph autofix. Please keep the planning path for dry-run, or reintroduce an explicit plan-only branch here.

release-please-config.json
```diff
@@ -94,196 +94,6 @@
               "replace": "version: \"{{version}}\""
             }
           ]
-        },
```
tohboeh5, line 97: This trims the release-please config down to the root package only, but the repo still has versioned artifacts that used to be synchronized here (`app/package.json`, `website/package.json`, `editors/vscode/package.json`, and the example `syu.yaml` files). If this stays as-is, future releases will stop updating those versions and they will drift immediately. Please either keep those extra-files entries or document the new versioning strategy before removing them.

tohboeh5 (commented):
The request-artifact docs read more clearly now, and the surrounding guide links make the intake flow easier to discover. The main concern is that this PR also changes behavior outside the docs path: `validate --fix --dry-run` no longer reports graph/registry fixes, and the release automation config drops version sync for several still-versioned artifacts. Both are user-visible regressions that are easy to miss in a docs-focused PR, so I would like to see those either restored or explicitly justified before merging.

PR #673, opened 2026-05-10
[Feature]: Add OpenAPI operation selectors to feature traces

tohboeh5:
Closes #634

Summary:
- add OpenAPI trace selectors with `file + method + path` validation
- accept JSON and YAML OpenAPI documents
- update validation docs and guides

Validation:
- `cargo test -q`
- `scripts/ci/coverage.sh summary`
- `scripts/ci/quality-gates.sh`
- CLI smoke test with a temporary OpenAPI workspace


src/command/check.rs
```diff
@@ -3461,7 +3503,227 @@ fn verify_trace_reference(
...
+                Some(format!(
+                    "Fix `{}` so the OpenAPI selector for `{owner_id}` can be validated.",
+                    display_path.display()
+                )),
+            ));
+            return false;
+        }
+    };
+
+    let Some(paths) = document
```
tohboeh5: OpenAPI validation currently accepts any YAML object with a `paths` map. That means an unrelated config file can pass as long as it has matching keys, because this branch never verifies that the document has an OpenAPI version field (`openapi` or `swagger`). Please gate this on a real OpenAPI root before treating the trace as valid.

PR #696, opened 2026-05-14
[Feature]: Add syu task scope to map a request onto requirements, policies, and features

### LINKED ISSUE #648: [Feature]: Add syu task scope to map requests into persistent spec scope without creating a task layer

After classifying a request, maintainers still need help finding the relevant requirements, checking whether policy or philosophy changes are needed, and deciding which features should move to `planned`. That scoping step is still fragmented across manual search, relate, and review flows.

## What change do you want?

Add a `syu task scope` command that maps a request onto the existing spec graph.

Acceptance criteria:
- the command can surface candidate requirements related to the request
- the result can flag when the request appears to imply policy or philosophy discussion, not only a feature edit
- the output can suggest candidate features that may need planned-state updates before implementation starts
- the command remains grounded in the checked-in spec graph instead of trying to infer code intent from git alone

## Specification impact

This likely needs task-planning feature coverage, CLI docs, and examples that connect the scope result to the rest of the implementation flow.

## Alternatives considered

Continue telling users to combine `search`, `relate`, and manual judgment. That works, but it leaves the most important planning step without one coherent workflow.

## Before submitting

tohboeh5:
Closes #648

Adds syu task scope to map a request artifact onto the checked-in spec graph before implementation starts. The command surfaces candidate requirements, flags likely policy/philosophy discussion, and highlights features that may need planned-state updates.

Validation:
- cargo run -- validate .
- cargo test --test task_command --test help_command --lib

Note: the full cargo test suite still has an unrelated failure in tests/init_command.rs::init_command_bootstraps_language_templates_that_validate_accept in this environment.

### DIFF
--- a/src/command/task.rs
+++ b/src/command/task.rs
@@ -527,12 +527,8 @@ fn print_classify_text_output(request_path: &Path, outcome: &ClassificationOutco
     print_items("explicit items", &outcome.explicit_items);
     print_items("related items", &outcome.related_items);
     println!("reasons:");
-    if outcome.reasons.is_empty() {
-        println!("- none");
-    } else {
-        for reason in &outcome.reasons {
-            println!("- {reason}");
-        }
+    for reason in &outcome.reasons {
+        println!("- {reason}");
     }
 }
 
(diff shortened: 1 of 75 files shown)


src/command/task.rs
```diff
@@ -580,6 +856,38 @@ fn print_items(heading: &str, items: &[SearchResult]) {
...
+                item.linked_requirements.join(", ")
+            );
+        }
+    }
+}
+
+fn bool_label(value: bool) -> &'static str {
+    if value { "yes" } else { "no" }
+}
+
```
tohboeh5, line 890: The registry entry here uses `scaffold_stem`, but the feature registry convention expects the `kind` to match the feature folder/layout. This will produce an entry like `kind: checkout-flow` with `file: core/core.yaml`, which is inconsistent and will make the generated registry harder to consume. Use the feature kind or folder name that matches the path instead.

PR #697, opened 2026-05-16
Extend syu log to inspect deleted IDs and lifecycle events

tohboeh5:
Summary:
- Extend `syu log` to resolve deleted IDs through the historical index.
- Report lifecycle events for creation, moves, removal, and redefinition attempts in text and JSON.
- Update reviewer docs and the command card to explain the deleted-ID fallback.

Validation:
- `cargo test --test log_command`

Closes #638

### DIFF
--- a/src/command/log.rs
+++ b/src/command/log.rs
@@ -220,9 +220,13 @@ pub fn run_log_command(args: &LogArgs) -> Result<i32> {
         scope: scope.as_ref(),
         path_filter: path_filter.as_deref(),
     })?;
-    if args.include_related && view.target.status != "historical" {
-        let related =
-            collect_related_tracked_paths(&workspace, &args.id, args.kind, path_filter.as_deref())?;
+    if args.include_related && view.target.status == "current" {
+        let related = collect_related_tracked_paths(
+            &workspace,
+            &view.target.id,
+            args.kind,
+            path_filter.as_deref(),
+        )?;
         view.target.tracked_paths.extend(related);
         dedupe_tracked_paths(&mut view.target.tracked_paths);
     }
--- a/src/command/log.rs
+++ b/src/command/log.rs
@@ -1670,6 +1670,7 @@ mod tests {
 
     #[test]
     fn resolve_historical_history_target_covers_all_supported_kinds() {
+        let _lock = PATH_LOCK.lock().unwrap_or_else(|err| err.into_inner());
         let repo = tempdir().expect("tempdir should exist");
         init_test_git_repository(repo.path());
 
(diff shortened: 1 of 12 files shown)


src/command/trace.rs
```diff
@@ -209,59 +277,43 @@ fn run_trace_range(workspace: &Workspace, range: &str, format: OutputFormat) ->
                         range: range.to_string(),
                         files: Vec::new(),
                         skipped_files: Vec::new(),
-                        summary: TraceRangeSummary {
-                            changed_files_total: 0,
```
tohboeh5: This guard only checks matched results. Skipped diff paths are ignored here, so a range with an out-of-workspace file can still exit 0 even though it should be out of scope.

src/command/log.rs
```diff
@@ -81,14 +88,57 @@ struct MatchedCommit {
...
+    target: &'a HistoryTarget,
+    repository_root: &'a Path,
+    kind: HistoryKind,
+    include_related: bool,
+    scope: Option<&'a HistoryScope>,
+    path_filter: Option<&'a Path>,
+    lifecycle_events: &'a [HistoryLifecycleEvent],
+    commits: &'a [MatchedCommit],
+}
+
```
tohboeh5, line 152: This still calls build_relation_report with the raw selector. When the ID has been deleted, the historical lookup above succeeds, but this related-surface expansion still resolves against the current workspace and can fail. Skip related expansion for historical targets or resolve it from the historical view first.

tohboeh5:
Overall, the PR is on the right track: the deleted-ID fallback in `syu log` is useful, and the new lifecycle output makes the historical path much clearer.

Strengths:
- The historical lookup is integrated into both text and JSON output.
- The lifecycle events give useful visibility into create/move/remove/redefine transitions.
- The new tests cover the core deleted-ID path and the normal history cases.

Main concerns:
- `syu review --range ... --allowed-id ...` still ignores skipped diff paths, so a range with an out-of-workspace file can return success even though it should be out of scope.
- `syu log <deleted-id> --include-related` still asks the current relation report to resolve the raw selector, which can fail for deleted IDs even after the historical target has been found.

Validation performed:
- `git status --short`
- `gh pr view 697 --json ...`
- `gh pr checks 697`
- `gh pr diff 697 --patch --color=never > /tmp/pr-697.patch`
- Direct repro of the skipped-path guard behavior with `target/debug/syu trace --range ... --allowed-id ...`
- Source inspection of the changed `trace` and `log` code paths

Follow-up issues:
- None beyond the two inline comments above.

PR #699, opened 2026-05-16
[Feature]: Add item history and lifecycle views to syu app

tohboeh5:
Summary:
- Added a git-backed history endpoint for `syu app` and shared it with the CLI log history builder.
- Extended the browser app to request history for the selected item ID, preserve deleted historical IDs from the index, and render lifecycle/commit timelines.
- Updated the app guide and browser-app feature spec to describe the history view and its relationship to `syu log`.

Validation:
- cargo test render_history_text --lib
- cargo test item_history_route_returns_git_history_for_selected_item --lib
- npm run check

Note:
- Playwright history e2e was attempted, but the webServer startup timed out in this environment.

Closes #639

### DIFF
--- a/src/command/log.rs
+++ b/src/command/log.rs
@@ -41,7 +41,7 @@ const GIT_ENVIRONMENT_KEYS: [&str; 8] = [
 ];
 
 #[derive(Debug, Serialize)]
-struct JsonLogOutput {
+pub(crate) struct HistoryResponse {
     id: String,
     entity_kind: &'static str,
     title: String,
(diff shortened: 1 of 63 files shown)


src/command/log.rs
```diff
@@ -271,10 +400,329 @@ fn resolve_history_entity<'a>(
...
+) -> Result<HistoricalDefinitionDiscovery> {
+    for record in load_git_history_records(repository_root, section_relative, &[], None)? {
+        for path in record.paths {
+            if !is_yaml_document(&path) {
+                continue;
+            }
+            let Some(contents) = git_show_file(repository_root, &record.commit.sha, &path)? else {
+                continue;
+            };
+            if !contents.contains(id) {
```
tohboeh5: Matching `id` with `contents.contains(id)` can pick any YAML file that merely references the ID, not the deleted item’s own definition. That can resolve the wrong title/path and lifecycle timeline when another document links to the same ID. Parse the document and match the actual item ID instead of doing a substring search.

PR #714, opened 2026-05-18
[Feature] Add syu task plan

tohboeh5:
Closes #702\n\nAdds `syu task plan` with Goal Plan generation, output file writing, YAML/JSON/text rendering, and updated help/spec/docs coverage.

### DIFF
--- a/src/command/task.rs
+++ b/src/command/task.rs
@@ -28,6 +28,7 @@ use crate::{
 use super::lookup::{SearchResult, WorkspaceEntity, WorkspaceLookup};
 
 const REQUEST_ARTIFACT_VERSION: u32 = 1;
+const GOAL_PLAN_VERSION: u32 = 1;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
 #[serde(rename_all = "snake_case")]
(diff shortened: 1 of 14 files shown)


src/command/task.rs
```diff
@@ -522,6 +618,594 @@ pub fn run_task_scaffold_command(args: &TaskScaffoldArgs) -> Result<i32> {
...
 }
 
+pub fn run_task_plan_command(args: &TaskPlanArgs) -> Result<i32> {
+    let workspace = load_workspace(&args.workspace)?;
+    let request_artifact = load_request_artifact(&args.request)?;
+    let explicit_ids = request_artifact.explicit_ids();
+    let scope_outcome = scope_request(&workspace, &request_artifact)?;
+    let plan = build_goal_plan(&workspace, &scope_outcome, &explicit_ids, &args.request)?;
+
+    let render_format = if args.output.is_some() && matches!(args.format, TaskPlanFormat::Text) {
```
tohboeh5: This rewrites `--format text` to YAML whenever `--output` is set, so the CLI can no longer emit a text plan to a file even though that combination is not documented as invalid. Either preserve the requested format or reject the combination explicitly.

src/command/task.rs
```diff
@@ -522,6 +618,594 @@ pub fn run_task_scaffold_command(args: &TaskScaffoldArgs) -> Result<i32> {
...
+                include: scope_include,
+                exclude: vec!["docs/generated/**".to_string(), "target/**".to_string()],
+            },
+        },
+        test_plan,
+        coverage: JsonTaskPlanCoverage {
+            mode: "changed_lines".to_string(),
+            threshold: 100,
+        },
+        completion: JsonTaskPlanCompletion {
```
tohboeh5, line 738: The completion checklist is hard-coded to `.syu/tasks/current.yaml`, so any `--output` path other than that conventional filename makes the generated plan inaccurate. Please derive this from `args.output` or keep the text generic.

PR #716, opened 2026-05-19
Keep CI coverage reporting, but remove minimum coverage thresholds

tohboeh5:
Summary:
- Remove coverage percentage enforcement from CI while keeping LCOV generation, coverage summaries, artifact upload, and changed-line reporting.
- Update the repository quality spec references to the new non-blocking coverage-reporting function.

Validation:
- bash -n scripts/ci/coverage.sh
- python3 -m py_compile scripts/ci/write-spec-coverage-summary.py
- git diff --check

Closes #715

### DIFF
--- a/docs/generated/site-spec/features/repository/quality.md
+++ b/docs/generated/site-spec/features/repository/quality.md
@@ -44,7 +44,7 @@ description: "Generated reference for docs/syu/features/repository/quality.yaml"
         - **symbols**:
           - configure_llvm_tools
           - enforce_diff_coverage
-          - enforce_lcov_threshold
+          - report_lcov_coverage
           - run_coverage
       - **file**: scripts/ci/check-generated-docs-freshness.sh
         - **symbols**:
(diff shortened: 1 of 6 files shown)


tohboeh5 (commented):
Overall: the PR cleanly removes the hard LCOV threshold and keeps the docs in sync, but it also removes the nonzero exit from the changed-line coverage path. That turns `scripts/ci/coverage.sh pr` into a reporting-only check, so uncovered changed lines will no longer fail CI.

Strengths:
- The docs/spec updates match the renamed helper.
- The LCOV reporting path still prints coverage and uncovered executable lines.

Main concern:
- `enforce_diff_coverage` no longer exits nonzero when misses are present, so the PR coverage gate is effectively disabled.

Validation:
- `git status --short`
- `gh pr view 716 --json id,number,title,body,author,headRefName,baseRefName,files,comments,reviews,state,isDraft,mergeStateStatus,closingIssuesReferences,headRefOid`
- `gh pr checks 716`
- `gh pr diff 716 --patch --color=never`
- Reviewed `scripts/ci/coverage.sh` and the generated/spec docs diff.

PR #751, opened 2026-05-30
Replace the app server with a Workbench API and realtime event bus

tohboeh5:
Summary:
- Added a new `syu-workbench-server` crate with a typed Axum API, SSE event stream, filesystem reload watching, and local bind safety.
- Added the `syu workbench` CLI wrapper and `workbench.bind` / `workbench.port` config support.
- Updated the workbench spec and generated docs for the new server feature.

Validation:
- `cargo fmt --all`
- `cargo check -p syu-workbench-server`
- `cargo test -p syu-workbench-server`
- `cargo check -p syu`

Closes #737


tohboeh5 (commented):
Overall, the PR is well structured and the checks are green. The new Workbench server, CLI wiring, and generated docs hang together coherently.

Main concerns:
- `request.new` is exposed in the action registry, but `execute_action` has no `request.new` arm and nothing persists `state.request`, so the active-request flow cannot bootstrap.
- The filesystem watcher ignores deletions, so removing a spec or workspace file will not trigger a reload and the server can keep serving stale data.

Validation performed:
- `git status --short`
- `gh pr view 751 --json id,number,title,body,author,headRefName,baseRefName,files,comments,reviews,state,isDraft,mergeStateStatus,closingIssuesReferences,headRefOid`
- `gh pr checks 751`
- `gh pr diff 751 --patch --color=never > /tmp/pr-751.patch`

Follow-up issues: none.

